PR #1, opened 2018-02-27
As an user, I can view information for a role in Pulp

daviddavis:
fixes #3179
https://pulp.plan.io/issues/3179


pulp_ansible/app/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+from django.db import models
+
+from pulpcore.plugin.models import Content, Importer, Publisher
+
+
+log = getLogger(__name__)
+
+
+class AnsibleRoleContent(Content):
```
daviddavis: Should we call this `AnsibleRole`?

bmbouter: Yes I think so.

pulp_ansible/app/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+    class Meta:
+        unique_together = (
+            'namespace',
+            'name',
+            'version'
+        )
+
+
+class AnsiblePublisher(Publisher):
```
bmbouter: Will this publisher only handle `AnsibleRole` stuff or will it handle other types in the future as well like `TheNextAnsibleThing`? If it's always-and-forever `AnsibleRole`, should this be called `AnsibleRolePublisher`?

pulp_ansible/app/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+from pulpcore.plugin.models import Content, Importer, Publisher
+
+
+log = getLogger(__name__)
+
+
+class AnsibleRole(Content):
+    """
+    A content type respresenting an ansible role
```
bmbouter: s/ansible role/Ansible Role/

pulp_ansible/app/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+class AnsiblePublisher(Publisher):
+    """
+    A Publisher for Ansible content.
+    """
+
+    TYPE = 'ansible'
+
+
+class AnsibleImporter(Importer):
```
bmbouter: Similar question here as with `AnsiblePublisher`.

bmbouter:
I was able to install the plugin and makemigrations for it! It did make migrations correctly for the 3 models.

pulp_ansible/app/models.py
daviddavis: My assumption (at least for now) was that we'd want to publish all Ansible content in a repo together (not just roles).

bmbouter: Great then this is the right name. Ack.

PR #4, opened 2018-03-12
Adding publishing code and updating README

daviddavis:
This commits adds basic publishing which allows the ansible-galaxy to install the role version directly through a URL. Also, updated the README.

fixes #3467
https://pulp.plan.io/issues/3467

### DIFF
--- a/pulp_ansible/app/tasks/publishing.py
+++ b/pulp_ansible/app/tasks/publishing.py
@@ -1,0 +1,75 @@
+import logging
+
+from gettext import gettext as _
+
+from celery import shared_task
+
+from pulpcore.plugin.models import (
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/tasks/publishing.py
```diff
@@ -0,0 +1,75 @@
...
+
+    log.info(
+        _('Publishing: repository=%(repository)s, version=%(version)d, publisher=%(publisher)s'),
+        {
+            'repository': repository_version.repository.name,
+            'version': repository_version.number,
+            'publisher': publisher.name,
+        })
+
+    with WorkingDirectory():
```
daviddavis: Need to check if this is being used.

PR #6, opened 2018-04-05
Create a roles content endpoint in the api

daviddavis:
fixes #3549
https://pulp.plan.io/issues/3549

New endpoints:

```
/api/v3/content/ansible/roles/
/api/v3/content/ansible/roles/<role_pk>/versions/
```

### DIFF
--- a/pulp_ansible/app/models.py
+++ b/pulp_ansible/app/models.py
@@ -2,16 +2,18 @@ from logging import getLogger
 
 from django.db import models
 
-from pulpcore.plugin.models import Content, Model, Importer, Publisher
+from pulpcore.plugin.models import Content, ContentArtifact, Importer, Publisher
 
 
 log = getLogger(__name__)
 
 
-class AnsibleRole(Model):
+class AnsibleRole(Content):
     """
     A model representing an Ansible Role
     """
+
+    TYPE = 'ansible-role'
 
     namespace = models.CharField(max_length=64)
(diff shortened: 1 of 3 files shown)


bmbouter:
I tested this code and it works really nicely. The two endpoints show off sync'd content nicely.

bmbouter:
So is it possible with this PR to upload a AnsibleRoleVersion to an empty Pulp instance?

daviddavis:
@bmbouter you can currently upload a role version artifact. That works. You cannot create a AnsibleRoleVersion though. I added a NotImplementedError to this path:

https://github.com/pulp/pulp_ansible/pull/6/files#diff-bfcb746c009dd53f978b79fa78e902a1R71

And I left this bit of functionality for this ticket:

https://pulp.plan.io/issues/3529

daviddavis:
There is one downside to switching from Model to Content and that is the relationship to Artifact. I don't think it's a big one though as long as we make sure not to expose it in the REST API. I did consider maybe creating a new class in pulpcore (e.g. ArtifactlessContent) that could be extended by Content and AnsibleRole.

The benefit of the change is that Model couldn't be used in the Content viewset—which I think has to do with the master-model stuff. Another benefit is that we get the natural keys methods which don't seem to do much ATM though. And lastly, if we ever add some future functionality to Content, AnsibleRole will automatically have it.

PR #17, opened 2018-04-24
Updated w/r/t core changes

dralley:
* Removed sync_mode and download_policy from instructions
* Switched pulp's API root from ```/api/v3/``` to ```/pulp/api/v3/```
* Switched pulp's content root from ```/content/``` to ```/pulp/content/```


dralley:
```sync_mode``` and ```download_policy``` are still there, originally I was going to remove them also in this PR, but it does involve some minor code changes, so I left it.

PR #30, opened 2018-05-17
Only run the deploy stage if tagged

daviddavis:


bmbouter:
@daviddavis For this, was the deploy stage running, but it just wasn't pushing the bits and this prevents if from running? I'm just curious if we were pushing builds unexpectedly.

daviddavis:
yea, so it was running the deploy stage but NOT pushing the bits unless there was tag. this skips the stage.

PR #31, opened 2018-05-23
Renaming version field to name on AnsibleRoleVersion

daviddavis:
fixes #3603
https://pulp.plan.io/issues/3603

### DIFF
--- a/pulp_ansible/app/galaxy/serializers.py
+++ b/pulp_ansible/app/galaxy/serializers.py
@@ -24,8 +24,7 @@ class GalaxyAnsibleRoleVersionSerializer(serializers.Serializer):
     A serializer for Ansible role versions in Galaxy.
     """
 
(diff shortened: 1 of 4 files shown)


bmbouter:
@daviddavis This will be a a beta-cycle backwards incompatible change right? I created us a breaking-changes label and tagged this PR with it.

I think we should adopt the same release notes approach that we use for core where we just link to the PR changes, release by release using the time-limited URL. I can help put that together if you ack that this is how we want to do it.

I plan to test this next week and if it tests well, I'll merge it.

daviddavis:
@bmbouter it does introduce a backwards incompatible change which reminds me that I need to update the README. Adding WIP tag.

daviddavis:
I think we're redoing the data model in pulp_ansible which makes this change unnecessary.

PR #35, opened 2018-06-19
Add new linting rules, formatting + docstrings

dralley:


dralley:
Most of the docstrings aren't particularly useful atm, but they can be made more useful as time goes on.  I think it's good to get the linting checks enabled early, though.

PR #45, opened 2018-08-03
Detect more types of errors, and save time

Ichimonji10:
Detect more types of errors, and save time

`.travis/script.sh` runs unit tests and functional tests. Unfortunately,
it only fails when certain expressions fail. For example, if
`pulp-manager migrate --noinput` fails, then the script as a whole will
(eventually) return non-zero, but if the prior `pulp-manager
makemigrations pulp_ansible` expression fails, then the script as a
whole might return zero. This means that some types of errors won't be
discovered.

This script also continues executing even when failures are detected
early on. For example, if the migrations command fails, then the script
will continue on and attempt to run unit tests and functional tests.
This is a huge waste of time. Why bother running any tests at all if the
migrations have failed?

The script also catches non-zero exit codes and re-raises them as "1".
This is problematic, because a script's exact exit code can be very
meaningful, and raising "1" hides this information.

Make the script bail out when most kinds of errors are detected. This
makes the test script catch more types of errors and saves time.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -1,40 +1,41 @@
-#!/usr/bin/env sh
-set -v
+#!/usr/bin/env bash
+# coding=utf-8
+set -veuo pipefail
 
-flake8 --config flake8.cfg || exit 1
+# lint code.
+flake8 --config flake8.cfg
 
-result=0
-
+# Run migrations.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
-
 pulp-manager migrate auth --noinput
 pulp-manager makemigrations pulp_app --noinput
 pulp-manager makemigrations pulp_ansible
 pulp-manager migrate --noinput
-if [ $? -ne 0 ]; then
-  result=1
-fi
 
-pushd ../pulp
-coverage run manage.py test pulp_ansible.tests.unit
-if [ $? -ne 0 ]; then
-  result=1
-fi
-popd
+# Run unit tests.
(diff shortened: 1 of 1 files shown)


.travis/script.sh
```diff
@@ -1,40 +1,35 @@
...
-if [ $? -ne 0 ]; then
-  result=1
-  cat ~/django_runserver.log
-  cat ~/resource_manager.log
-  cat ~/reserved_worker-1.log
-fi
-
-exit $result
+show_logs_and_return_non_zero() {
+    readonly local rc="$?"
```
Ichimonji10: If pytest returns a non-zero exit code, then we capture it here, and re-raise the same exit code later. This is much nicer than `return 1`, which masks pytest's exit code.

Ichimonji10:
@daviddavis 

```sh
# The changes made to this script's environment appear to affect Travis' own
# scripts. Perhaps this script is sourced by Travis?
set +euo pipefail
```

.travis/script.sh
```diff
@@ -1,40 +1,42 @@
-#!/usr/bin/env sh
-set -v
-
-flake8 --config flake8.cfg || exit 1
-
-result=0
-
+#!/usr/bin/env bash
```
Ichimonji10: sh and bash are quite different. :smile:

daviddavis: I personally prefer `sh` since I don't use `bash` but I'm fine with this change.

PR #47, opened 2018-08-03
Change how test requirements are listed

Ichimonji10:
The pulp-ansible plugin distributes unit and functional tests for its
code alongside the code itself. Given this, the plugin should tell users
what they need to install in order to run those tests.

Define a new optional dependency called "test." It can be used like so:

    pip install pulp-ansible[test]

This install not only pulp-ansible and its hard dependencies, but also
everything needed for running unit tests and functional tests. The
tooling for actually *running* these tests once the dependences are
installed is not so great. This can be improved with some hard thought
and elbow grease.

See:
https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,19 @@ setup(
     author_email='pulp-dev@redhat.com',
     url='https://github.com/pulp/pulp_ansible',
     install_requires=requirements,
+    extras_require={
+        'test': [
+            'coverage',
+            'coveralls',
+            'flake8',
+            'flake8-docstrings',
+            'flake8-quotes',
+            'flake8-tuple',
+            'mock',
+            'pulp-smash',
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Also relevant: https://caremad.io/posts/2013/07/setup-vs-requirement/

daviddavis:
> I'm pretty sure coveralls isn't needed. But I didn't drop it in this PR, because that's a logically separate change.

Can you remind me to do so after we merge this?

Ichimonji10:
@daviddavis Yes, I can remind you.

daviddavis:
Opened https://github.com/pulp/pulp_ansible/pull/48

PR #61, opened 2018-09-29
Remove AnsiblePublisher

daviddavis:
fixes #4048
https://pulp.plan.io/issues/4048

Requires https://github.com/pulp/pulp/pull/3681 and https://github.com/pulp/pulp/pull/3680


pulp_ansible/app/tasks/publishing.py
```diff
@@ -2,72 +2,25 @@
...
-            'repository': repository_version.repository.name,
-            'version': repository_version.number,
-            'publisher': publisher.name,
-        })
-
-    with WorkingDirectory():
-        with Publication.create(repository_version, publisher) as publication:
-            populate(publication)
+    with Publication.create(repository_version, pass_through=True) as publication:
+        pass
```
daviddavis: There's two things this block does for us. One is to set publication as the created_resource on the task and also it handles setting complete on the publication.

Another option would be to copy over this code to do something like:

```python
with transaction.atomic():
  publication = Publication(repository_version=repository_version, pass_through=True, complete=True)
  publication.save()
  resource = CreatedResource(content_object=publication)
  resource.save()
```

Thoughts?

daviddavis:
Another questoin: I am not sure if the call to create Ansible publications should be sync or async. I think that it needs to be because it should lock the repo version?

bmbouter:
To do it similarly to how we've been doing it already, it would be async and would lock on either the repository version or the repository itself. A publisher isn't involved so we don't need to lock on that. We need to shore up our docs in this area some.

Here are the locks pulp_file uses for its publisher for example.  https://github.com/pulp/pulp_file/blob/4c788d7aed99f993a376771270b672de3b46e345/pulp_file/app/viewsets.py#L136

daviddavis:
Cool, that's what [this PR does](https://github.com/pulp/pulp_ansible/pull/61/files#diff-bfcb746c009dd53f978b79fa78e902a1R185) so I think this is good to go then.

bmbouter:
@daviddavis legit, I agree. Ty

bmbouter:
@daviddavis I'm going to test this now via the README

daviddavis:
@bmbouter great. make sure you pull the latest changes from pulpcore.

bmbouter:
I just ran through all of these instructions and it all works great! I was able to download content from the Publication as the last step. Thank you so much @daviddavis !

PR #64, opened 2018-10-05
Adds migrations and upgrade docs

bmbouter:
This contains an initial migration with the current state of the model
layer. Also since it could change over time, the instructions are
updated to only have users run the migrations and not make them.

https://pulp.plan.io/issues/4065
closes #4065


bmbouter:
This can't be merged because core doesn't have committed migrations yet. We probably can't complete this for a few more weeks once it gets into an RC.

PR #67, opened 2018-12-04
Update smash tests to be in line w/ other plugins

dralley:


pulp_ansible/tests/functional/api/test_sync.py
```diff
@@ -27,6 +27,38 @@ class BasicSyncTestCase(unittest.TestCase):
     def setUpClass(cls):
         """Create class-wide variables."""
         cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.json_handler)
+
+    def test_file_decriptors(self):
```
dralley: @bmbouter Double check that this test makes sense for Pulp Ansible plugin

bmbouter: I think this is a good test to have even though there isn't anything during sync that suggests file descriptor use specifically. We don't want to leak file descriptors from common operations though, so it's still good to have.

PR #69, opened 2018-12-11
Use ContentFilter instead of BaseFilterSet

dralley:


dralley:
@bmbouter While this shouldn't be merged yet, it can be reviewed as-is.  The only additions I plan to make will be to smash tests, which QE will be the reviewers of.

dralley:
@bmbouter On second though, this totally can be merged as-is.  I can update the smash tests as a totally separate piece of work.

PR #70, opened 2018-12-12
Fix the pulpcore path

daviddavis:
[noissue]


daviddavis:
There was a big in this PR. I had `pip install -e ./pulpcore` instead of `pip install -e ./pulp`. Travis should hopefully be green now.

PR #74, opened 2019-01-04
Various updates

dralley:
[noissue]


.travis/script.sh
```diff
@@ -7,6 +7,7 @@ flake8 --config flake8.cfg || exit 1
 
 # Run migrations.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+export PULP_CONTENT_HOST=localhost:8080
 pulp-manager migrate auth --noinput
```
daviddavis: I think you can drop this line. I assume it's perhaps what's causing the error in Travis?

bmbouter: Yes the state of pulp_ansible is that it needs to be updated to use the Stages API. Until then we can't get the checks passing because it needs newer code that it can use from core.

I'm ok if we merge anyway because these will reduce the amount of future work and it's already broken. @daviddavis are you ok w/ this?

daviddavis: That sounds good to me.

PR #75, opened 2019-01-19
Add policy to remote tests

dralley:
Update _gen_verbose_remote to add download policy field. Add constant
DOWNLOAD_POLICIES.

re #4182
https://pulp.plan.io/issues/4182


bmbouter:
Per an irc convo w/ @daviddavis we're going to blind merge this. Travis is already broken because master received updates but hasn't been ported to work with the Stages API yet. The changes here will be needed, so we all agree to blind merge.

PR #77, opened 2019-02-20
Use hyphens for Pulp3' systemd units name.

ipanova:
closes #4187
https://pulp.plan.io/issues/4187


daviddavis:
Merging as this is currently causing Travis to fail.

PR #79, opened 2019-03-03
Port sync to stages

mdellweg:


mdellweg:
@pulp/qe it seems like the crud remote tests pass only occasionally.

daviddavis:
@mdellweg this is a community plugin and so the tests aren't maintained by @pulp/qe. I'll try to take a look into the test failure.

By the way, thanks a lot for submitting this PR. It's really appreciated. 👍

daviddavis:
Here's the failure: `{'_content': b'{"policy":["\\"cache_only\\" is not a valid choice."]}'`. Looks like it's being set here in the test randomly. I think maybe we should set DOWNLOAD_POLICIES to only `['immediate']`. AFAIK, this plugin doesn't support lazy syncing (yet).

mdellweg:
Thank you. I think we can try to be bold and just turn it on. I think the stages cover that for us.

daviddavis:
I was thinking we could do that in a follow up PR. Up to you.

daviddavis:
Actually it looks like the policy of 'cache_only' got renamed to 'streamed'. If you could just update [this line in the tests](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/constants.py#L10), I can merge this PR.

Thanks again @mdellweg.

mdellweg:
@daviddavis Thank you, that worked. :champagne:

daviddavis:
💃🥂🎉🎈

PR #81, opened 2019-03-16
 Fix content serializers based on pulpcore changes

daviddavis:


bmbouter:
I went to makemigrations for this app and I think it's still depedning on `pulp_app` being the name of the migrations in pulpcore.

`django.db.migrations.exceptions.NodeNotFoundError: Migration ansible.0001_initial dependencies reference nonexistent parent node ('pulp_app', '0001_initial')`

bmbouter:
My own local migrations were incompatible. Removing them and making them fresh resolved it.

PR #84, opened 2019-03-28
Cascade delete of role versions

mdellweg:
Using 'on_delete=PROTECT' prevents roles to be deleted. In turn deleting
orphans fails. Setting 'on_delete=CASCADE' deletes role versions
together with their role.


bmbouter:
@mdellweg I believe we're going to remodel this to not have both of them be content units. See https://pulp.plan.io/issues/3818 for more info on the plan.

@daviddavis do you agree that the resolution of https://pulp.plan.io/issues/3818 will resolve the issue this PR is intended to solve too? If so should we close this PR or ?

daviddavis:
I think that is correct. I was leaving this PR open just as a reminder to test out orphan delete until https://pulp.plan.io/issues/3818 got fixed.

PR #90, opened 2019-04-16
Collection support

bmbouter:


bmbouter:
@daviddavis I'm merging this now since it fully works for me, like we talked about. We can fix things when you come back if there are issues.

PR #91, opened 2019-04-17
Combining role and role version into a single content unit

daviddavis:
https://pulp.plan.io/issues/3818
fixes #3818


bmbouter:
I went through the entire readme and everything worked for me! It also makes this conceptually simpler for users.

pulp_ansible/app/models.py
```diff
@@ -10,30 +10,14 @@
 
 class AnsibleRole(Content):
     """
-    A model representing an Ansible Role.
+    A content type representing an Ansible Role version.
     """
 
     TYPE = 'ansible-role'
```
bmbouter: This causes the API serializer to have `"_type": "ansible.ansible-role",` What about renaming `AnsibleRole` to `Role` throughout and this to `'role'`?

bmbouter: We can handle this separately. I wrote this issue and put it onto the 0.1 release. https://pulp.plan.io/issues/4709

PR #93, opened 2019-05-01
Renames AnsibleRole -> Role

bmbouter:
This caused to user-visible change except for links they would receive
while browsing the content API.

I also noticed the publication change needed a README update so I put
that in here too.

https://pulp.plan.io/issues/4710
closes #4710


bmbouter:
@daviddavis I'm merging this while you are out, but tagging you so you can see it. We can review/fix anything when you come back.

PR #94, opened 2019-05-02
Update docs to correct URLs

bmbouter:
I just hand-tested the docs and these updates are from the latest
responses so they are up to date.

https://pulp.plan.io/issues/4515
closes #4515


bmbouter:
@daviddavis here are some docs fixes that I'm merging. We can fix them when you get back if there are problems.

PR #97, opened 2019-05-07
Fixes mazer config url

bmbouter:
https://pulp.plan.io/issues/4634
closes #4634


bmbouter:
@daviddavis here is the mazer url fix. I'm merging and we can fix/adjust it later if its not right.

PR #100, opened 2019-05-21
Convert to sphinx and majorly reorganize

bmbouter:
This introduces a Sphinx docs project and moves almost all of the
content from the README into the sphinx site. After merging, this will
be enabled onto read the docs.

This also reworks the workflows sections significantly and follows the
_snippets approach pioneered in pulp_python.

https://pulp.plan.io/issues/4766
closes #4766


docs/conf.py
```diff
@@ -0,0 +1,167 @@
...
+# -- Project information -----------------------------------------------------
+
+project = 'Pulp Ansible'
+copyright = '2019, Pulp Ansible Team'
+author = 'Pulp Ansible Team'
+
+# The short X.Y version
+version = '0.1'
+# The full version, including alpha/beta/rc tags
+release = '0.1.0rc5'
```
daviddavis: I'm not super familiar with how docs work. Is this something we have to remember to bump when we do a release of pulp_ansible? Is there a way to fetch this from setup.py?

docs/workflows/mazer-client.rst
```diff
@@ -0,0 +1,97 @@
...
+Install a Collection, hosted by Pulp, using ``mazer``
+=====================================================
+
+
+Mazer Configuration
+-------------------
+
+`Install mazer <https://galaxy.ansible.com/docs/mazer/install.html#latest-stable-release>`_ and
+use the `url` option to point to the `Distribution` your content should fetch from. For example,
+using the `Distribution` created in the sync workflow, the config would be::
```
daviddavis: This paragraph is very vague (ie which url option?)

docs/conf.py
bmbouter: I can add this into my next push. Good idea.

docs/workflows/mazer-client.rst
bmbouter: I know :/

bmbouter: These docs need a *lot*. I think we need to fully split roles and collections into their own single large parts of the site.

docs/workflows/sync.rst
```diff
@@ -0,0 +1,89 @@
...
+<https://docs.pulpproject.org/en/3.0/nightly/restapi.html#tag/repositories>`_
+
+
+Create a Remote
+---------------
+
+Creating a Remote object informs Pulp about an external content source, which most commonly is
+``https://galaxy.ansible.com/`` or another ``pulp_ansible`` instance. In this case, we will be
+syncing all Roles where ``namespace=elastic`` on Galaxy. You can see those roles
+`here <https://galaxy.ansible.com/api/v1/roles/>`_.
```
daviddavis: Does this need `?namespace=elastic`?

bmbouter: it does, adding in next push.

docs/workflows/sync.rst
```diff
@@ -0,0 +1,89 @@
...
+Reference (pulpcore): `Repository API Usage
+<https://docs.pulpproject.org/en/3.0/nightly/restapi.html#tag/repositories>`_
+
+
+Create a Remote
+---------------
+
+Creating a Remote object informs Pulp about an external content source, which most commonly is
+``https://galaxy.ansible.com/`` or another ``pulp_ansible`` instance. In this case, we will be
+syncing all Roles where ``namespace=elastic`` on Galaxy. You can see those roles
```
daviddavis: Here you have 'Roles' and 'roles'.

docs/workflows/sync.rst
```diff
@@ -0,0 +1,89 @@
...
+    {
+        "_href": "/pulp/api/v3/remotes/ansible/ansible/e1c65074-3a4f-4f06-837e-75a9a90f2c31/",
+    }
+
+.. todo::
+
+    Add a reference link to the live API
+
+
+Sync repository foo with remote
```
daviddavis: Repository and Remote should be titlecased as in the previous sections.

docs/workflows/sync.rst
```diff
@@ -0,0 +1,89 @@
...
+-------------------------------
+
+Use the remote object to kick off a synchronize task by specifying the repository to
+sync with. You are telling Pulp to fetch content from the remote and add to the repository.
```
daviddavis: Same here.

docs/workflows/sync.rst
daviddavis: Probably should use an article before 'Remote' too.

docs/workflows/sync.rst
daviddavis: You should probably do a search for Role in docs and make sure they're titlecased.

docs/workflows/upload-role.rst
```diff
@@ -0,0 +1,59 @@
...
+
+
+Reference (pulpcore): `Artifact API Usage
+<https://docs.pulpproject.org/en/3.0/nightly/restapi.html#tag/artifacts>`_
+
+
+Create a Role from the Artifact
+-------------------------------
+
+Now that Pulp has the tarball, its time to make it into a Role unit of content. We hope to add
```
daviddavis: it's time

docs/workflows/mazer-client.rst
daviddavis: That makes sense to me. So let's postpone that.

docs/conf.py
bmbouter: I messed with this for a bit, but I want to leave it alone. I think instead we should adopt https://pypi.org/project/bumpversion/ which I know a lot of successful projects rely on. How about that for now?

docs/workflows/sync.rst
bmbouter: fixed in next push, ty.

docs/workflows/sync.rst
bmbouter: fixed in next push, ty.

docs/conf.py
daviddavis: Sounds good to me. Maybe file an issue if you feel like it.

docs/workflows/upload-role.rst
bmbouter: agreed, but we need some way to determine the version. We need the tarball to contain that data somehow and that would be above the Role standard, but I think it's what is needed.

docs/workflows/sync.rst
bmbouter: fixed in next push.

bmbouter:
@daviddavis I pushed a new version with all those fixes.

docs/conf.py
bmbouter: I made one here  https://pulp.plan.io/issues/4864

PR #102, opened 2019-05-28
Add mazer whitelist sync

bmbouter:
Adds a CollectionRemote which users can use to sync a whitelist of
collections. With the sync a repsository is required and a new
respoitory version is created with the sync'd content downloaded.

If mazer downloads content that already exists it will not add
additional Artifacts or Collection units to pulp_ansible.

If sync discovers no new collections it does not create a
RepositoryVersion.

https://pulp.plan.io/issues/4764
closes #4764


bmbouter:
@daviddavis I'm going to merge this since you're out for a few days. If there is anything we need to fix I'm happy to fix it. I tested it all myself and it worked for me.

PR #103, opened 2019-05-28
version bump to 0.2.0b1

bmbouter:
Also a small docs fix

[noissue]


bmbouter:
@daviddavis here is the 0.2 version bump like we talked about so we can get back to a beta cycle.

PR #104, opened 2019-05-29
Switch to using generated Travis CI scripts

bmbouter:
This switches to the tool introduced by the script added to the plugin
template with this PR:

https://github.com/pulp/plugin_template/pull/42

https://pulp.plan.io/issues/4571
closes #4571

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,33 +7,65 @@ python:
     - "3.6"
     - "3.7"
 env:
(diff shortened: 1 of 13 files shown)


bmbouter:
@daviddavis this switches us to the installer and gives us the pypi bindings. I'm going to merge it and we can fix later if it's not right.

PR #105, opened 2019-05-29
Travis pipeline needs to `make html`

bmbouter:
The pipeline wasn't running `make html` correctly so it wouldn't fail if
there were docs errors. This causes the pipeline to run `make html`
correctly.

It also fixes the docs verison number to be correct. We really need to
start using bumpversion.

Also it adds in the latest api schema which was forgotten too.

https://pulp.plan.io/issues/4571
closes #4571


bmbouter:
@daviddavis I'm merging this so that the pipeline runs the docs correctly.

@dkliban: thank you so much for all the epic help!

PR #106, opened 2019-05-29
Adds docs links to bindings bits

bmbouter:
This adds a bindings section to the docs which has links to both the
Ruby and Python bindings, the core bindings, and the REST api browsable
docs for both too.

https://pulp.plan.io/issues/4844
closes #4844


bmbouter:
@daviddavis this adds bindings to the RTD site. We can review when you get back and fix anything that needs fixing then. I'm going to merge for now.

PR #114, opened 2019-06-17
Set version to .dev for the master branch

goosemania:
A correct version would be shown for installations from the master branch.
Bindings will be generated with a proper version as well.

re #4936
https://pulp.plan.io/issues/4936

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ with open('README.rst') as f:
 
 setup(
     name='pulp-ansible',
-    version='0.2.0b1',
+    version='0.2.0b1.dev',
     description='Pulp plugin to manage Ansible content, e.g. roles',
     long_description=long_description,
     license='GPLv2+',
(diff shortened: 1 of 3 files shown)


goosemania:
I think 0.2.0b1 hasn't been released yet, so 0.2.0b1.dev should be the one. @bmbouter , @daviddavis ?

setup.py
```diff
@@ -12,7 +12,7 @@
 
 setup(
     name='pulp-ansible',
-    version='0.2.0b1',
+    version='0.2.0b1.dev',
```
bmbouter: Also every plugin has the version in a second place too. It's the __version__ = in the __init__.py file of the package. That needs to be changed for any of them also.

goosemania: Is it for towncrier purposes?

bmbouter: Not especially for towncrier, although towncrier does use it. It's from pep396 https://www.python.org/dev/peps/pep-0396/. If we were doing it super to the spec we would derive the setup.py from __version__ or vice versa as in [step 8](https://www.python.org/dev/peps/pep-0396/#specification). I think using `bumpversion` would be good too.

bmbouter:
I agree it should be 0.2.0b1.dev because we haven't released 0.2.0b1 yet.

bmbouter:
Thanks @goosemania !

PR #122, opened 2019-07-03
Add docs on Collection upload

bmbouter:
The page existed, but had role examples where there should be Collection
upload examples. This provides two upload examples, and also links to
the mazer section of the docs for uploading via `mazer publish`.

https://pulp.plan.io/issues/4939
closes #4939

### DIFF
--- a/docs/workflows/upload-collection.rst
+++ b/docs/workflows/upload-collection.rst
@@ -1,56 +1,29 @@
 Upload Content
 ==============
 
-Upload an Artifact to Pulp
---------------------------
+Upload a Collection via the REST API
+------------------------------------
 
-This example will upload the ``geerlingguy/ansible-role-postgresql`` role. You can download that
-locally with::
+A collection tarball can be uploaded via the REST API and the server determines the metadata and
+creates the Collection model.
 
-    curl -L https://github.com/geerlingguy/ansible-role-postgresql/archive/master.tar.gz -o pg.tar.gz
+Suppose you have a file locally named collection.tgz you can upload that to ``pulp_ansible`` with::
 
-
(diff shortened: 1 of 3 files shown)


docs/workflows/upload-collection.rst
```diff
@@ -1,56 +1,29 @@
...
 ==============
 
-Upload an Artifact to Pulp
---------------------------
+Upload a Collection via the REST API
+------------------------------------
 
-This example will upload the ``geerlingguy/ansible-role-postgresql`` role. You can download that
-locally with::
+A collection tarball can be uploaded via the Rest API and the server determines the metadata and
```
daviddavis: s/Rest/REST/

docs/workflows/upload-collection.rst
```diff
@@ -1,56 +1,29 @@
...
+creates the Collection model.
 
-    curl -L https://github.com/geerlingguy/ansible-role-postgresql/archive/master.tar.gz -o pg.tar.gz
+Suppose you have a file locally named collection.tgz you can upload that to ``pulp_ansible`` with::
 
+    http --form POST :24817/ansible/collections/ file@collection.tgz sha256=abc123
```
daviddavis: Maybe drop the sha256 option from this command since you cover it in the next section.

docs/workflows/upload-collection.rst
bmbouter: Fixed in next push, ty.

docs/workflows/upload-collection.rst
bmbouter: Good catch, this was a mistake.

PR #123, opened 2019-07-03
Cause second sync content to associate correctly

bmbouter:
The content associate only occured when the sync created the content
unit. When that content unit is associated with another repository the
sync doesn't associate the content correctly.

This associates content regardless of if this sync created it or it
already existed.

https://pulp.plan.io/issues/4997
closes #4997


bmbouter:
This would conflict with another PR, so the fix is being included there. This fix is now in https://github.com/pulp/pulp_ansible/pull/125

PR #127, opened 2019-07-09
Advanced collecttions models

cutwater:
Implement collection database models that store additional
collection information retrieved by importing collection artifact.

fixes #5105
https://pulp.plan.io/issues/5105

### DIFF
--- a/pulp_ansible/app/models.py
+++ b/pulp_ansible/app/models.py
@@ -1,8 +1,14 @@
 from logging import getLogger
 
 from django.db import models
+from django.contrib.postgres import fields as psql_fields
 
-from pulpcore.plugin.models import Content, Remote, RepositoryVersionDistribution
+from pulpcore.plugin.models import (
+    Content,
+    Model,
+    Remote,
+    RepositoryVersionDistribution
+)
 
 
 log = getLogger(__name__)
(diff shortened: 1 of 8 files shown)


pulp_ansible/app/models.py
```diff
@@ -34,30 +40,73 @@ class Meta:
...
+        on_delete=models.CASCADE,
+        related_name='versions',
+        null=True
+    )
     version = models.CharField(max_length=128)
 
+    deprecated = models.BooleanField(default=False)
+    quality_score = models.FloatField(null=True)
+    metadata = psql_fields.JSONField(default=dict)
+    contents = psql_fields.JSONField(default=list)
```
daviddavis: We support other dbs (e.g. mariadb, mysql, etc) in pulp_ansible. I'm debating about how we can optimally continue to support them but store this data. My first thought is to store this as a json-encoded string in a textfield. Would that work?

bmbouter: One option would be to make it more structured altogether and have more fields. Could we get some examples of this data in practice shown?

cutwater: We are planning to use PostgreSQL FTS for search, aren't we?
Even though it can be de-normalized, in Galaxy we perform some filter queries against data stored in JSON fields.

daviddavis: So for full text search, we could still make pulp_ansible work for users who choose not to use postgresql. They just won't have full-text search.

It'd be nice to keep support for other databases if possible. If there's no easy way to support other databases, then we can go the postgresql-only route. I just want to check.

cutwater: I'm concerned about extra complexity it may create for models and how to support this in Django.
PostgreSQL FTS works efficient with appropriate (psql-specific) indexes and tsvector fields created in database. 
I wonder if it's even possible to enable \ disable at least on per-model level in django database specific models.

daviddavis: I'd like to at least get a picture of what that complexity would look like before we decide. I'm not ruling out dropping other dbs but I'd like to have all the info before making a decision.

bmbouter: The full text search feature is going to raise an exception when the 'q' object is used. So running on MariaDB with that is still feasible with that feature missing. With this change however the whole thing will be postgresql. That is what we talked about at the call so this isn't surprising to me. Today I was thinking about its impact though, and this makes a lot of sense for storing the data formats Galaxy already has, postgreSQL only could make running this with the 10+ other plugins that run on mariaDB could be a problem. It's not a deal-breaker, but I want us to at least consider options.

Can we see more examples of data that goes in there?

cutwater: @bmbouter I think the main question is how to split and maintain driver-specific models \ fields within the same django application?

chouseknecht:
Work related to https://github.com/ansible/galaxy-dev/issues/5

pulp_ansible/app/models.py
```diff
@@ -34,30 +40,76 @@ class Meta:
...
+    collection = models.ForeignKey(
+        Collection,
+        on_delete=models.CASCADE,
+        related_name='versions',
+        editable=False,
+    )
+    version = models.CharField(max_length=128, editable=False)
+
+    quality_score = models.FloatField(null=True, editable=False)
+    metadata = psql_fields.JSONField(default=dict, editable=False)
```
bmbouter: What about making this a `CharField` and before writing to it serialize the json data to a string at the db level? The serializer could continue to read/write json natively.

daviddavis: The problem is querying the field. You can't query a TextField the same way. See:

https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/fields/#querying-jsonfield

bmbouter: I see what you are saying. @daviddavis so that filtering also works with the filtering at the API level and its `__` syntax too?

@cutwater Is the data stored in this field have the same schema or are there many schemas? Also what types of queries will be sent against this?

PR #128, opened 2019-07-09
Add 'latest' boolean for Collection filtering

bmbouter:
When 'latest' is True, only include a  Collection if it is the newest
version of <namespace, name>.

https://pulp.plan.io/issues/5076
closes #5076

### DIFF
--- a/pulp_ansible/app/viewsets.py
+++ b/pulp_ansible/app/viewsets.py
@@ -1,5 +1,8 @@
+from collections import defaultdict
 from gettext import gettext as _
+from packaging.version import parse
 
+from django_filters import BooleanFilter
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework import serializers, viewsets
 from rest_framework.decorators import detail_route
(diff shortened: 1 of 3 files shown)


bmbouter:
@dkliban when I look at this filterset in the docs browser it doesn't have a label or help text. Do you have any advice on how I can add it?

image

bmbouter:
Those strange names in the picture I'm resolving as part of another issue.

pulp_ansible/app/viewsets.py
```diff
@@ -70,6 +73,38 @@ class CollectionFilter(ContentFilter):
...
+
+        Args:
+            queryset: The already-formed queryset for modification
+            name: The name of the parameter, 'latest'
+            value: The value of the argument. This is checked if 'True' or not.
+
+        Returns:
+            Queryset with non-latest collections excluded if value is True.
+        """
+        if value:
```
daviddavis: This is nitpick but a guard clause might be better here. 

https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

pulp_ansible/app/viewsets.py
```diff
@@ -70,6 +73,38 @@ class CollectionFilter(ContentFilter):
...
+            namespace_name_dict = defaultdict(lambda: defaultdict(list))
+            for collection in queryset.all():
+                version_entry = (parse(collection.version), collection.pk)
+                namespace_name_dict[collection.namespace][collection.name].append(version_entry)
+
+            not_latest_pks = []
+            for namespace, name_dict in namespace_name_dict.items():
+                for name, version_list in name_dict.items():
```
daviddavis: Does python provide a best practice for naming unused variables?

pulp_ansible/app/viewsets.py
```diff
@@ -70,6 +73,38 @@ class CollectionFilter(ContentFilter):
...
+                    version_list.sort(reverse=True)
+                    version_list_not_latest = version_list[1:]  # remove the latest one
+                    for version, pk in version_list_not_latest:
```
daviddavis: Here too.

daviddavis:
Seems like it would be much more efficient to filter by the latest pks. That list will be shorter than the non-latest.

bmbouter:
I filed this issue to handle the names not showing up correctly.  https://pulp.plan.io/issues/5101

pulp_ansible/app/viewsets.py
bmbouter: I agree but this if statement isn't nested though?

daviddavis: So like with a guard clause, this would be:

```python
if not value:
    return queryset

namespace_name_dict = defaultdict(lambda: defaultdict(list))
# and so on...
```

pulp_ansible/app/viewsets.py
bmbouter: I don't think it does, but it strives for readability as a first-class concern. There is some talk of including variable typing as an anti-pattern, but I think the jury is still out on that. Also having the types to me makes them more comprehensible.

pulp_ansible/app/viewsets.py
bmbouter: I try to name them what they are for maximized clarity.

pulp_ansible/app/viewsets.py
daviddavis: I'm not sure I understand. What does typing have to do with naming?

pulp_ansible/app/viewsets.py
bmbouter: Oh word that is way better. Thank you for explaining this to me. I'm pushing the fix now.

PR #131, opened 2019-07-09
Add full-text search to Collection filterset

bmbouter:
With this the user can submit their full-text querystring as the 'q'
parameter. The results are rank sorted. It provides a weighted result
with the following weights:

- Namespace name (weight A)
- Collection name (weight A)
- Collection tags (weight B)
- Collection content names (weight C)
- Collection description (weight D)

https://pulp.plan.io/issues/5075
closes #5075

### DIFF
--- a/pulp_ansible/app/models.py
+++ b/pulp_ansible/app/models.py
@@ -2,6 +2,7 @@ from logging import getLogger
 
 from django.db import models
 from django.contrib.postgres import fields as psql_fields
+from django.contrib.postgres import search as psql_search
 
 from pulpcore.plugin.models import Content, Model, Remote, RepositoryVersionDistribution
 
@@ -56,7 +57,7 @@ class Tag(Model):
 
 class CollectionVersion(Content):
     """
-    A content type representing a Collection.
+    A content type representing a CollectionVersion.
 
     This model is primarily designed to adhere to the data format for Collection content. That spec
     is here: https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html
@@ -87,6 +88,7 @@ class CollectionVersion(Content):
 
     TYPE = "collection_version"
 
+    # Data Fields
     authors = psql_fields.ArrayField(models.CharField(max_length=64), default=list, editable=False)
     contents = psql_fields.JSONField(default=list, editable=False)
     dependencies = psql_fields.JSONField(default=dict, editable=False)
@@ -101,10 +103,14 @@ class CollectionVersion(Content):
     repository = models.URLField(default="", blank=True, max_length=128, editable=False)
     version = models.CharField(max_length=32, editable=False)
 
+    # Foreign Key Fields
     collection = models.ForeignKey(
         Collection, on_delete=models.CASCADE, related_name="versions", editable=False
     )
     tags = models.ManyToManyField(Tag, editable=False)
+
+    # Search Fields
+    search_vector = psql_search.SearchVectorField(default="")
 
     @property
     def relative_path(self):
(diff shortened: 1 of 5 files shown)


daviddavis:
This code looks good to me. Are you planning to add indexes? If so, I'll wait to test it out and approve.

bmbouter:
I want to add add a few things:
* indexes
* a check that raises an exception if 'q' is used with mariadb
* docs

bmbouter:
Since we're dropping mariadb from pulp_ansible soon I'm not adding the check. The indexes are added so that is done. I'd like to merge with the changelog only and we can look at the docs more later.

bmbouter:
I had uploaded the `greetings_namespace-hello-11.11.11.tar.gz` package from mazer's example README and then searched with `q=greet`. You can do this 

```
http :24817/pulp/api/v3/content/ansible/collections/?q=greet
```

You can also combine it with other filterset options like:

```
http :24817/pulp/api/v3/content/ansible/collections/?q=greet&namespace=foobar
```

pulp_ansible/app/models.py
```diff
@@ -34,6 +34,36 @@ class Meta:
...
+
+        Args:
+            text: The text to search for
+
+        Returns:
+            A queryset annotated with the .extra() query which brings in the full-text query results
+                from PostgreSQL.
+
+        """
+        return self.extra(
```
cutwater: Django documentation discourages `extra()` method usage [1]. This kind of request can be built with standard Django ORM functionality. Please refer to [2] and [3] for an example.

[1] https://docs.djangoproject.com/en/2.2/ref/models/querysets/#extra
[2] https://github.com/ansible/galaxy/blob/devel/galaxy/api/internal/search.py#L86
[3] https://github.com/ansible/galaxy/blob/devel/galaxy/api/internal/search.py#L130

pulp_ansible/app/migrations/0002_add_fulltext_search_indexes.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ansible', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "CREATE INDEX ansible_collection_ts_idx ON ansible_collection USING gin(to_tsvector('english', version || ' ' || name || ' ' || namespace));"
```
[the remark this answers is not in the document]
bmbouter: I was able to get this working with all the data stored in the new model. Let me know if this doesn't look right. Thanks!

pulp_ansible/app/tasks/collections.py
```diff
@@ -128,6 +128,7 @@ def nowhere(*args, **kwargs):
                         import_pb.increment()
 
         collection_versions = CollectionVersion.objects.filter(pk__in=collection_version_pks)
+        raise Exception('oh no!')
```
ipanova: I think this was meant to be removed

bmbouter: Yes this is here for testing. I added the WIP PR because the relavence rank sorting isn't working yet. I'll try to keep the right labels on these going forward. Thank you for looking at the open PRs.

bmbouter:
I pushed my final round of changes. I tested it locally and everything seemed to work fine for me including search results being included or not, and ranking including the weighted rank results.

You can test using:

```
http :24817/pulp/api/v3/content/ansible/collections/?q=greet
```

You can also combine it with other filterset options like:

```
http :24817/pulp/api/v3/content/ansible/collections/?q=greet&namespace=foobar
```

PR #135, opened 2019-07-12
version bump 0.2.0b1 release and changelog

bmbouter:
[noissue]


setup.py
```diff
@@ -5,15 +5,15 @@
 requirements = [
     'mazer',
     'packaging',
-    'pulpcore-plugin~=0.1rc2',
+    'pulpcore-plugin~=0.1rc3',
```
bmbouter: I bumped the version here. It seems right to keep up with the latest to me. @daviddavis do you agree?

PR #141, opened 2019-07-23
 Add test to consume content using mazer

kersommoura:
Add test to consume content hosted by Pulp using mazer.

closes:#4915

Required PR:PulpQE/pulp-smash#1212

### DIFF
--- a/.travis/ansible_playbook.yml
+++ b/.travis/ansible_playbook.yml
@@ -1,0 +1,36 @@
+---
+- hosts: all
+  pre_tasks:
+    - name: Load DB variables
+      include_vars: '{{ pulp_db_type }}.yml'
+  vars:
+    pulp_default_admin_password: admin
+    pulp_source_dir: '{{ ansible_env.TRAVIS_BUILD_DIR | dirname }}/pulpcore/'
+    pulp_plugin_source_dir: "{{ ansible_env.TRAVIS_BUILD_DIR | dirname }}/pulpcore-plugin"
+    pulp_install_plugins:
+      pulp-ansible:
+        app_label: "ansible"
+        source_dir: "$TRAVIS_BUILD_DIR"
+    ansible_python_interpreter: '/opt/pyenv/shims/python3'
+    pulp_user: 'travis'
+    developer_user: 'travis'
+    pulp_install_db: false
+    pulp_preq_packages: []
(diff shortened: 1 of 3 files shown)


daviddavis:
This LGTM. @bmbouter ?

kersommoura:
The failing test is passing locally. 

``` bash
 pulp_ansible/tests/functional/api/collection/test_consume_content.py::MazerConsumeCotentTestCase.test_consume_content ✓100% ██████████

Results (6.34s):
       1 passed
```

Locally the extra variables were added to the the systemd env.
```
Environment="PULP_ANSIBLE_API_HOSTNAME=http://localhost:24817" Environment="PULP_ANSIBLE_CONTENT_HOSTNAME=http://localhost:24816/pulp/content"
```

daviddavis:
@kersommoura you could set them in the travis environment by setting them in `.travis/post_before_script.sh`

kersommoura:
@daviddavis, thanks.

Somehow I did think that this [PR](https://github.com/pulp/pulp_ansible/commit/3c3eb5d5dc32a14af739bfe7d6baca89538f71f7) added the same settings, in a different manner.

daviddavis:
Theoretically we should be able to update the settings in [the Travis playbook](https://github.com/pulp/pulp_ansible/blob/master/.travis/playbook.yml) but it's managed by the plugin_template and there's no way to do it currently.

bmbouter:
@kersommoura pulp_ansible needs to have these variables set all the time so I believe the right place is to set them in here with this PR https://github.com/pulp/pulp_ansible/blob/master/.travis/playbook.yml#L26

test_requirements.txt
```diff
@@ -8,3 +8,4 @@ git+https://github.com/PulpQE/pulp-smash.git#egg=pulp-smash
 # pin pydocstyle until https://gitlab.com/pycqa/flake8-docstrings/issues/36 is resolved
 pydocstyle<4
 pytest
+mazer
```
daviddavis: Is this needed? I feel like in a test environment, we'd install the packages from setup.py and mazer is there already:

https://github.com/pulp/pulp_ansible/blob/master/setup.py#L5

kersommoura: You are right. I removed the other one.

PR #142, opened 2019-07-26
updating pulp_ansible docs

fabricio-aguiar:
[noissue]


daviddavis:
@fabricio-aguiar this looks great. Looks like there is a failure though.

PR #147, opened 2019-07-31
validating collection remote url

fabricio-aguiar:
closes #4996

### DIFF
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -66,6 +66,27 @@ class CollectionRemoteSerializer(RemoteSerializer):
     A serializer for Collection Remotes.
     """
 
+    def validate(self, data):
+        """
+        Validate a a url to ensure it does not end with slashes.
+
+        Args:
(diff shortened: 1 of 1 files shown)


pulp_ansible/app/serializers.py
```diff
@@ -66,6 +66,27 @@ class CollectionRemoteSerializer(RemoteSerializer):
     A serializer for Collection Remotes.
     """
 
+    def validate(self, data):
+        """
+        Validates data.
```
bmbouter: Can this state more clearly what is being done here?

pulp_ansible/app/serializers.py
```diff
@@ -66,6 +66,27 @@ class CollectionRemoteSerializer(RemoteSerializer):
...
+
+        Args:
+            data (dict): User data to validate
+
+        Returns:
+            dict: Validated data
+
+        Raises:
+            rest_framework.serializers.ValidationError: If invalid data
```
bmbouter: Can the invalid cases be described here too please.

PR #148, opened 2019-07-31
Moving 'metadata' to be fields

bmbouter:


pulp_ansible/app/models.py
```diff
@@ -65,14 +65,30 @@ class CollectionVersion(Content):
...
+    authors = psql_fields.JSONField(default=list, editable=False)
+    contents = psql_fields.JSONField(default=list, editable=False)
+    description = models.TextField(blank=True, editable=False)
+    documentation = psql_fields.JSONField(default=dict, editable=False)
+    license = psql_fields.JSONField(default=list, editable=False)  # list?
+    name = models.CharField(max_length=32, editable=False)
+    namespace = models.CharField(max_length=32, editable=False)
+    quality_score = models.FloatField(null=True, editable=False)
+    readme = models.TextField(blank=True, editable=False)  # Is this really the filename?
+    tags = psql_fields.JSONField(default=list, editable=False)
```
bmbouter: become a separate model. Have it listed in the serializer of CollectionVersion (not it's own viewset and serializer).

pulp_ansible/app/models.py
```diff
@@ -65,14 +65,30 @@ class CollectionVersion(Content):
 
     TYPE = "collection_version"
 
+    authors = psql_fields.JSONField(default=list, editable=False)
```
bmbouter: Array field.

pulp_ansible/app/models.py
```diff
@@ -65,14 +65,30 @@ class CollectionVersion(Content):
...
+    contents = psql_fields.JSONField(default=list, editable=False)
+    description = models.TextField(blank=True, editable=False)
+    documentation = psql_fields.JSONField(default=dict, editable=False)
+    license = psql_fields.JSONField(default=list, editable=False)  # list?
```
bmbouter: Array field.

bmbouter: licenses

pulp_ansible/app/models.py
```diff
@@ -65,14 +65,30 @@ class CollectionVersion(Content):
 
     TYPE = "collection_version"
 
+    authors = psql_fields.JSONField(default=list, editable=False)
+    contents = psql_fields.JSONField(default=list, editable=False)
+    description = models.TextField(blank=True, editable=False)
+    documentation = psql_fields.JSONField(default=dict, editable=False)
```
bmbouter: documentation_blobs

pulp_ansible/app/models.py
```diff
@@ -65,14 +65,30 @@ class CollectionVersion(Content):
...
+    license = psql_fields.JSONField(default=list, editable=False)  # list?
+    name = models.CharField(max_length=32, editable=False)
+    namespace = models.CharField(max_length=32, editable=False)
+    quality_score = models.FloatField(null=True, editable=False)
```
bmbouter: remove me

pulp_ansible/app/models.py
awcrosby: `docs_blob` might be a good name here

pulp_ansible/app/models.py
```diff
@@ -65,14 +65,30 @@ class CollectionVersion(Content):
...
+    readme = models.TextField(blank=True, editable=False)  # Is this really the filename?
```
awcrosby: Since the importer will return all that is in the manifest, it will likely keep returning this value (the filename), but as discussed earlier the readme text (and many other documentation bits) will be returned in docs json blob - so perhaps not model this?

bmbouter: I agree with you. Is it in the spec though? If it's in there and we remove it we won't be compatible. I'm not sure if it is or isn't.

awcrosby: It's in the spec so the collection author must include it, and the importer will use it to find the readme file and process it, but as far as post-import storing it in the database or not... i dont think that changes compatibility

pulp_ansible/app/tasks/collections.py
```diff
@@ -131,10 +131,34 @@ def import_collection(artifact_pk):
...
-            collection_version = CollectionVersion(
-                collection=collection, version=collection_info["version"]
-            )
+            if created:
+                CreatedResource.objects.create(content_object=collection)
+
+            dependencies = collection_info.pop('dependencies')
+            tags = collection_info.pop('tags')
+
+            # Mazer returns many None values. We need to let the default='' in models.py prevail
```
awcrosby: The importer currently also returns these as `null`/`None` so this will continue to be needed

pulp_ansible/app/tasks/collections.py
```diff
@@ -131,10 +131,34 @@ def import_collection(artifact_pk):
...
+
+            collection_version = CollectionVersion(collection=collection, **collection_info)
             collection_version.save()
+
+            for dependency in dependencies.values():
+                dependency_split = dependency.split('.')
+                namespace = dependency_split[0]
+                name = dependency_split[1]
+                found_dependency = CollectionVersion.objects.get(namespace=namespace, name=name)
+                collection_version.dependencies.add(found_dependency)
```
awcrosby: We want to fail import if a dependency + matching version is not found (FYI, can be in a later PR). Note: while community galaxy checks for these dependencies, it doesn’t store them as a FK (but that change may have already be discussed/decided). Galaxy’s dep check: https://github.com/ansible/galaxy/blob/028a0ae075d5497ef43b6704c431a6f4c19b4162/galaxy/worker/importers/collection.py#L28

pulp_ansible/app/models.py
bmbouter: I left it in because the whole model adheres to the spec right now.

pulp_ansible/app/tasks/collections.py
bmbouter: Yes this will cause that behaviour I expect. At L#153 the get() fetching 0 items will fail.

bmbouter:
@awcrosby I pushed the serializer changes. They have everything except the foreign key parts. I also need to make the viewset for the Tags model itself, and get it nested into CollectionVersion serializer. If you could look at what I have now that would be great.

pulp_ansible/app/tasks/collections.py
```diff
@@ -131,10 +131,42 @@ def import_collection(artifact_pk):
...
+                    collection_info.pop(key)
+
+            collection_version = CollectionVersion(collection=collection, **collection_info)
             collection_version.save()
+
+            for dependency in dependencies.values():
+                dependency_split = dependency.split(".")
+                namespace = dependency_split[0]
+                name = dependency_split[1]
+                found_dependency = CollectionVersion.objects.get(namespace=namespace, name=name)
```
cutwater: How is that supposed to work? There can be >1 version matching dependency specifier.

pulp_ansible/app/models.py
```diff
@@ -45,44 +43,67 @@ class Meta:
...
-    @property
-    def namespace(self):
-        """Returns collection namespace name."""
-        return self.collection.namespace
-
-    @property
-    def name(self):
-        """Returns collection name."""
-        return self.collection.name
+    dependencies = models.ManyToManyField("self", symmetrical=False, editable=False)
```
cutwater: Dependencies shouldn't be stored as many-to-many. A dependency version requirement is a version range specifier for example `> 1.2.3`, which means the following:

1. A collection may have more than one version of a specific collection it depends on. (See my comment below).
2. This relationship does not store information about version ranges provided by metadata.
3. Maintaining such dependency list requires checking for reverse dependencies for entire database. Let's say a collection `A` has a dependency on a collection `B = ">1.2.3"`. Collection `A` dependencies list has to be updated each time new version of a collection `B` is imported to pulp.

pulp_ansible/app/models.py
```diff
@@ -45,44 +43,67 @@ class Meta:
...
+        authors (psql_fields.ArrayField): A list of the CollectionVersion content's authors.
+        contents (psql_fields.JSONField): A JSON field with data about the contents.
+        description (models.TextField): A short summary description of the collection.
+        docs_blob (psql_fields.JSONField): A JSON field holding the various documentation blobs in
+            the collection.
+        documentation (models.URLField): The URL to any online docs.
+        homepage (models.URLField): The URL to the homepage of the collection/project.
+        issues (models.URLField): The URL to the collection issue tracker.
+        license (psql_fields.ArrayField): A list of licenses for content inside of a collection.
+        license_file (models.CharField): The path to the license file for the collection.
```
cutwater: Storing filename in database doesn't make sense. If `license_file` is provided in collection metadata, it should be loaded from the artifact and stored as a text. What's the point of storing a name of some file within an artifact?

This field is relevant for importer, it specifies a filename where to look for custom license text. For example the same works for README file. How README file is named inside an artifact doesn't matter, what matters is a file content.


CC: @awcrosby

pulp_ansible/app/tasks/collections.py
bmbouter: It's not going to work. I didn't realize the versionspec part of the design until later. This will be changing very soon (before merging).

pulp_ansible/app/models.py
bmbouter: I agree we should just store the data since it's an expression of version specifier not a specific named one. Let me rework this and then let's see what you think.

pulp_ansible/app/tasks/collections.py
cutwater: It'll also cause a ` MultipleObjectsReturned` exception if fetching `>1` items.

pulp_ansible/app/models.py
bmbouter: This model comes from the spec, so since it's in the spec it's here.  https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#structure

cutwater: There are several steps in the collection import process.

User writes a human readable `galaxy.yml` file, which describes in loose format collection metadata. Then mazer processes it and builds `MANIFEST.json` file, which is machine readable and has more strict format, so that it can be deserialized in unambiguous way.

`MANIFEST.json` is used to import a collection (e.g. load collection metadata, find where content is located, find where external files like README and LICENSE are stored), so that these fields can be loaded if needed.

Keynote: You're not working directly with the spec, you're not getting the exact data defined in spec.

The `license_file` attribute returned by importer doesn't have any sense because it is useful only in context of artifact being processed by importer.

Instead importer should either return license full text loaded from that files or ignore it until custom licenses are officially supported. Since this field is in spec, I believe we'll have to support it, but in a different way.

pulp_ansible/app/models.py
```diff
@@ -45,44 +43,67 @@ class Meta:
...
+        name (models.CharField): The name of the collection.
+        namespace (models.CharField): The namespace of the collection.
+        readme (models.TextField): The path to the Markdown (.md) readme file.
```
cutwater: Same for readme field. By the way:

Is it a path? If it's a path why it's TextField?
Is it a file contents? if so, please update docstring.

bmbouter: The docstrings are copy/pastes from the site here so whatever is written there is what I think should be written here.  https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#structure

bmbouter: It shouldn't be a `TextField` though that is a mistake. I'm going to fix that now. This should be a `CharField` since its the path.

cutwater: What's the point of storing data that is relevant only for importer?
What is actually relevant for program that calls importer is a content of this file.

pulp_ansible/app/models.py
bmbouter: I'm confused. do the fields in `MANIFEST.json` correspond with the fields described here:  https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#structure

pulp_ansible/app/models.py
bmbouter: These docs read to users as if they describe the specification for a Collection. The idea is that those same users will expect those fields whenever they manage Collection content inside of pulp_ansible.

I agree some of these fields are silly, but if they need to be removed then shouldn't they be removed from the spec?

I can at the same time understand why this field is here, because what if a client pushing/pulling collection data wants to automated pull out the readme. It needs to know it's path to do that.

More comments/feedback is welcome.

pulp_ansible/app/models.py
awcrosby: @bmbouter yes those fields correspond. But those fields are relevant to the importer (for the importer to find the readme and license_file files). After the import there is no requirement to store every field as they were found in `MANIFEST.json`.

For the custom license, if `license_file` is present the importer will minimally check the file exists, but since there is no direction beyond that, I think the model should ignore the `license_file` key for now.

cutwater: Not necessarily. The majority of metadata fields are stored as is. What more important is why these fields are there. In community Galaxy contents of `MANJFEST.json` are stored as is, therefore license and readme path are stored as well, we do not process that data.

But as we started modeling this data, let's focus on use cases for this data. Most of it is informational and is relevant for user. Some of it is technical, like filenames. It gives hint for a program that works with artifacts metadata where to look for specific files.

Again. This information is not relevant for anybody except program that processes the artifact.

pulp_ansible/app/models.py
cutwater: @bmbouter please see the explanation from me and Andrew above.

pulp_ansible/app/models.py
cutwater: @awcrosby Later we may want to store text of license in order to display it to user. As it's not supported by importer now, we should either not store this field at all or add empty text field for future.

bmbouter: I agree let's keep it use case focused. The main thing I'm hearing is that those fields are only used by the importer. The importer isn't the main thing I think about as the motivator of this change. I have a use case in mind that any user who see's a field written [here](https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#structure) would expect it in pulp_ansible too.

As a pulp_ansible user (not a galaxy developer) does this use case make sense?

To give an analogy, pulp_rpm models all attributes about RPMs in [its serializers](https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/serializers.py#L35-L188).

cutwater: @bmbouter Generally this information is not reserved for importer only. At the moment the importer and partially ansible-galaxy client (former mazer) are the only tools that process this data. This data is for any tool that works with an artifact. Once the artifact is loaded some of that data becomes irrelevant.

What is the use case of storing raw filenames that are never gonna be used?

If  API client wants to get README or LICENSE contents, it may request it from API server.
If this data is not stored on server, API client should work with artifact itself, it should download an artifact and read `MANIFEST.json`.
Use case when API client has to do both: download an artifact and additionally query API to get metadata that is already in the artifact is silly.

bmbouter: I feel like I understand your claim, but it doesn't speak to my concern.

I'll restate my concern. pulp_ansible should store all of the spec-defined metadata fields for content types it stores. If there wasn't a spec then this would be more debatable, but in this case there is. So since the spec defines it, pulp_ansible should store it. Please help me understand the problem with this approach.

cutwater: Can you please explain the reason for this?
I'm not against storing some extra data, it's totally fine if you store filenames along with its contents.

I'm saying that storing just filename, but not storing the filename contents, which is more relevant than the filename itself is very confusing to me.

bmbouter: Sure, the idea is the model field names should follow from the specification. The goal is to avoid a situation where pulp_ansible is defining yet-another-standard of what a collection means, in terms of its data definition. I'm not going to argue in favor of these being especially useful, they aren't, but that's not for pulp_ansible to decide. If it's in the spec we should include it.

Does this make any sense? What do you both think?

awcrosby: When referring to spec, its the spec of the galaxy.yml file in order for importing to occur. If in the future we decide to store custom license file contents, or model the documentation text snippets, we will be "defining" another data definition of a post-imported-collection. But of course that would be adding instead of removing fields.

Agreed it is not useful (if some downloaded a collection artifact tarball, and wanted to know the path of the custom license they could look at the MANIFEST.json rather than the pulp api results from the model). But maybe having it match what the author submitted could be ok for consistency.

cutwater: I think the key point here is: 

The document which is described in the spec (e.g. `galaxy.yml`) is not even stored in the artifact itself. 
Yes, we store data from it.
Yes, it's very similar.
But this is **NOT** the `galaxy.yml` file, which is described in the spec.

So referring to that spec in order to fully comply with it is wrong. 
This spec is only relevant for user repository and the program that builds an artifact out of it. Period.

bmbouter: @cutwater ok so which fields are the ones that you feel should be removed?

Also where can I learn more about the actual spec? I'm not as familiar with the standards of galaxy/ansible, maybe I'm working from the wrong spec as your saying.

cutwater: @bmbouter I don't think we have publicly available spec for `MANIFEST.json`. It should basically be a superset of `galaxy.yml`, with slight adjustments (for example license field is only a list and never a string).

Like I said, I'm ok with keeping those fields (doesn't make sense to me, but whatever, having two extra fields is not a big deal). 

What I'm trying to say it that instead of (in addition to) having those fields we should store actual file contents, that are more useful for pulp and potentially for galaxy, because users likely to be interested in file contents more than just filenames.

@awcrosby Please confirm on README file, do we store its contents separately or it's a part of documentation blob. As for license I would like importer to return file contents along with just filename, so that we store it as something useful.

awcrosby: We will store the README file's rendered html in the artifact's `docs_blob` key.

The importer can return the contents of the license file - it can be named `license_file_contents` or `custom_license`, it can be added at the same level as the `metadata` key. I'm thinking not embed the key inside `metadata` since that is validation of the MANIFEST.json `collection_info` rather than the results of processing the data.

bmbouter: OK the tests are now passing and all the other funcitonality is there so I want to adjust these fields and merge. Removing the `readme` field makes sense, especially since we'll be oferring the contents themselves through the `docs_blob` key.

Having `custom_license` for the license contents of the not SPDX licenses sounds good. What should we do for the `license_file` field though? Remove it?

awcrosby: as far as the model, i think removing `license_file` is fine (importer will still return it but it can just be ignored when saving to database)

bmbouter: OK I'm removing `license_file` and `readme`. They are silly indeed.

pulp_ansible/app/models.py
```diff
@@ -45,44 +43,72 @@ class Meta:
...
+        name (models.CharField): The name of the collection.
+        namespace (models.CharField): The namespace of the collection.
+        repository (models.URLField): The URL of the originating SCM repository.
+        version (models.CharField): The version of the collection.
 
     Relations:
 
-        collection: Reference to a collection model.
+        collection (models.ForeignKey): Reference to a collection model.
+        dependencies (models.ManyToManyField): An asymmetric reference to other CollectionVersion
```
awcrosby: @bmbouter I assume this should be removed from the relations docstring

bmbouter:
@awcrosby @cutwater @daviddavis this is ready for it's final review I think.

pulp_ansible/app/models.py
bmbouter: yup this is wrong. Thank you! Will fix now and push.

pulp_ansible/app/models.py
cutwater: Looks good to me.

pulp_ansible/app/models.py
```diff
@@ -45,44 +43,68 @@ class Meta:
...
 
+class Tag(Model):
+    """A model representing a Tag.
+
+    Fields:
+
+        name (models.CharField): The Tag's name.
+    """
+
+    name = models.CharField(max_length=32, unique=True, editable=False)
```
daviddavis: Why only 32 chars?

bmbouter: I figured they wouldn't be longer, but now it seems short. What should it be?

daviddavis: Actually 32 seems fine.

daviddavis: Or let's do 64 to be safe?

awcrosby: may want to encourage shorter than 64 tags, maybe keep 32 for now?

bmbouter:
Thanks everyone for all the help in shaping this work! After it's merged new bindings will be published that pair with it.

PR #152, opened 2019-08-01
wip

bmbouter:


pulp_ansible/app/tasks/collections.py
```diff
@@ -145,3 +92,117 @@ def import_collection(artifact_pk):
...
+
+        def collection_page_url(url, page=1):
+            parsed_url = urlparse(url)
+            new_query = parse_qs(parsed_url.query)
+            new_query["page"] = page
+            return urlunparse(parsed_url._replace(query=urlencode(new_query, doseq=True)))
+
+        def parse_metadata(download_result):
+            with open(download_result.path) as fd:
+                return json.load(fd)
```
bmbouter: return json.load(fd.read())  ?

pulp_ansible/app/tasks/collections.py
```diff
@@ -173,3 +106,111 @@ def import_collection(artifact_pk):
...
+            page_count = math.ceil(float(metadata["count"]) / float(PAGE_SIZE))
+            progress_bar.total = page_count
+            progress_bar.save()
+
+            yield metadata
+            progress_bar.increment()
+
+            # Concurrent downloads are limited by aiohttp...
+            not_done = set(
+                remote.get_downloader(url=get_page_url(remote.url, page)).run()
```
bmbouter: You mentioned the too-many open files and I think it's this that is causing it. Yes while aiohttp is limiting the concurrency of co-routines that are scheduled I think each co-route created (scheduled or not) will allocate file descriptors. I believe this loop could create a large number of downloaders.

The Stages API has these restrictors built in for the downloaders it creates because we also don't want to hold a large number of them in memory. So you probably want to integrate this get_downloader into the loop below so that the loop itself delays the generation of additional downloaders until some number of existing ones are completed.

Does this make sense?

fabricio-aguiar: It makes sense, but I am confused, why this piece of code works on roles but is not working here

bmbouter: Is it that the actual count of downloaders is effectively different since it's driven by the the page counts reported by the servers? Maybe run a comparison on the number of downloaders generated? Also you can look at the file descriptor count if you also monitor the process with `lsof` to see when they are being allocated for sure. Just a few ideas.

pulp_ansible/app/migrations/0004_alter_collection_version_name.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('ansible', '0003_add_tags_and_collectionversion_fields'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='collectionversion',
+            name='name',
+            field=models.CharField(editable=False, max_length=64),
```
fabricio-aguiar: some names have more than 32 characters

pulp_ansible/app/models.py
```diff
@@ -96,7 +96,7 @@ class CollectionVersion(Content):
     homepage = models.URLField(default="", blank=True, max_length=128, editable=False)
     issues = models.URLField(default="", blank=True, max_length=128, editable=False)
     license = psql_fields.ArrayField(models.CharField(max_length=32), default=list, editable=False)
-    name = models.CharField(max_length=32, editable=False)
+    name = models.CharField(max_length=64, editable=False)
```
fabricio-aguiar: some names have more than 32 characters

pulp_ansible/tests/functional/api/collection/test_consume_content.py
```diff
@@ -30,6 +30,7 @@ def setUpClass(cls):
         except exceptions.CalledProcessError:
             raise unittest.SkipTest("This test requires mazer client.")
 
+    @unittest.skip("Not using mazer on collections")
```
fabricio-aguiar: I don't know if we will need this test after I change to stages api, for now, it is breaking, so I've skipped

pulp_ansible/tests/functional/constants.py
```diff
@@ -38,15 +38,17 @@
...
+ANSIBLE_COLLECTION_FIXTURE_URL = urljoin(ANSIBLE_GALAXY_COLLECTION_URL, NAMESPACE_ANSIBLE)
 
 ANSIBLE_COLLECTION_REMOTE_PATH = urljoin(BASE_REMOTE_PATH, "ansible/collection/")
 
 COLLECTION_WHITELIST = "testing.ansible_testing_content"
 
 ANSIBLE_COLLECTION_CONTENT_NAME = "ansible.collection_version"
 
-ANSIBLE_COLLECTION_FIXTURE_COUNT = 1
+ANSIBLE_COLLECTION_FIXTURE_COUNT = 114
```
fabricio-aguiar: apparently, filters are not working for https://galaxy.ansible.com/api/v2/collections/
so this number could vary

pulp_ansible/tests/functional/constants.py
```diff
@@ -38,15 +38,17 @@
 # FIXME: replace this with the location of one specific content unit of your choosing
 ANSIBLE_URL = urljoin(ANSIBLE_FIXTURE_URL, "")
 
-ANSIBLE_GALAXY_COLLECTION_URL = "https://galaxy-dev.ansible.com"
+ANSIBLE_GALAXY_COLLECTION_URL = "https://galaxy.ansible.com/api/v2/collections/"
+
+ANSIBLE_COLLECTION_FIXTURE_URL = urljoin(ANSIBLE_GALAXY_COLLECTION_URL, NAMESPACE_ANSIBLE)
```
fabricio-aguiar: this filter https://galaxy.ansible.com/api/v2/collections/?namespace__name=ansible is not working, so we have 32 pages, and sometimes tests break because it takes more than 10 minutes

pulp_ansible/app/tasks/collections.py
```diff
@@ -46,92 +57,9 @@ def sync(remote_pk, repository_pk, mirror):
...
-                            collection, created = Collection.objects.get_or_create(
-                                namespace=info["namespace"], name=info["name"]
-                            )
-                            collection_version, created = CollectionVersion.objects.get_or_create(
-                                collection=collection, version=info["version"]
-                            )
-
-                            if created:
-                                # Create the tags
-                                tags = info.pop("tags")
```
fabricio-aguiar: where can I get this data?

pulp_ansible/app/tasks/collections.py
fabricio-aguiar: Now I got it, it was because I was using: https://galaxy-dev.ansible.com/api/v2/collections/
which has `"count": 11731`. When I changed to: https://galaxy.ansible.com/api/v2/collections/ which has `"count": 32`, the code finally worked!

pulp_ansible/app/tasks/collections.py
bmbouter: I expect that to come from the inspected manifest.json file:  https://github.com/ansible/mazer/blob/devel/tests/ansible_galaxy/collection_examples/hello/galaxy.yml

That data would show up here as the key 'tags' for example: https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/tasks/collections.py#L83

pulp_ansible/tests/functional/constants.py
bmbouter: I'm confused. https://galaxy.ansible.com/api/v2/collections/?namespace__name=ansible consistently yields 32 results (not pages).

fabricio-aguiar: sorry, it was results, not pages, but it is not filtering, you can see that some results have different names

pulp_ansible/tests/functional/constants.py
bmbouter: One idea I've seen before in pup_ansible tests when the live galaxy could change is to assert that there are "at least" that many units. That is already in place for the roles tests AIUI.

pulp_ansible/app/models.py
bmbouter: Great, thanks.

pulp_ansible/app/tasks/collections.py
bmbouter: @fabricio-aguiar So did instantiating a lot of downloaders cause the too many files issue? I think that would be a good bug to file if so.

pulp_ansible/tests/functional/constants.py
fabricio-aguiar: this tests could take more than 10 minutes: `test_successive_syncs_repo_version` because of it: https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/test_sync.py#L87-L89

pulp_ansible/tests/functional/api/collection/test_consume_content.py
bmbouter: This test is crucial because it's what ensures pulp content is installable. I can give pointers on how to fix, just let me know. It's likely an issue with the data being saved during sync causing the viewsets mazer relies on to be incorrect. I just went through this myself with the modelling changes.

bmbouter:
@fabricio-aguiar can you update the title and description of this PR please?

fabricio-aguiar:
@bmbouter I am not allowed to change the title and description, because you've opened this PR.

pulp_ansible/app/tasks/collections.py
fabricio-aguiar: how can I get manifest.json from the galaxy api?

pulp_ansible/tests/functional/constants.py
bmbouter: @fabricio-aguiar oh I see what you mean. The filter results are incorrect coming back from galaxy. Can you open an issue here https://github.com/ansible/galaxy/issues and then bring it up in #pulp-ansible

bmbouter: @fabricio-aguiar  For the full mirror test, I think we want a test that does that but we want to have Travis running the tests exclude that test. It would be great if the nightly cron ran it. I'm also ok for deleting the full mirror test for now and filing an issue about making this cron/PR nightly test separation separately. What do you think?

bmbouter:
@fabricio-aguiar then let me close it so you can open one, how does that sound?

pulp_ansible/app/tasks/collections.py
bmbouter: it's in the file that is downloaded. That downloading happens by the stages API. I think you want to inspect this file as aprt of the pre-save for content, or maybe the post-save.

fabricio-aguiar:
@bmbouter sounds good to me

pulp_ansible/app/tasks/collections.py
fabricio-aguiar: I've checked the `remote.download_concurrency` as you suggested and it is 20 for both, roles and collections

PR #156, opened 2019-08-06
Add is_highest field to CollectionVersion model

cutwater:
`is_highest` is a boolean field that indicates if a specific
collection version has the highest version in a set of available
collection versions.
This patch adds the field, migration and data migration for it.

NOTE: Since the semver comparison is implemented in Python, data
migration fetches all required data into memory. Even considering
limited number of fields that queries fetch, the operation may have
impact on performance or even fail if out of memory.

https://pulp.plan.io/issues/5223
closes #5223

### DIFF
--- a/pulp_ansible/app/models.py
+++ b/pulp_ansible/app/models.py
@@ -3,6 +3,7 @@ from logging import getLogger
 from django.db import models
 from django.contrib.postgres import fields as psql_fields
 from django.contrib.postgres import search as psql_search
+from django.db.models import UniqueConstraint, Q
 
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/models.py
```diff
@@ -123,6 +128,13 @@ def relative_path(self):
 
     class Meta:
         unique_together = ("namespace", "name", "version")
+        constraints = [
+            UniqueConstraint(
+                fields=("collection", "is_highest"),
+                name="unique_is_highest",
+                condition=Q(is_highest=True),
```
bmbouter: Oh I see what you mean about the partial condition.

PR #171, opened 2019-08-15
Removing latest filter Collection Version

fabricio-aguiar:
closes #5227

### DIFF
--- a/CHANGES/5227.removal
+++ b/CHANGES/5227.removal
@@ -1,0 +1,1 @@
+Removing `latest` filter Collection Version.
--- a/pulp_ansible/app/viewsets.py
+++ b/pulp_ansible/app/viewsets.py
@@ -76,6 +76,7 @@ class CollectionVersionFilter(ContentFilter):
 
     namespace = filters.CharFilter(field_name="namespace")
     name = filters.CharFilter(field_name="name")
+    is_highest = filters.BooleanFilter(field_name="is_highest")
     q = filters.CharFilter(field_name="q", method="filter_by_q")
 
     def filter_by_q(self, queryset, name, value):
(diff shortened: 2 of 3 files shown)


CHANGES/5227.misc
```diff
@@ -0,0 +1 @@
+Removing latest filter Collection Version.
```
bmbouter: I think this needs to be named `5227.removal` so it's listed in the breaking changes section of the release notes.

bmbouter: Also maybe put latest in single quotes, double quotes, or backticks.

CHANGES/5278.misc
```diff
@@ -0,0 +1 @@
+Adding is_highest filter for Collection Version.
```
bmbouter: Since this is an addition that is user facing I think it should be `5278.feature`.

bmbouter: Also similar w/ the single quotes, backticks, etc.

pulp_ansible/app/viewsets.py
```diff
@@ -105,39 +103,9 @@ def filter_by_q(self, queryset, name, value):
...
-                version_list.sort(reverse=True)
-                latest_pk = version_list[0][1]
-                latest_pks.append(latest_pk)
-
-        return queryset.filter(pk__in=latest_pks)
-
     class Meta:
         model = CollectionVersion
-        fields = ["namespace", "name", "version"]
+        fields = ["namespace", "name", "version", "is_highest"]
```
bmbouter: Should 'q' be in here too? What happens to CollectionVersion filtering if `fields` isn't specified at all?

fabricio-aguiar: I'm new to Django, and `q` wasn't there before, I don't know why!

CHANGES/5227.removal
```diff
@@ -0,0 +1 @@
+Removing `latest` filter Collection Version.
```
bmbouter: :+1:

PR #176, opened 2019-08-19
Adds collection upload to Galaxy V3 API

bmbouter:
The collection import view was not available in the Galaxy V3 API. Now
it is available at via post to `collections/` in the V3 API namespace.

https://pulp.plan.io/issues/5243
closes #5243


bmbouter:
This is slightly different than the ticket. It accepts a POST on the `/collections/` endpoint itself which seems more RESTful to me. @cutwater do you want it this way or with the POST to `/artifact/collections/` like the story is written and with a new view name like `collection-create`.

cutwater:
I don't think it's more restful. POST to `/collections/` list in terms of RESTful API should create a Collection resource, shouldn't it? However it does not create a collection itself, but rather a collection version. But making POST request to `/collections/<namespace>/<name>/versions/` is not possible because it requires `namespace` and `name` parameters to be set in path. Therefore we decided to move upload endpoint out of collections namespace and make it complementary with download API:

1. `GET /artifacts/collections/` - Download an artifact
2. `POST /artifacts/collections/` - Upload an artifact (create a new resource "artifact", additionally creating collection \ collection versions).

bmbouter:
@cutwater I can see your reasoning. I'm adjusting the url of this PR to `POST /artifacts/collections`/

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -81,6 +90,43 @@ def get_object(self):
...
         )
 
+    @swagger_auto_schema(
+        operation_description="Create an artifact and trigger an asynchronous task to create "
+        "Collection content from it.",
+        operation_summary="Upload a collection",
+        request_body=CollectionOneShotSerializer,
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def create(self, request, path=None):
```
bmbouter: @cutwater do you want me to move this to it's own view?

pulp_ansible/app/urls.py
```diff
@@ -35,7 +35,14 @@
...
     path(
-        "collections/", views_v3.CollectionViewSet.as_view({"get": "list"}), name="collections-list"
+        "collections/",
+        views_v3.CollectionViewSet.as_view({"get": "list"}),
+        name="collections-list",
+    ),
+    path(
+        "artifacts/collections/",
+        views_v3.CollectionViewSet.as_view({"post": "create"}),
+        name="artifact-collection-create",
```
bmbouter: @cutwater is there another name you imagined here?

pulp_ansible/app/galaxy/v3/views.py
cutwater: Yes, please

pulp_ansible/app/urls.py
cutwater: `collection-artifact-create` or `collection-artifact-upload` sounds better IMO.

pulp_ansible/app/galaxy/v3/views.py
bmbouter: fixed in most recent push.

pulp_ansible/app/urls.py
bmbouter: I went with `collection-artifact-upload`. Thank you.

bmbouter:
I tested this by:

1. Creating a collection from the `ansible-galaxy collections` CLI that is merged to ansible/ansible:devel. The ones mazer produces won't validate correctly anymore.

2. Upload the tar.gz file to the API with `http --form POST :24817/pulp_ansible/galaxy/foo/api/v3/artifacts/collections/ file@greetings_namespace-hello-1.0.0.tar.gz`

3. Verify no errors occurred in the logs

4. Verify that the collection can be seen `http --form POST 24817/pulp/api/v3/content/ansible/collections/`

PR #179, opened 2019-08-21
Move collection version metadata to subtree

cutwater:
This change is needed for compatibility with `ansible-client`.

[noissue]


alikins:
Looks good to me. 


(The nested serializers will allow this to work for read and write correct?  Not that we currently or plan to support PUT/POST on a /collections/{namespace}/{name} at the moment, but theoretically we could?)

cutwater:
> The nested serializers will allow this to work for read and write correct?

Not sure. This solution is only for read. Uploaded collections are mostly read only.

PR #180, opened 2019-08-21
collection batches syncing

fabricio-aguiar:
closes #5323
https://pulp.plan.io/issues/5323

### DIFF
--- a/CHANGES/5323.bugfix
+++ b/CHANGES/5323.bugfix
@@ -1,0 +1,1 @@
+Sync collections in batches.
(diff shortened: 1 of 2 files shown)


fabricio-aguiar:
I change it to work with batches, for assuring the amount of opened files, it seems to work.
From what I see, now we are having problems with wrong urls there:
```bash
pulp_smash.exceptions.TaskReportError: ('Task report /pulp/api/v3/tasks/1e7d13a8-109d-4afa-86a8-b6db76046fc1/ contains a error: {\'code\': None, \'description\': \'Cannot connect to host ansible-galaxy-dev.s3.amazonaws.com:443 ssl:None [Name or service not known]\'
```

fabricio-aguiar:
It is better to solve the issue by looking into the downloader itself.

PR #181, opened 2019-08-22
Enabled pulp-to-pulp syncing

daviddavis:
fixes #5288
https://pulp.plan.io/issues/5288


pulp_ansible/app/galaxy/serializers.py
```diff
@@ -55,16 +55,16 @@ class Meta:
...
     """
     A serializer for a Collection.
     """
 
     name = serializers.CharField()
     namespace = serializers.CharField()
-    version = serializers.CharField()
     href = serializers.SerializerMethodField(read_only=True)
     versions_url = serializers.SerializerMethodField(read_only=True)
+    versions_url = serializers.SerializerMethodField(read_only=True)
```
bmbouter: this is the same as the line before?

pulp_ansible/app/galaxy/serializers.py
```diff
@@ -80,6 +80,33 @@ def get_versions_url(self, obj):
...
+        return "{hostname}/pulp_ansible/galaxy/{path}/api/v2/collections/{namespace}/{name}/".format(
+            path=obj.path,
+            hostname=settings.ANSIBLE_API_HOSTNAME,
+            namespace=obj.namespace,
+            name=obj.name,
+        )
+
+    class Meta:
+        fields = ("name", "namespace", "href", "versions_url")
+        model = CollectionVersion
```
bmbouter: This looks right, but here you can see how it motivates the name `GalaxyCollectionVersionSerializer`. I'm not sure what would be best. I'm ok w/ the change as you've written it also.

pulp_ansible/app/galaxy/serializers.py
```diff
@@ -89,14 +116,26 @@ def get_href(self, obj):
             "versions/{version}/".format(
                 path=obj.path,
                 hostname=settings.ANSIBLE_API_HOSTNAME,
-                namespace=obj.namespace,
-                name=obj.name,
+                namespace=obj.collection.namespace,
```
bmbouter: :+1:

bmbouter:
@daviddavis I left a few questions but overall this looks good. Once Travis is passing I can lgtm and merge it. Thank you for this.

bmbouter:
It also needs a changelog entry, but I don't think Travis got that far.

pulp_ansible/app/galaxy/serializers.py
daviddavis: It should be `Collection`. I've updated it.

daviddavis:
@bmbouter thanks for the review. This should be ready to merge.

PR #192, opened 2019-08-29
Write galaxy-importer contents and docs_blob to CollectionVersion

awcrosby:
Closes: https://pulp.plan.io/issues/5368

### DIFF
--- a/CHANGES/5368.feature
+++ b/CHANGES/5368.feature
@@ -1,0 +1,1 @@
+Write galaxy-importer result of contents and docs_blob into CollectionVersion model
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -1,7 +1,9 @@
 import asyncio
 from gettext import gettext as _
+import json
 import logging
 import math
+import tarfile
 
 from django.core.exceptions import ValidationError
 from django.db import transaction
@@ -119,9 +121,10 @@ def import_collection(
 
     artifact = Artifact.objects.get(pk=artifact_pk)
     log.info("Processing collection from {path}".format(path=artifact.file.path))
+    import_logger = logging.getLogger("pulp_ansible.app.tasks.collection.import_collection")
 
     try:
-        importer_result = process_collection(str(artifact.file.path))
+        importer_result = process_collection(str(artifact.file.path), logger=import_logger)
         collection_info = importer_result["metadata"]
 
         if expected_namespace and expected_namespace != collection_info["namespace"]:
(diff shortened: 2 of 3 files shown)


CHANGES/5368.feature
```diff
@@ -0,0 +1 @@
+Write galaxy-importer result of contents and docs_blob into CollectionVersion model
```
bmbouter: nitpick: For situations when files are concatenated I always put a newline at the end of these.

pulp_ansible/app/tasks/collections.py
```diff
@@ -103,7 +103,12 @@ def import_collection(artifact_pk, repository_pk=None):
             if collection_info[key] is None:
                 collection_info.pop(key)
 
-        collection_version = CollectionVersion(collection=collection, **collection_info)
+        collection_version = CollectionVersion(
+            collection=collection,
+            **collection_info,
+            contents=importer_result["contents"] or [],
```
bmbouter: Instead of duplicated these defaults here is it possible the `galaxy_importer` could return the data and have pulp_ansible not interpret it?

bmbouter: My goal is to avoid a situation where the default value `[]` is specified in multiple places inside pulp_ansible code.

awcrosby: @bmbouter the next release of `galaxy-importer` sets these default values, I added these only to allow this to be merged beforehand and avoid null constraint errors.  I can remove this and we can wait to merge this until after `galaxy-importer` sets these values (will release after ansible 2.9 alpha available on pypi)

awcrosby:
With processing contents, `galaxy-importer` fails import if some things are not present - like valid role and plugin names, and required collection readme - this causes some tests to fail.

`test_mirror_sync()` seems to pull based on https://galaxy.ansible.com/api/v2/collections/?namespace__name=testing and uses softasap.swarm_box with no collection readme (this looks like a prod not test collection, and is not in the testing namespace).

`test_mirror_sync_with_requirements()` seems to pull https://galaxy-dev.ansible.com testing.ansible_testing_content which is not a collection but multi-content repo (a short-lived format scrapped for collections) - this causes the importer to fail with bad role name of `test-role-c`.

bmbouter:
@awcrosby I agree w/ your writeup about the collections used in these tests not being the correct ones and because of that it's causing them to fail with the new validation the galaxy-importer provides.

My question though is: what do you test against? We should have the test mimic what you do because you'll know the right reference data for us to rely on. Does this question make sense?

bmbouter:
@kersommoura fyi also this convo ^

awcrosby:
@bmbouter I think testing against this collection would be good: https://galaxy-dev.ansible.com/testing/k8s_demo_collection ... if you need multiple collections we can get something additional added

alikins:
Works for me.

PR #196, opened 2019-09-04
Add CollectionImport log handler and Validation

bmbouter:
Add add_log_message() facility to the CollectionImport model. This is
called by the log handler which is auto-configured by pulp_ansible with
this commit.

The logger is 'pulp_ansible.app.tasks.collection.import_collection'.

Add validation feature to CollectionImport where the
`expected_namespace`, `expected_name`, and `expected_version` are
specified. When used the import will fail if the tarball does not
contain the same values. These are optional parameters and only
available in the Galaxy V3 API currently.

https://pulp.plan.io/issues/5366
closes #5366

https://pulp.plan.io/issues/5422
closes #5422


bmbouter:
I was able to successfully log into the handler with:

```
    logger = logging.getLogger('pulp_ansible.app.tasks.collection.import_collection')
    logger.info('a logging message')
```

It produced this result in the db:

```
>>> CollectionImport.objects.all()[0].messages                                                                                                                                                                 
[{'time': 1568056517.460521, 'level': 'INFO', 'message': 'a logging message'}]
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -71,22 +71,52 @@ def sync(remote_pk, repository_pk, mirror):
...
+            assert expected_namespace is collection_info["namespace"]
+
+        if expected_name:
+            assert expected_name is collection_info["name"]
+
+        if expected_version:
+            assert expected_version is collection_info["version"]
+    except AssertionError:
+        artifact.delete()
+        raise
```
daviddavis: Any reason to raise a generic exception here?

pulp_ansible/app/tasks/collections.py
```diff
@@ -71,22 +71,52 @@ def sync(remote_pk, repository_pk, mirror):
...
+    repository_pk=None,
+    expected_namespace=None,
+    expected_name=None,
+    expected_version=None,
+):
     """
-    Create a Collection from an uploaded artifact.
+    Create a Collection from an uploaded artifact and optionally validate its expected metadata.
+
+    This task provides optional validation of the `namesapce`, `name`, and `version` metadata
```
daviddavis: s/namesapce/namespace/

pulp_ansible/app/tasks/collections.py
bmbouter: I'm open to other suggestions. I think I mostly shyed away from defining a custom Exception object. What would you recommend?

daviddavis: I think a generic exception is fine actually. No need to overengineer it.

cutwater: I don't think it's fine. It's not fine at all. 

1. You should **NEVER** use assert operator for business logic.
Assertions are disabled if python is called with `-O` flag. 

2. You should **NEVER** compare strings with `is` operator. `is` operator compares objects identity, not value identity. You cannot rely on internal python optimizations and intern strings for comparison.

Correct code:

```
if (
    expected namespace != collection_info["namespace"]
    or expected_name != collection_info["name"]
    or expected_version != collection_info["version"]
):
    artifact.delete()
    raise ValidationErrorClass("Collection metadata doesn't match expected metadata").

Note: The best would be to put artifact deletion in a context manager, to make sure it's always deleted if exception occurs within the scope.
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -71,22 +71,52 @@ def sync(remote_pk, repository_pk, mirror):
...
+        if expected_namespace:
+            assert expected_namespace is collection_info["namespace"]
+
+        if expected_name:
+            assert expected_name is collection_info["name"]
+
+        if expected_version:
+            assert expected_version is collection_info["version"]
+    except AssertionError:
+        artifact.delete()
```
cutwater: Artifact is deleted if it fails validation against expected data, but it's not deleted if process_collection fails.

pulp_ansible/app/tasks/collections.py
bmbouter: Thanks for pointing these things out. I agree and I'll push a fix now.

pulp_ansible/app/tasks/collections.py
bmbouter: Will fix on next push, ty.

pulp_ansible/app/tasks/collections.py
bmbouter: Agreed that case should cause the Artifact to delete also. Will fix now, ty.

pulp_ansible/app/tasks/collections.py
bmbouter: fixed in this push @cutwater lmk what you think.

pulp_ansible/app/tasks/collections.py
bmbouter: fixed in next push.

PR #198, opened 2019-09-06
Port our Collection code, docs, tests to use ansible-galaxy cli

daviddavis:
fixes #5335
https://pulp.plan.io/issues/5335

### DIFF
--- a/pulp_ansible/tests/functional/api/collection/test_consume_content.py
+++ b/pulp_ansible/tests/functional/api/collection/test_consume_content.py
@@ -1,5 +1,7 @@
 # coding=utf-8
-"""Tests that content hosted by Pulp can be consumed by mazer."""
+"""Tests that content hosted by Pulp can be consumed by ansible-galaxy."""
+from os import path
+import tempfile
 import unittest
 
 from pulp_smash import api, cli, config, exceptions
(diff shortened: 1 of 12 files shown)


kersommoura:
@daviddavis, those changes look good to me. I will check again once CI runs.

pulp_ansible/tests/functional/api/collection/test_consume_content.py
```diff
@@ -55,12 +57,15 @@ def test_consume_content(self):
         )
         self.addCleanup(self.client.delete, distribution["_href"])
 
+        collection_dir = tempfile.mkdtemp()
```
kersommoura: We could add a clean up for this dir. Not a blocker.

daviddavis: I figured since it was a temp dir, we wouldn't have to worry about it?

daviddavis: Do we have to cleanup temp directories?

bmbouter: The temp facilities will handle this for us is my understanding.

PR #201, opened 2019-09-12
Fix S3 files not having path attribute

cutwater:
This patch replaces `artifact.file.path` attribute usage, which
may not be implemented for all types of django file storage,
with using django File interface, specifically an `open()` method
to work with artifact as a python file-like object.

DependsOn: ansible/galaxy-importer#29

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -1,13 +1,15 @@
 import asyncio
+import contextlib
 from gettext import gettext as _
 import json
 import logging
 import math
 import tarfile
 
-from django.core.exceptions import ValidationError
 from django.db import transaction
 from galaxy_importer.collection import import_collection as process_collection
+from galaxy_importer.collection import CollectionFilename
+from galaxy_importer.exceptions import ImporterError
 from rq.job import get_current_job
 
 from pulpcore.plugin.models import (
@@ -104,15 +106,15 @@ def import_collection(
         repository_pk (str): Optional. If specified, a new RepositoryVersion will be created for the
             Repository and any new Collection content associated with it.
         expected_namespace (str): Optional. The namespace is validated against the namespace
-            specified in the Collection's metadata. If it does not match a ValidationError is
+            specified in the Collection's metadata. If it does not match a ImporterError is
             raised.
         expected_name (str): Optional. The name is validated against the name specified in the
-            Collection's metadata. If it does not match a ValidationError is raised.
+            Collection's metadata. If it does not match a ImporterError is raised.
         expected_version (str): Optional. The version is validated against the version specified in
-            the Collection's metadata. If it does not match a ValidationError is raised.
+            the Collection's metadata. If it does not match a ImporterError is raised.
 
     Raises:
-        ValidationError: If the `expected_namespace`, `expected_name`, or `expected_version` do not
+        ImporterError: If the `expected_namespace`, `expected_name`, or `expected_version` do not
             match the metadata in the tarball.
 
     """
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -105,46 +107,24 @@ def import_collection(
...
-                msg.format(
-                    expected_version=expected_version, collection_version=collection_info["version"]
-                )
-            )
-    except Exception:
-        artifact.delete()
-        raise
+    artifact = Artifact.objects.get(pk=artifact_pk)
+    filename = None
+    if expected_namespace and expected_name and expected_version:
```
bmbouter: This adjusts it to require all 3 of them which I can see a use case when a person only asserts on the namespace and not version for example. Can we switch this back?

awcrosby: @cutwater @bmbouter if we agree to do that, the importer can validate on only what is passed to it

awcrosby:
FYI in my fork [branch](https://github.com/pulp/pulp_ansible/compare/master...awcrosby:pr/201/worker_pass_file_obj?expand=1) I fixed merge conflicts and changed the version of `galaxy-importer` and this passed CI

pulp_ansible/app/tasks/collections.py
cutwater: @bmbouter I don't have objections, however it'll require update to importer to expect empty fields in `CollectionFilename` tuple.

bmbouter: I agree. I'm going to see if I can work with @awcrosby to coordinate via slack.

pulp_ansible/app/tasks/collections.py
```diff
@@ -422,8 +411,8 @@ class CollectionContentSaver(ContentSaver):
             collection_version = d_content.content
             for d_artifact in d_content.d_artifacts:
                 artifact = d_artifact.artifact
-                with tarfile.open(str(artifact.file.path), "r") as tar:
-                    log.info(_("Reading MANIFEST.json from {path}").format(path=artifact.file.path))
+                with artifact.file.open() as artifact_file, tarfile.open(artifact_file, "r") as tar:
```
awcrosby: This needs to be `with artifact.file.open() as artifact_file, tarfile.open(fileobj=artifact_file, mode="r") as tar:`

setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup, find_packages
 
 requirements = [
-    "galaxy_importer==0.1.3b5",
+    "galaxy_importer==0.1.3b7",
```
awcrosby: @cutwater if this is bumped to `galaxy_importer==0.1.3b8` it should pass ci

pulp_ansible/app/tasks/collections.py
```diff
@@ -120,47 +122,24 @@ def import_collection(
...
-        if expected_name and expected_name != collection_info["name"]:
-            msg = _(
-                "Expected name '{expected_name}', but instead Collection has name "
-                "'{collection_name}'."
-            )
-            raise ValidationError(
-                msg.format(expected_name=expected_name, collection_name=collection_info["name"])
-            )
+        except ImporterError as exc:
+            # NOTE(cutwater): Because this one is expected, probably should wrap into some
```
bmbouter: With the future pulp_ansible work tracked in Redmine, can we convert things like this into Remdine tickets instead of here? I strive to keep planning in one place.

cutwater: It's more like a question, I'm not sure if this has to be fixed or not.

bmbouter: oic. In that case I would say let's not fix it for now and not file it as future work. We can improve this later when we get a bugfix. What do you think?

cutwater: So what do I do with this comment? Remove it?

bmbouter: +1 to removing the comment yes.

pulp_ansible/app/tasks/collections.py
```diff
@@ -120,47 +122,24 @@ def import_collection(
...
-            )
-            raise ValidationError(
-                msg.format(
-                    expected_namespace=expected_namespace,
-                    collection_namespace=collection_info["namespace"],
+    with _artifact_guard(artifact):
+        try:
+            with artifact.file.open() as artifact_file:
+                importer_result = process_collection(
+                    artifact_file, filename=filename, logger=import_logger
```
bmbouter: How does the validation occur here? `expected_namespace`, `expected_name`, and `expected_version` don't seem to be used.

awcrosby: the validation occurs within `galaxy-importer` and raises an error if there is no match: https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/collection.py#L103-L113

bmbouter: That seems to use a filename check, which when used in pulp_ansible will not have the same filename structure as saved in the content addressable storage. Am I reading this right?

awcrosby: Since expected namespace, name and version are passed into this method, the filename structure could be different before this method is called. Those 3 parameters are passed into galaxy-importer then each, if populated, are validated against metadata.

bmbouter: I see now. The filename is formed from the `expected_namespace`, etc on L#125. So this should work for pulp_ansible just like before.

pulp_ansible/app/tasks/collections.py
```diff
@@ -120,47 +122,24 @@ def import_collection(
...
 
-        if expected_name and expected_name != collection_info["name"]:
-            msg = _(
-                "Expected name '{expected_name}', but instead Collection has name "
-                "'{collection_name}'."
-            )
-            raise ValidationError(
-                msg.format(expected_name=expected_name, collection_name=collection_info["name"])
-            )
+        except ImporterError as exc:
```
awcrosby: The docstring should also be `ImporterError`

cutwater: Done

pulp_ansible/app/tasks/collections.py
cutwater: Done

pulp_ansible/app/tasks/collections.py
```diff
@@ -112,55 +114,30 @@ def import_collection(
             the Collection's metadata. If it does not match a ValidationError is raised.
 
     Raises:
-        ValidationError: If the `expected_namespace`, `expected_name`, or `expected_version` do not
+        ImporterError: If the `expected_namespace`, `expected_name`, or `expected_version` do not
```
bmbouter: It's also above in the docstring in more places (per attribute). Removing it from above would also be ok w/ me.

cutwater: Done

PR #206, opened 2019-09-17
Return collection version 'namespace' as a dict (Fixes #5459)

alikins:
https://pulp.plan.io/issues/5459
closes #5459

### DIFF
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -131,6 +131,14 @@ class CollectionMetadataSerializer(serializers.ModelSerializer):
         )
 
 
+class CollectionNamespaceSerializer(serializers.Serializer):
+    """
+    A serializer for a Collection Version namespace field.
+    """
+
+    name = serializers.CharField(source="namespace")
+
+
 class CollectionVersionSerializer(CollectionVersionListSerializer):
     """
     A serializer for a CollectionVersion.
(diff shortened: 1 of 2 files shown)


bmbouter:
@alikins This looks like it's on the right track. It will need a CHANGES changelog entry though for it to pass the CI.

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -131,6 +131,13 @@ class Meta:
         )
 
 
+class CollectionNamespaceSerializer(serializers.Serializer):
+    """
+    A serializer for a Collection Version namespace field.
+    """
+    name = serializers.CharField(max_length=32, source="namespace")
```
cutwater: For serializer that doesn't actually validate data `max_length` is not necessary.

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -141,6 +148,7 @@ class CollectionVersionSerializer(CollectionVersionListSerializer):
     download_url = serializers.SerializerMethodField()
 
     metadata = CollectionMetadataSerializer(source="*")
+    namespace = CollectionNamespaceSerializer(source="*")
```
cutwater: Minor: SerializerMethodField would've been more than enough.

pulp_ansible/app/galaxy/v3/serializers.py
alikins: updated

pulp_ansible/app/galaxy/v3/serializers.py
alikins: https://github.com/pulp/pulp_ansible/compare/master...alikins:collection_ver_5459_method is also available, but Serializer class seems more pulp style.

PR #213, opened 2019-09-19
add new upload serializer

fabricio-aguiar:
Required PR: pulp/pulpcore-plugin#122
closes #5462
https://pulp.plan.io/issues/5462

### DIFF
--- a/CHANGES/5462.misc
+++ b/CHANGES/5462.misc
@@ -1,0 +1,1 @@
+Add upload functionality to the ansible contents endpoints.
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -1,19 +1,22 @@
 from gettext import gettext as _
 
 from django.conf import settings
+from django.db import transaction
+from galaxy_importer.collection import CollectionFilename, import_collection
 from rest_framework import serializers
 
 from pulpcore.plugin.serializers import (
     ContentChecksumSerializer,
     ModelSerializer,
     RemoteSerializer,
-    SingleArtifactContentSerializer,
+    SingleArtifactContentUploadSerializer,
     RepositoryVersionDistributionSerializer,
 )
 
 from .models import (
     AnsibleDistribution,
     AnsibleRemote,
+    Collection,
     CollectionImport,
     CollectionVersion,
     CollectionRemote,
@@ -23,7 +26,7 @@ from .models import (
 from pulp_ansible.app.tasks.utils import parse_collections_requirements_file
 
 
-class RoleSerializer(SingleArtifactContentSerializer):
+class RoleSerializer(SingleArtifactContentUploadSerializer):
     """
     A serializer for Role versions.
     """
@@ -54,11 +57,9 @@ class RoleSerializer(SingleArtifactContentSerializer):
         return data
 
     class Meta:
-        fields = tuple(set(SingleArtifactContentSerializer.Meta.fields) - {"relative_path"}) + (
-            "version",
-            "name",
-            "namespace",
-        )
+        fields = tuple(
+            set(SingleArtifactContentUploadSerializer.Meta.fields) - {"relative_path"}
+        ) + ("version", "name", "namespace")
         model = Role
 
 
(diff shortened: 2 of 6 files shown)


pulp_ansible/app/serializers.py
```diff
@@ -270,17 +271,65 @@ class CollectionVersionSerializer(SingleArtifactContentSerializer, ContentChecks
...
 
-    repository = serializers.URLField(
-        help_text=_("The URL of the originating SCM repository."), allow_blank=True, max_length=128
-    )
-
     tags = TagNestedSerializer(many=True, read_only=True)
 
     version = serializers.CharField(help_text=_("The version of the collection."), max_length=32)
 
+    def validate(self, data):
```
mdellweg: Please consider performing that validation work in `deferred_validate`.

See: https://pulp.plan.io/issues/5403?pn=1#note-9

pulp_ansible/app/serializers.py
```diff
@@ -270,17 +271,65 @@ class CollectionVersionSerializer(SingleArtifactContentSerializer, ContentChecks
...
+        Raises:
+            rest_framework.serializers.ValidationError: If invalid data
+
+        """
+        data = super().deferred_validate(data)
+        artifact = data["artifact"]
+        with artifact.file.open() as artifact_file:
+            importer_result = import_collection(artifact_file)
+
+        collection_info = importer_result["metadata"]
```
bmbouter: I think the `defered_validate()` should stop here and move the functionality below to a `create()` call on the `CollectionVersionSerializer`. Otherwise it's [validated and created at once in this call](https://github.com/pulp/pulpcore/pull/305/files#diff-96bbb4ed1da9a24fe45e8f0626df8ad4R17), and then it will be "created" again later with its own `save()` [here](https://github.com/pulp/pulpcore/pull/305/files#diff-96bbb4ed1da9a24fe45e8f0626df8ad4R18).

CHANGES/5462.misc
```diff
@@ -0,0 +1 @@
+Add upload functionality to the ansible contents endpoints.
```
bmbouter: Similar request to call out the removal of the on shot upload URL.

pulp_ansible/app/urls.py
```diff
@@ -72,7 +72,7 @@
 ]
 
 urlpatterns = [
-    path("ansible/collections/", CollectionUploadViewSet.as_view({"post": "create"})),
+    path("ansible/collections/", CollectionVersionViewSet.as_view({"post": "create"})),
```
fabricio-aguiar: @bmbouter I'm not removing the url, just changing the viewset

pulp_ansible/app/serializers.py
```diff
@@ -115,47 +116,6 @@ class Meta:
...
-
-    file = serializers.FileField(help_text=_("The Collection tarball."), required=True)
-
-    sha256 = serializers.CharField(
-        help_text=_("An optional sha256 checksum of the uploaded file."),
-        required=False,
-        default=None,
-    )
-
-    expected_namespace = serializers.CharField(
```
bmbouter: We need the user to be able to submit these 3 attributes as the "expected names" to be validated against the data returned from `importer_result = import_collection(file=artifact_file)` line in `deferred_validate()`. To do that we probably still need pulp_ansible to use it's own viewset and serializer that extends the base provided ones.

pulp_ansible/app/urls.py
bmbouter: Ah true. Thank you.

bmbouter:
@fabricio-aguiar you've convinced me that we can't adopt this if we're going to continue to use the `expected_name` because we'd end up bringing back the very code you removed. Should it be closed and it's corresponding issue?

fabricio-aguiar:
> @fabricio-aguiar you've convinced me that we can't adopt this if we're going to continue to use the `expected_name` because we'd end up bringing back the very code you removed. Should it be closed and it's corresponding issue?

agreed!

pulp_ansible/app/serializers.py
mdellweg: It should be sufficient, to include them as write_only. Then they will be available in `deferred_validation`.

mdellweg:
I just tested this, and it looks really good now. The only thing i am not sure about is whether the galaxy v2 viewset / serializer should be reworked, too. Maybe they are just completely different.

https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/serializers.py
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/views.py

CI Fails at the moment, because the Required PR thing does not work, once it is merged.

bmbouter:
@fabricio-aguiar the "Galaxy APIs" are the viewsets that are identified in this [urls.py](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/urls.py) file. The "v2" part are the ViewSets [here](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/urls.py#L26-L34).

If there is a way to refactor the V2 API and have the existing collection install test continue working then I'm +1 on it.

fabricio-aguiar:
> @fabricio-aguiar the "Galaxy APIs" are the viewsets that are identified in this [urls.py](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/urls.py) file. The "v2" part are the ViewSets [here](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/urls.py#L26-L34).
> 
> If there is a way to refactor the V2 API and have the existing collection install test continue working then I'm +1 on it.

I believe we can work on v2 and v3 on another story

bmbouter:
I also agree because those viewsets need to be deduplicated in other ways. Also we're likely going to remove V2 support once the V3 API is able to fully interact with the `ansible-galaxy collections` API.

@mdellweg let us know if you think we're missing an easy opportunity here.

mdellweg:
> I believe we can work on v2 and v3 on another story

... only that you already brought v3 nicely in line:

https://github.com/pulp/pulp_ansible/pull/213/files#diff-0cdb0aa026aae09261274992ee42ede5R42

bmbouter:
@mdellweg agreed

PR #222, opened 2019-10-04
Add ansible collection viewset

daviddavis:
fixes #5520
https://pulp.plan.io/issues/5520


pulp_ansible/app/viewsets.py
```diff
@@ -60,9 +63,33 @@ class RoleViewSet(SingleArtifactContentUploadViewSet):
     filterset_class = RoleFilter
 
 
+class CollectionFilter(BaseFilterSet):
+    """
+    FilterSet for Ansible CollectionVersions.
```
bmbouter: s/CollectionVersions/Collection/ ?

pulp_ansible/app/viewsets.py
```diff
@@ -60,9 +63,33 @@ class RoleViewSet(SingleArtifactContentUploadViewSet):
...
+        model = Collection
+        fields = ["namespace", "name"]
+
+
+class CollectionViewset(NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin):
+    """
+    Viewset for Ansible Collections.
+    """
+
+    endpoint_name = "ansible/collections"
```
bmbouter: Does this still nest under `/pulp/api/v3/` ?

bmbouter:
Can you also rebuild the docs so that the `api.json` schema is generated. Otherwise this viewset won't be available in the rebuilt bindings.

pulp_ansible/app/viewsets.py
daviddavis: Yes

daviddavis:
Yea, I can. Do we have to update api.json every time we modify the API? Is there any way it can be automated?

bmbouter:
I share the desire for automation, but there isn't any that is simple and obvious to me. Travis doesn't author code changes in any way today, so we can't have it do it easily. The only thing I can think of is a git commit hook that grabs the file at commit time. What about something like that?

pulp_ansible/app/viewsets.py
bmbouter: Thanks!

bmbouter:
Also I didn't answer your question: Yes we do have to update the api.json everytime.

daviddavis:
I guess my question is why have api.json in version control instead of generating it at publish time?

bmbouter:
It's a great question, I spent over a day trying to find a better way. To have this not be version control at docs-build time you would need a running Pulp to produce it for you. For `pulpcore` and `pulpcore-plugin` the docs are built on Travis, and in those two environments specifically there is a running Pulp so those don't need to `api.json` checked into version control.

The issue with plugins is that they don't have their docs built on Travis and pushed to `docs.pulpproject.org` instead all the plugins use readthedocs. RTD won't install a python package much less the multi-process setup of Pulp that would be needed. So the workaround is that the api.json is included in the docs themselves and RTD can build referring from that.

To fully solve the problem we'd have to have all docs start building on Travis and then we'd need to find a place to host them.

Any suggestions or ideas are welcome.

PR #228, opened 2019-10-11
Creating a CollectionImport immediately after the task

daviddavis:
Galaxy was polling the CollectionImport which wasn't being created until
the task started. This caused 404s.

fixes #5569

### DIFF
--- a/CHANGES/5569.bugfix
+++ b/CHANGES/5569.bugfix
@@ -1,0 +1,1 @@
+Fix bug where CollectionImport was not being created in viewset causing 404s for galaxy.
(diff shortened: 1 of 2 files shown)


fabricio-aguiar:
towncrier's CHANGE is missing

daviddavis:
@bmbouter I was waiting for @cutwater's feedback. Just want to make sure there was no specific reason that CollectionImport was being created in the task.

bmbouter:
@daviddavis I actually implemented this aspect originally. The only reason was at that time, in the implementation you couldn't do it like this because CollectionImport also had namespace, name, version which you couldn't know in all cases (with pre-uploaded Artifacts) so you could only create it feasibly at task-time. We can still ask @cutwater if you want.

daviddavis:
@bmbouter ah ok. That's good enough for me. I'll update and merge. Thanks.

newswangerd:
Here's the related automation hub issue: https://github.com/ansible/galaxy-dev/issues/124

cutwater:
LGTM, thanks for fixing that.

PR #231, opened 2019-10-15
Fix invalid reference to _id

daviddavis:
fixes #5572
https://pulp.plan.io/issues/5572


pulp_ansible/app/tasks/collections.py
```diff
@@ -180,7 +180,7 @@ def import_collection(
 
         if repository_pk:
             repository = Repository.objects.get(pk=repository_pk)
-            content_q = CollectionVersion.objects.filter(_id=collection_version.pk)
+            content_q = CollectionVersion.objects.filter(pk=collection_version.pk)
```
fabricio-aguiar: `pk` is much better!

PR #233, opened 2019-10-18
Allow users to update `deprecated` for collections

fabricio-aguiar:
closes #5577
https://pulp.plan.io/issues/5577

### DIFF
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -1,3 +1,4 @@
+from gettext import gettext as _
 from django.conf import settings
 from rest_framework.reverse import reverse
 from rest_framework import serializers, relations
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -29,6 +29,13 @@ class Meta:
         )
         model = models.CollectionVersion
 
+    def update(self, obj, validated_data):
+        """Update Collection."""
+        collection_data = validated_data.pop("collection")
+        collection = models.Collection.objects.filter(pk=obj.collection.pk)
+        collection.update(**collection_data)
```
bmbouter: I think we only want to allow the single field by name. I'm concerned about content becoming "mutable" when we just want that 1 field to be mutable.

PR #236, opened 2019-10-18
Problem: CI with old plugin-template fails

mikedep333:
due to PULP_SETTINGS being unset for multiple commands like django-admin.

Solution: Set PULP_SETTINGS in .travis.yml

re: #5560
PULP_SETTINGS environment variable does not work
https://pulp.plan.io/issues/5560

[noissue]


fabricio-aguiar:
thanks @mikedep333

PR #237, opened 2019-10-18
Add the ability to set a certification status

fabricio-aguiar:
closes #5579
https://pulp.plan.io/issues/5579


pulp_ansible/app/models.py
```diff
@@ -139,7 +144,9 @@ class CollectionVersion(Content):
     version = models.CharField(max_length=32, editable=False)
 
     is_highest = models.BooleanField(editable=False, default=False)
-    is_certified = models.BooleanField(default=False)
+    certification = models.CharField(
+        max_length=13, choices=CERTIFICATION_CHOICES, default="not_certified"
```
newswangerd: Can you change the default to `needs_review`?

fabricio-aguiar: sure!

PR #238, opened 2019-10-22
Collection V3 Tests

ironfroggy:
Added collection V3 tests for publishing, listing, details, versions, and artifact download.


pulp_ansible/tests/functional/api/collection/v3/test_collection.py
```diff
@@ -0,0 +1,236 @@
...
+    response = pulp_client.using_handler(api.json_handler).get(url)
+    return response
+
+
+@pytest.fixture(scope='session')
+def pulp_client():
+    cfg = config.get_config()
+    delete_orphans(cfg)
+    client = api.Client(cfg)
+    headers = cfg.custom['headers']
```
rochacbruno: What if `settings,json` has no `custom['headers']`?

Or we do a `cfg.custom.get("headers", "some default value")`

or

```python
try:
    headers = cfg.custom['headers']
except KeyError:
    pytest.exit("pulp-smash settings file is missing custom['headers'] attribute.")
```

setup.py
```diff
@@ -31,6 +31,7 @@
             "flake8-quotes",
             "flake8-tuple",
             "mock",
+            "orionutils",
```
bmbouter: Do we want this to be a runtime dependency? I see it is only used in tests.

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
```diff
@@ -0,0 +1,281 @@
...
+)
+from pulp_ansible.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+
+from orionutils.generator import build_collection
+
+
+logger = logging.getLogger(__name__)
+
+
+def upload_handler(client, response):
```
kersommoura: This one could be added to pulp-smash itself. if we think other tests will re-use this.

bmbouter:
@ironfroggy is there an option to have the tests omit asserting on download_url somehow? Or do you need to wait for us to switch that to being an absolute URL?

bmbouter:
@ironfroggy you read my mind! Let's do that.

Also note the requirements files got reworked underneath you, now they are broken into a few files for dependencies of various types. LMK if the conflict resolution isn't clear.

daviddavis:
This looks great! Thanks for working on it. Any chance we can squash some or all of the commits together?

PR #239, opened 2019-10-23
Depend on pulpcore directly instead of pulpcore-plugin

dralley:
Required PR: https://github.com/pulp/pulpcore/pull/347

re: #5580
https://pulp.plan.io/issues/5580


dralley:
This is a half-step, I skipped all the containers stuff and changed as much as I could otherwise.

PR #246, opened 2019-11-01
Use latest galaxy_importer which now uses latest ansible

awcrosby:
Uses the latest galaxy_importer on pypi, currently `0.1.3` which uses the latest `ansible`, currently `2.9.0`


bmbouter:
Thanks!

PR #251, opened 2019-11-15
Updating pulp_ansible to work with pulpcore 3.0 RC8

daviddavis:
[noissue]


pulp_ansible/tests/functional/constants.py
```diff
@@ -82,7 +82,7 @@
 
 ANSIBLE_GALAXY_COLLECTION_URL_V2 = urljoin(GALAXY_ANSIBLE_BASE_URL, "api/v2/collections/")
 
-ANSIBLE_COLLECTION_FIXTURE_URL_V2 = urljoin(ANSIBLE_GALAXY_COLLECTION_URL_V2, NAMESPACE_TESTING)
+ANSIBLE_COLLECTION_FIXTURE_URL_V2 = urljoin(ANSIBLE_GALAXY_COLLECTION_URL_V2, "geerlingguy/k8s")
```
daviddavis: Looks like there's not a namespace filter for the collections endpoint in v2, so I changed it to use `geerlingguy/k8s` instead.

PR #254, opened 2019-11-19
Fix 500 error on /pulp/api/v3/

daviddavis:
fixes #5748
https://pulp.plan.io/issues/5748


daviddavis:
This fix seems to have revealed another error when attempting to generate the api schema doc:

```
drf_yasg.errors.SwaggerGenerationError: cannot instantiate nested serializer as Parameter
```

daviddavis:
I set the nested serializers to readonly which seems to have fixed the problem ^

PR #258, opened 2019-11-21
Updating tests for initial repo version change

daviddavis:
[noissue]

PR #260, opened 2019-11-22
Update from plugin template

bmbouter:
[noissue]


bmbouter:
I'm going to admin merge because this won't pass without https://github.com/pulp/pulp_ansible/pull/258 merged which won't pass without this.

PR #274, opened 2020-01-21
Port CLI test to be bindings based

bmbouter:
https://pulp.plan.io/issues/5930
closes #5930


bmbouter:
Thanks @fabricio-aguiar !

@daviddavis I'm going to merge now, but take a look and let me know if you see anything else I should do. I'm thinking if this is good we could move https://pulp.plan.io/issues/5929

PR #276, opened 2020-01-24
Fix collection upload and add a test for it

bmbouter:
https://pulp.plan.io/issues/5905
closes #5905

https://pulp.plan.io/issues/5931
closes #5931

### DIFF
--- a/pulp_ansible/app/galaxy/views.py
+++ b/pulp_ansible/app/galaxy/views.py
@@ -33,7 +33,7 @@ class GalaxyVersionView(views.APIView):
         """
         Return a response to the "GET" action.
         """
-        api_info = {"available_versions": {"v1": "/v1/"}, "current_version": "v1"}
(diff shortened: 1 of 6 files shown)


pulp_ansible/app/galaxy/views.py
```diff
@@ -154,10 +154,7 @@ def post(self, request, path):
         )
         serializer.is_valid(raise_exception=True)
 
-        expected_digests = {"sha256": serializer.validated_data["sha256"]}
-        artifact = Artifact.init_and_validate(
-            serializer.validated_data["file"], expected_digests=expected_digests
```
fabricio-aguiar: why we are not validating the checksum anymore?

bmbouter: We can't anymore because the ansible-galaxy CLI doesn't send it. If we continue to require it then user's couldn't upload for many months since ansible 2.9 would be broken and 2.10 isn't coming out for a long time.

PR #279, opened 2020-02-01
Declare compatability with pulpcore==3.1

bmbouter:
[noissue]


bmbouter:
@daviddavis I'm going to admin merge this to make sure we're compatible with the latest pulpcore bits. Let me know if you think we should do something different in code (or process).

PR #281, opened 2020-02-02
Next release

bmbouter:


bmbouter:
@daviddavis I'm pushing this beta so the new bits can be available on pypi. Let me know what you think.

PR #282, opened 2020-02-10
Update pulpcore requirement

fabricio-aguiar:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import setup, find_packages
 requirements = [
     "galaxy_importer",
     "packaging",
-    "pulpcore>=3.0.<3.2",
+    "pulpcore>=3.0.<3.3",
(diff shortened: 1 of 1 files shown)


bmbouter:
I think we need to leave the master branch testable against pulpcore:master which uses `3.2.0.dev`currently. When we go to GA we'll end up setting this just before release time on the release branch like this PR would. I think we should close this PR. What do you all think?

fabricio-aguiar:
> I think we need to leave the master branch testable against pulpcore:master which uses `3.2.0.dev`currently. When we go to GA we'll end up setting this just before release time on the release branch like this PR would. I think we should close this PR. What do you all think?

I just followed this PR https://github.com/pulp/pulp-certguard/pull/40 
but it not seems to be a problem in pulp_ansible, I'm ok to close it

PR #292, opened 2020-02-29
Adds scale testesting tools and docs

bmbouter:
This adds four scale testing tools:

* `generate_collections.py` which generates collections in parallel
  with Pulp workers.

* `fast_load_collections.py` which loads collections in parallel from a
  filesystem with Pulp workers.

* `create_repos_with_collections.py` which creates repositories and
  repository versions containing CollectionVersions already in Pulp.

* `promote.py` will select a content unit randomly and add it to N
  randomly selected AnsibleRepository objects. This is designed to
  benchmark adding new content to many repositories.

This also adds a new `testing` section in the docs which identifies some
of the different functional tests we have including these tools.

https://pulp.plan.io/issues/6272
closes #6272


pulp_ansible/app/tasks/test_tasks.py
```diff
@@ -0,0 +1,94 @@
...
+    # pydevd_pycharm.settrace('localhost', port=29437, stdoutToServer=True, stderrToServer=True)
+    random_collection = CollectionVersion.objects.order_by('?').first()
+    repos_to_dispatch = []
+    locks = []
+    for repo in AnsibleRepository.objects.all():
+        repos_to_dispatch.append(repo.pk)
+        locks.append(repo)
+        if len(repos_to_dispatch) == repos_per_task:
+            task_args = (random_collection.pk, repos_to_dispatch)
+            enqueue_with_reservation(promote_content, locks, args=task_args)
```
alikins: promote_content task being recursive reads a little odd to me, but I suppose it is to pass the locked objects down to each sub-promote?

(is this going to create one too level transaction? Not sure where the transaction entry/exit points would be)

pulp_ansible/app/tasks/test_tasks.py
```diff
@@ -0,0 +1,94 @@
...
+    if repos_to_dispatch:
+        task_args = (random_collection.pk, repos_to_dispatch)
+        enqueue_with_reservation(add_content_to_repositories, locks, args=task_args)
+
+
+def add_content_to_repositories(collection_pk, repositories_pks):
+    for repository in AnsibleRepository.objects.filter(pk__in=repositories_pks):
+        with repository.new_version() as new_version:
+            qs = CollectionVersion.objects.filter(pk=collection_pk)
+            new_version.add_content(qs)
```
alikins: I wonder if Repository.add_content(qs) for a qs of CollectionVersions may be hitting the weakpoints of CollectionVersion (ansible_collectionversion). Ie that each row has a lot of data and a lot of it is in jsonfields and can be a pretty slow to read fully from the db  (ie, `select * from ansible_collectionversion` takes minutes).

bmbouter: We merged a performance improvement this morning, and you can see this query analyzed here: https://explain.depesz.com/s/Wi2Q

What is unexpected is that at step 15 is doing Seq Scan, when there is an index available for it. Then I realize though that it's returning enough rows that I think it knows its more optimal to do Seq Scan. When I query for content by pk alone and not as part of `add_content(qs)` it uses IndexScan in that case.

pulp_ansible/app/tasks/test_tasks.py
bmbouter: This was a bug, it's been updated. Thanks for commenting!

docs/testing.rst
```diff
@@ -0,0 +1,47 @@
...
+The Galaxy V2 and V3 APIs have tests.
+
+* ``pulp_ansible.tests.functional.api.collection.v2`` - The V2 collection API
+
+* ``pulp_ansible.tests.functional.api.collection.v3`` - The V3 collection API
+
+* ``pulp_ansible.tests.functional.api.role`` - The V2 roles API
+
+
+Scale Testing Tools
```
fabricio-aguiar: Nice!

daviddavis:
LGTM too. Agreed on maybe setting up a nightly performance test. Can be a separate issue/PR.

bmbouter:
I agree these would be great to add periodically run and track that data over time. We need to find a good environment to run it in.

PR #294, opened 2020-03-04
Including file type extension

fabricio-aguiar:
https://pulp.plan.io/issues/6223
closes #6223


CHANGES/6223.bugfix
```diff
@@ -0,0 +1 @@
+Including file type extension
```
bmbouter: Can this call out the upload case as when the user would experience it?

fabricio-aguiar: sure!

pulp_ansible/app/galaxy/views.py
```diff
@@ -157,8 +157,10 @@ def post(self, request, path):
         artifact = Artifact.init_and_validate(serializer.validated_data["file"])
         artifact.save()
 
+        filename = serializer.validated_data["file"].name
```
bmbouter: One of the concerns is about falsified content so we can't rely on this user provided data. We can rely that the name should always be `namespace-name-version.tar.gz` like the example here https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#id2

Maybe we need another helper function like this one? https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/models.py#L166 but for the filename generation.

fabricio-aguiar: the extension will always be `tar.gz`?

pulp_ansible/app/tasks/collections.py
```diff
@@ -169,9 +171,7 @@ def import_collection(
         collection_version.save()  # Save the FK updates
 
         ContentArtifact.objects.create(
-            artifact=artifact,
-            content=collection_version,
-            relative_path=collection_version.relative_path,
```
fabricio-aguiar: if it is always `tar.gz`
we only have to do:
```python
relative_path=f"{collection_version.relative_path}.tar.gz",
```

daviddavis: That's the solution I had in mind rather that using the filename of the file uploaded by the user.

daviddavis: Also, the relative_path property or method is only used here. Seems like we should probably update it.

bmbouter: I was also thinking this. To check if it's safe is it used elsewhere in the code at all? Also does https://github.com/ansible/galaxy-dev/ or https://github.com/ansible/galaxy-importer/ use it? If not then I agree completely.

daviddavis: Good question. I don't see it when I do a code search.

fabricio-aguiar: changing the property made 3 tests fail:
```
------------------------------------ Captured stdout call -------------------------------------
- Collection pulp.etkrvtsdjqgacynfmdrhgzojtx was created successfully
Created collection for pulp.etkrvtsdjqgacynfmdrhgzojtx at /tmp/tmpydr3m36u/pulp-etkrvtsdjqgacynfmdrhgzojtx-1.0.0.tar.gz
Publishing collection artifact '/tmp/tmpydr3m36u/pulp-etkrvtsdjqgacynfmdrhgzojtx-1.0.0.tar.gz' to cmd_arg http://pulp3-source-fedora30.localhost.example.com/pulp_ansible/galaxy/83e92be9-e496
-4f26-9eed-f4b805ce7ed9/api
Collection has been published to the Galaxy server cmd_arg http://pulp3-source-fedora30.localhost.example.com/pulp_ansible/galaxy/83e92be9-e496-4f26-9eed-f4b805ce7ed9/api
Waiting until Galaxy import task http://pulp3-source-fedora30.localhost.example.com/pulp_ansible/galaxy/83e92be9-e496-4f26-9eed-f4b805ce7ed9/api/v2/collection-imports/9e1a6973-0329-419b-8f30
-064a2bc5df10 has completed
/
------------------------------------ Captured stderr call -------------------------------------
ERROR! Error when getting import task results at http://pulp3-source-fedora30.localhost.example.com/pulp_ansible/galaxy/83e92be9-e496-4f26-9eed-f4b805ce7ed9/api/v2/collection-imports/9e1a697
3-0329-419b-8f30-064a2bc5df10 (HTTP Code: 404, Message: Not Found Code: Unknown)
==================== 3 failed, 32 passed, 10 skipped in 143.72s (0:02:23) =====================
```

daviddavis: Do you have any more context? The relative_path should be the entire filename including the extension so it's strange that tests are failing.

fabricio-aguiar: the log above is a stderr for `ansible-galaxy`, all the 3 tests that used it failed

daviddavis: The failing tests are because this only adds the file extension on uploaded collection versions here. It also needs to be added [to synced collection versions](https://github.com/pulp/pulp_ansible/blob/8c6c54061eee21015bd19ef2111499f0549cae4f/pulp_ansible/app/tasks/collections.py#L284-L288) or else they will also be missing their file extensions too. Once I did that, I was able to set the file extension on the `relative_path` property and all the tests are passing now: https://github.com/pulp/pulp_ansible/pull/295.

fabricio-aguiar: should I close this PR in favor of https://github.com/pulp/pulp_ansible/pull/295 ?

daviddavis: No, we still need the migration. I was hoping you could work on that and I could just close out #295 ?

fabricio-aguiar: Ok, I can do that

CHANGES/6223.bugfix
```diff
@@ -0,0 +1 @@
+Including file type extension when uploading collections
```
bmbouter: Can this also indicate that it fixes the user's busted database with a data migration.

PR #296, opened 2020-03-10
Fix extension migration

fabricio-aguiar:
[noissue]


bmbouter:
The code looks great, thanks @fabricio-aguiar 

To have the issue associated in Redmine can we associate it with https://pulp.plan.io/issues/6223 and add a `6223.misc` entry to indicate it was fixed pre-release. Does that make sense to do?

PR #300, opened 2020-03-18
Get bindings host from pulp-smash

fabricio-aguiar:
https://pulp.plan.io/issues/6339
closes #6339

### DIFF
--- a/pulp_ansible/tests/functional/utils.py
+++ b/pulp_ansible/tests/functional/utils.py
@@ -4,7 +4,7 @@ from functools import partial
 from unittest import SkipTest
 from time import sleep
 
-from pulp_smash import api, selectors
+from pulp_smash import api, config, selectors
 from pulp_smash.pulp3.utils import (
     gen_remote,
     gen_repo,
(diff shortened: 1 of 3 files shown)


pulp_ansible/tests/functional/utils.py
```diff
@@ -32,6 +32,7 @@
 
 
 configuration = Configuration()
+configuration.host = config.get_config().get_base_url()
 configuration.username = "admin"
```
bmbouter: I see the pulp smash config also contains the username and password can this PR also have those configured from the settings too?

If we do that, with multiple configuration points coming from the smash config what about making a new function that will effectively return the fully configured `Configuration()` object that all bindings could be instantiated with?

What do you think?

fabricio-aguiar: sounds good, I will do that

bmbouter: @fabricio-aguiar great, I can re-review if you ping me when it's ready.

fabricio-aguiar: done!

bmbouter:
Once the pulp-smash release happens today this will be unblocked to merge. The only thing I want to request is a pulp_ansible specific issue in Redmine and have the changelog refer to that. Would you be willing to do that?

bmbouter:
Well that release just happened :)  https://github.com/pulp/pulp_ansible/pull/300#issuecomment-603436915

fabricio-aguiar:
trying to close and reopen to solve travis bug

PR #301, opened 2020-03-19
Bindings installation on dev environment

fabricio-aguiar:
https://pulp.plan.io/issues/6221
ref #6221

### DIFF
--- a/docs/bindings.rst
+++ b/docs/bindings.rst
@@ -51,3 +51,26 @@ as a GET parameter. For example for ``pulp_ansible`` only endpoints use a query 
 The schema can then be used as input to the openapi-generator-cli. The documentation on getting
 started with openapi-generator-cli is available on
 `openapi-generator.tech <https://openapi-generator.tech/#try>`_.
+
+
+Generating a client on dev environment
+--------------------------------------
+
+Pulp dev environment provided by `ansible-pulp <https://github.com/pulp/ansible-pulp>`_
+introduces a set of useful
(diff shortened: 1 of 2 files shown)


CHANGES/6221.doc
```diff
@@ -0,0 +1 @@
+Documented bindings installation on dev environment
```
bmbouter: Unfortunately I think we need to only have a project refer to issues in Redmine that are filed under that project's tracker. The reason is at release time when one issue is used in multiple plugins the redmine state shows it as CLOSED - CURRENT RELEASE while the other plugins are still unreleased.

To resolve, please file a one-liner issues describing the lack of docs as a defect against Pulp-ansible in Redmine and then use that issue number for this please. It's kind of inconvenient to do, but I think it's best for our users.

docs/bindings.rst
```diff
@@ -51,3 +51,24 @@ as a GET parameter. For example for ``pulp_ansible`` only endpoints use a query
 The schema can then be used as input to the openapi-generator-cli. The documentation on getting
 started with openapi-generator-cli is available on
 `openapi-generator.tech <https://openapi-generator.tech/#try>`_.
+
+
+Generating a client on dev environment
+--------------------------------------
+
+Pulp dev environment provided by `ansible-pulp <https://github.com/pulp/ansible-pulp>`_
+introduces a set of useful `aliases <https://github.com/pulp/ansible-pulp/tree/master/roles/pulp-devel#aliases>`_, like `pbindings`.
```
bmbouter: Can this line be line wrapped under 100 chars? I missed this before sorry.

PR #303, opened 2020-04-30
Release 0.2.0b12

fao89:
Solving conflict due `galaxy_ng` and `pulp_container` requiring `pulpcore` 3.3
```
"There are incompatible versions in the resolved dependencies:", 
"  pulpcore==3.3.0 (from -r requirements.in (line 1))", 
"  pulpcore<3.3,>=3.0 (from pulp-ansible==0.2.0b11->-r requirements.in (line 2))", 
"  pulpcore<3.4,>=3.0 (from galaxy-ng==4.2.0a5->-r requirements.in (line 3))", 
"  pulpcore<3.4,>=3.3 (from pulp-container==1.3.0->-r requirements.in (line 4))"],
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,25 @@ Changelog
 
 .. towncrier release notes start
 
+0.2.0b12 (2020-04-30)
+=====================
+
+Improved Documentation
+----------------------
+
+- Documented bindings installation on dev environment
+  `#6390 <https://pulp.plan.io/issues/6390>`_
+
+
+Misc
+----
+
+- `#6391 <https://pulp.plan.io/issues/6391>`_
+
+
+----
+
+
 0.2.0b11 (2020-03-13)
 =====================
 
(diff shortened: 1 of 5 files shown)


bmbouter:
Usually a commit like this comes as part of a release, but it would need a towncrier commit and also a version bump of the next version. It's typically 3 commits, here's a link to the last one as an example: https://github.com/pulp/pulp_ansible/pull/299

It roughly follow these docs here:  https://pulp.plan.io/projects/pulp/wiki/Pulp3_Release_Guide

Would you want to make the 0.2.0b12 release?

To share some context about the conventions of the state of master, generally I recommend we leave the `.dev` and leave pulpcore only with it's minimum version as a general practice. To share some of the reasoning about why, the `.dev` indicates source installs that they have a source install because only released versions don't have the `.dev`. For the restriction of the pulpcore we only do that at release time so that the master version of a plugin doesn't need to be changed/updated even as pulpcore's master branch gets newer versions over time.

PR #305, opened 2020-05-08
Enabled dispatching customization for importing collections

fao89:
https://pulp.plan.io/issues/6673
closes #6673

### DIFF
--- a/pulp_ansible/app/galaxy/mixins.py
+++ b/pulp_ansible/app/galaxy/mixins.py
@@ -1,0 +1,33 @@
+import shutil
+from tempfile import NamedTemporaryFile
+
+from django.conf import settings
+from pulpcore.plugin.tasking import enqueue_with_reservation
+
+from pulp_ansible.app.tasks.collections import import_collection
+
+
+class UploadGalaxyCollectionMixin:
+    """
+    Provides a method that dispatches a task.
+    """
+
+    def _create_temp_file(self, validated_file):
+        """
+        Create the temporary file.
+        """
+        with NamedTemporaryFile(dir=settings.WORKING_DIRECTORY, delete=False) as temp_import_file:
+            shutil.copyfileobj(validated_file, temp_import_file)
+            return temp_import_file.name
+
+    def _dispatch_import_collection_task(self, repository, temp_file_path, **kwargs):
+        """
+        Dispatch a Import Collection creation task.
+        """
+        locks = []
+        kwargs["temp_file_path"] = temp_file_path
+        if repository:
+            locks.append(repository)
+            kwargs["repository_pk"] = repository.pk
+
+        return enqueue_with_reservation(import_collection, locks, kwargs=kwargs)
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -2,7 +2,6 @@ from datetime import datetime
 from gettext import gettext as _
 import semantic_version
 
-from django.db import IntegrityError
 from django.shortcuts import get_object_or_404
 from django.utils.dateparse import parse_datetime
 from drf_yasg import openapi
@@ -17,7 +16,6 @@ from rest_framework import viewsets
 from pulpcore.plugin.exceptions import DigestValidationError
 from pulpcore.plugin.models import Artifact, Content, ContentArtifact
 from pulpcore.plugin.serializers import AsyncOperationResponseSerializer
-from pulpcore.plugin.tasking import enqueue_with_reservation
 from rest_framework.reverse import reverse
 
 from pulp_ansible.app.galaxy.v3.exceptions import ExceptionHandlerMixin
@@ -31,8 +29,8 @@ from pulp_ansible.app.serializers import (
     CollectionOneShotSerializer,
     CollectionImportDetailSerializer,
 )
-from pulp_ansible.app.tasks.collections import import_collection
 
+from pulp_ansible.app.galaxy.mixins import UploadGalaxyCollectionMixin
 from pulp_ansible.app.viewsets import CollectionVersionFilter
 
 
@@ -109,7 +107,9 @@ class CollectionViewSet(
         return Response(serializer.data)
 
 
-class CollectionUploadViewSet(ExceptionHandlerMixin, viewsets.GenericViewSet):
+class CollectionUploadViewSet(
+    ExceptionHandlerMixin, viewsets.GenericViewSet, UploadGalaxyCollectionMixin
+):
     """
     ViewSet for Collection Uploads.
     """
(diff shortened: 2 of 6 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -178,6 +180,17 @@ def create(self, request, path):
         }
         return Response(data, status=http_status.HTTP_202_ACCEPTED)
 
+    def _dispatch_import_collection_task(self, repository, temp_file_path, **kwargs):
```
bmbouter, line 183: Since the subclasser isn't in control of being able to pass additional kwargs, I don't believe this should get `**kwargs` in its signature.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -178,6 +180,17 @@ def create(self, request, path):
...
+        """
+        Dispatch a Import Collection creation task.
+        """
+        kwargs["temp_file_path"] = temp_file_path
```
bmbouter, line 187: This method would create its own `kwargs = {}`.

pulp_ansible/app/galaxy/views.py
```diff
@@ -154,18 +155,25 @@ def post(self, request, path):
...
-        kwargs = {"artifact_pk": artifact.pk}
-        if distro.repository:
-            locks.append(distro.repository)
-            kwargs["repository_pk"] = distro.repository.pk
-
-        async_result = enqueue_with_reservation(import_collection, locks, kwargs=kwargs)
+        async_result = self._dispatch_import_collection_task(distro.repository, temp_file_path)
         return OperationPostponedResponse(async_result, request)
 
+    def _dispatch_import_collection_task(self, repository, temp_file_path, **kwargs):
```
bmbouter, line 166: Can we dedpulicate this from being defined in multiple places with a mixin object? Let me know if this suggestion needs more details to implement.

pulp_ansible/app/tasks/collections.py
```diff
@@ -82,7 +82,7 @@ def sync(remote_pk, repository_pk, mirror):
 
 
 def import_collection(
-    artifact_pk,
+    temp_file_path,
```
bmbouter, line 85: :+1:

pulp_ansible/app/tasks/collections.py
```diff
@@ -119,21 +119,20 @@ def import_collection(
...
+            importer_result = process_collection(
+                artifact_file, filename=filename, logger=import_logger
+            )
 
-        except ImporterError as exc:
-            log.info(f"Collection processing was not successfull: {exc}")
-            raise
+    except ImporterError as exc:
+        log.info(f"Collection processing was not successfull: {exc}")
+        os.remove(temp_file_path)
```
bmbouter, line 134: We need to remove the temp file in all cases not only `ImporterError`. To do that I'd suggest replacing `ImporterError` with the more general `Exception`.

bmbouter: This comment is really for L#132

pulp_ansible/app/galaxy/v3/views.py
bmbouter: I'm wrong about this because there is one case where you need to pass along these `expected_version`, `expected_namespace`, and `expected_name`. Please disregard.

pulp_ansible/app/galaxy/v3/views.py
bmbouter: Disregard, this caller needs to build the `kwargs` for the cases when extra args need to be passed.

pulp_ansible/app/galaxy/views.py
```diff
@@ -154,18 +155,25 @@ def post(self, request, path):
         )
         serializer.is_valid(raise_exception=True)
 
-        artifact = Artifact.init_and_validate(serializer.validated_data["file"])
-        artifact.save()
+        temp_file = serializer.validated_data["file"]
```
bmbouter, line 158: What path does this get? Is it alread inside `/var/lib/pulp`/?

fao89: `/var/lib/pulp/tmp`

CHANGES/6673.feature
```diff
@@ -0,0 +1 @@
+Enabled dispatching customization for importing collections
```
bmbouter, line 1: I think this should be a `misc` entry because user's capabilities are not changed by it.

pulp_ansible/app/galaxy/mixins.py
```diff
@@ -0,0 +1,20 @@
+from pulpcore.plugin.tasking import enqueue_with_reservation
+
+from pulp_ansible.app.tasks.collections import import_collection
+
+
+class UploadGalaxyCollectionMixin:
```
bmbouter, line 6: :+1:

pulp_ansible/app/galaxy/views.py
bmbouter: What happens if we use `temp_file_path = serializer.validated_data["file"]` rather than using `shutil` at all? I suspect it may not work, but I wanted to ask.

fao89: it was my original approach,
```python
serializer.validated_data["file"] == PulpTemporaryUploadedFile == tempfile.NamedTemporaryFile
```
By the time the task is executed, the file is gone

bmbouter: I see, yes that is what I was concerned about. Then just using `temp_file_path` as I was thinking won't work.

There is a small (although unlikely) issue with your code building the name instead of having tempfile pick the file name instead. I believe `tempfile` contains code which guarantees to be free of race conditions. Here's the gist of what I was thinking.

```
with tempfile.NamedTemporaryFile(dir=settings.WORKING_DIRECTORY, delete=False) as temp_import_file:
    shutil.copyfileobj(serializer.validated_data["file"], temp_import_file)
    temp_file_path = temp_import_file.name   # I forget the attribute name but it's like name or named
```

What do you think about this?

Also the code could be deduplicated in another method named `_create_temp_file` on your exiting Mixin, which might be nice too.

fao89: it looks good, I will work on it

pulp_ansible/app/galaxy/mixins.py
```diff
@@ -0,0 +1,32 @@
...
+    Provides a method that dispatches a task.
+    """
+
+    def _create_temp_file(self, validated_file):
+        """
+        Create the temporary file.
+        """
+        with NamedTemporaryFile(dir=settings.WORKING_DIRECTORY, delete=False) as temp_import_file:
+            shutil.copyfileobj(validated_file, temp_import_file)
+            return temp_import_file
```
bmbouter, line 21: Maybe you need to return the name because the file obj will already be closed?

fao89:
Closing in favor of: https://github.com/pulp/pulp_ansible/pull/306

PR #307, opened 2020-05-12
Fix flake8 issues

fao89:
[noissue]

### DIFF
--- a/.travis/cherrypick.py
+++ b/.travis/cherrypick.py
@@ -66,7 +66,7 @@ repo.git.checkout(STABLE_BRANCH)
 
 g = Github(GITHUB_TOKEN)
 grepo = g.get_repo(REPOSITORY)
-(label,) = (l for l in grepo.get_labels() if l.name == PR_LABEL)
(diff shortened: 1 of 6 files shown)


bmbouter:
@fao89 does this come from the plugin_template? If so would you be willing to open up a PR against galaxy_ng also? We won't maintain their CI with PRs for ever, but until we can hand it over to them I think it's still on us. What do you think?

fao89:
> @fao89 does this come from the plugin_template? If so would you be willing to open up a PR against galaxy_ng also? We won't maintain their CI with PRs for ever, but until we can hand it over to them I think it's still on us. What do you think?

it did not come from plugin_template, because plugin_template would bring the single container, and flake8 is disabled on galaxy_ng

bmbouter:
> > @fao89 does this come from the plugin_template? If so would you be willing to open up a PR against galaxy_ng also? We won't maintain their CI with PRs for ever, but until we can hand it over to them I think it's still on us. What do you think?
> 
> it did not come from plugin_template, because plugin_template would bring the single container, and flake8 is disabled on galaxy_ng

Thank you. As long as their CI isn't also broken I'm satisfied.

PR #309, opened 2020-05-14
Increased max length for CollectionVersion fields

fao89:
https://pulp.plan.io/issues/6648
closes #6648


CHANGES/6648.removal
```diff
@@ -0,0 +1 @@
+Increased max length for `documentation`, `homepage`, `issues`, `repository` in `CollectionVersion`
```
bmbouter, line 1: Can this be a `feature` instead? My reasoning is that it is backwards compatible since it supplies a migration so it could be either `.bugfix` or `.feature`. I'm proposing `.feature` because we avoid shipping migrations in Z-streams and technically any bugfux could go into a z-stream even though pulp_ansible doesn't have a Z stream yet.

PR #311, opened 2020-05-15
Updating v3 CollectionVersion download URL

fao89:
https://pulp.plan.io/issues/6510
closes #6510

### DIFF
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -181,7 +181,8 @@ class CollectionVersionSerializer(CollectionVersionListSerializer):
         """
         Get artifact download URL.
         """
-        host = settings.CONTENT_ORIGIN.strip("/")
-        prefix = settings.CONTENT_PATH_PREFIX.strip("/")
-        base_path = self.context["content_artifact"].relative_path.lstrip("/")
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -183,5 +183,7 @@ def get_download_url(self, obj):
         """
         host = settings.CONTENT_ORIGIN.strip("/")
         prefix = settings.CONTENT_PATH_PREFIX.strip("/")
-        base_path = self.context["content_artifact"].relative_path.lstrip("/")
-        return f"{host}/{prefix}/{base_path}"
+        distro_base_path = self.context["path"]
+        filename_path = self.context["content_artifact"].relative_path.lstrip("/")
+        download_url = f"{host}/{prefix}/{distro_base_path}/{filename_path}"
```
sivel, line 188: This still uses the original `CONTENT_ORIGIN`, which isn't required to have a scheme.  Should this be switched to a different config that does intentionally use a scheme?  `v2` uses `ANSIBLE_CONTENT_HOSTNAME` for building the `download_url`.

bmbouter: I agree thank you for pointing this out.

PR #313, opened 2020-05-18
Add AdminModels for pulp_ansible models

alikins:
Depends on https://github.com/pulp/pulpcore/pull/630

Adds django-admin setup for the pulp_ansible specific models


bmbouter:
Closing for the same reasons outlined [here](https://github.com/pulp/pulpcore/pull/838#issuecomment-709548516) (and I recommend the same).

PR #314, opened 2020-05-19
Reduce amount of time the repository lock is held

fao89:
https://pulp.plan.io/issues/6677
closes #6677

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -269,6 +269,15 @@ class CollectionImportViewSet(
         Returns a CollectionImport object.
         """
         instance = self.get_object()
+        task = instance.task
+
+        completed = ["completed", "failed", "canceled"]
+
+        for child in task.child_tasks.all():
+            if child.state not in completed:
+                task.state = child.state
+                task.finished_at = child.finished_at
+                break
 
         if "since" in self.request.query_params:
             since = parse_datetime(self.request.query_params["since"])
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -195,6 +196,17 @@ def create_collection_from_importer(importer_result):
     return collection_version
 
 
+def add_collection_to_repository(repository_pk, collection_version_pk):
```
alikins, line 199: This is presumably close enough to the same as https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/tasks/promotion.py#L4
that galaxy_ng should be able to use pulpcore.app.tasks.collections.add_collection_to_repository() once it is merged?

alikins: pinging @awcrosby

awcrosby: Yeah that is something we could look at replacing with this

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -269,6 +269,15 @@ def retrieve(self, request, *args, **kwargs):
...
         instance = self.get_object()
+        task = instance.task
+
+        completed = ["completed", "failed", "canceled"]
+
+        for child in task.child_tasks.all():
+            if child.state not in completed:
+                task.state = child.state
+                task.finished_at = child.finished_at
+                break
```
fao89, line 280: @bmbouter this is what I mentioned felt like a hack

fao89:
too old PR

PR #315, opened 2020-05-19
Speeding up CI with pre-built image

fao89:
[noissue]


.travis/post_before_script.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+echo "machine pulp
+login admin
+password password
+" > ~/.netrc
```
mdellweg, line 6: `ERROR! Unknown error when attempting to call Galaxy at 'http://pulp:80/pulp_ansible/galaxy/306fa13f-038e-4ca8-b891-dd6f94bee8f0/api': ~/.netrc access too permissive: access permissions must restrict access to only the owner (/home/travis/.netrc, line 3)`
Maybe set the umask before adding this.

fao89: got the solution from here:
https://github.com/coursera-dl/coursera-dl/issues/339#issuecomment-93878046

bmbouter:
Thank you @fao89 !

PR #319, opened 2020-05-28
Fix filename generation for ansible collection artifacts

cutwater:
https://pulp.plan.io/issues/6855
ansible/galaxy_ng#178

closes #6855


fao89:
LGTM!
Could you please add a changelog entry on CHANGES directory, example here:
https://github.com/pulp/pulp_ansible/blob/2a5a35334a35f9e66d918790d228c8e08547bb65/CHANGES/6510.bugfix

PR #324, opened 2020-06-08
Parametrize page size when syncing

fao89:
https://pulp.plan.io/issues/5494
closes #5494


pulp_ansible/app/constants.py
```diff
@@ -0,0 +1,2 @@
+# default results per page. used to calculate number of pages
+PAGE_SIZE = 10
```
daviddavis, line 2: What if we use a bigger number like 100 to reduce the number of page requests?

daviddavis:
Awesome job on this. The code LGTM. @bmbouter did you want to review as well?

pulp_ansible/app/constants.py
fao89: sounds good to me, I'll update it

CHANGES/5494.feature
```diff
@@ -0,0 +1 @@
+Parametrize page size when syncing roles/collections
```
bmbouter, line 1: Can this indicate the default of 100 also?

pulp_ansible/app/serializers.py
```diff
@@ -85,6 +87,16 @@ class Meta:
         model = AnsibleRepository
 
 
+class AnsibleRepositorySyncURLSerializer(RepositorySyncURLSerializer):
+    """
+    Serializer for Ansible RepositorySyncURL.
+    """
+
+    page_size = serializers.IntegerField(
+        help_text=_("The total of results per page."), default=PAGE_SIZE,
```
bmbouter, line 96: Does the default show in the openAPI schema for users?

fao89: not sure

bmbouter:
Which parts of this PR fix the bug when syncing from pulp to pulp? I see a lot of changes that make it configurable and that's great. I'm trying to understand which part is the bugfix.

bmbouter:
Also since it changes the openapi schema can you regenerate that with the docs so the latest openAPI schema will be committed?

pulp_ansible/app/tasks/utils.py
```diff
@@ -7,11 +7,12 @@
 from yaml.error import YAMLError
 
 
-def get_page_url(url, page=1):
+def get_page_url(url, page_size, page=1):
     """Get URL page."""
     parsed_url = urlparse(url)
     new_query = parse_qs(parsed_url.query)
     new_query["page"] = page
+    new_query["page_size"] = page_size
```
fao89, line 15: @bmbouter bugfix is here, adding `?page_size=PAGE_SIZE` on galaxy requests
https://pulp.plan.io/issues/5494#note-1

bmbouter: So when syncing from another Pulp serving the galaxy API, what would happen if page_size was unset? I guess I don't fully understand the bug claim itself.

pulp_ansible/app/serializers.py
fao89: it does
image

pulp_ansible/app/tasks/utils.py
fao89: actually I mixed things, the bug is fixed just by making `PAGE_SIZE`constant equals to 100. As it is the [default value for pulp](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/settings.py#L129). But it would break when syncing from galaxy, as its default is 10. So we are appending `?page_size=100`.
Making both pulp and galaxy default to 100. And if the user wants, giving them power to parameterize it.

PR #326, opened 2020-06-10
Enable token auth sync

fao89:
https://pulp.plan.io/issues/6540
closes #6540


pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        if not self.remote.token:
+            return await super()._run(extra_data=extra_data)
+
+        token = await self.update_token()
+        headers = {"Authorization": "Bearer {token}".format(token=token)}
+        # aiohttps does not allow to send auth argument and auth header together
+        self.session._default_auth = None
+
+        async with self.session.get(self.url, headers=headers, proxy=self.proxy) as response:
```
fao89, line 50: for stage env, I had to add `ssl=False`

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        async with self.token_lock:
+            log.info("Updating bearer token")
+            form_payload = {
+                "grant_type": "refresh_token",
+                "client_id": "cloud-services",
+                "refresh_token": self.remote.token,
+            }
+            url = self.remote.auth_url
+            async with self.session.post(url, data=form_payload, raise_for_status=True) as response:
```
fao89, line 71: for stage env, I had to add `ssl=False`

fao89:
I had an issue here:
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/tasks/collections.py#L340-L347

this endpoint https://ci.cloud.redhat.com/api/automation-hub/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/ returns:
```
{
    "href": "/pulp_ansible/galaxy/automation-hub/api/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/",
    "created_at": "2020-02-21T23:01:52.727205Z",
    "updated_at": "2020-02-21T23:01:52.727232Z",
    "namespace": "autohubtest2",
    "name": "collection_dep_a_hlrknkqw",
    "deprecated": false,
    "versions_url": "/pulp_ansible/galaxy/automation-hub/api/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/versions/",
    "highest_version": {
        "href": "/pulp_ansible/galaxy/automation-hub/api/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/versions/1.0.0/",
        "version": "1.0.0"
    }
}
```

so when getting `versions_url` it yields:
https://ci.cloud.redhat.com/pulp_ansible/galaxy/automation-hub/api/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/versions/
but the URL should be:
 https://ci.cloud.redhat.com/api/automation-hub/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/versions/

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+log = getLogger(__name__)
+
+
+class TokenAuthHttpDownloader(HttpDownloader):
+    """
+    Custom Downloader that automatically handles Token Based and Basic Authentication.
+
+    Additionally, use custom headers from DeclarativeArtifact.extra_data['headers']
```
bmbouter, line 18: Do these come from DeclarativeArtifact? I don't see them set anywhere.

bmbouter:
Can some user facing docs be added? We already have a [Collection section on syncing from galaxy.ansible.com](https://pulp-ansible.readthedocs.io/en/latest/workflows/collections.html#create-collection-remote), so I think this would be a "Syncing from cloud.redhat.com" as as an example.

@fao89 wdyt about adding this?

fao89:
> Can some user facing docs be added? We already have a [Collection section on syncing from galaxy.ansible.com](https://pulp-ansible.readthedocs.io/en/latest/workflows/collections.html#create-collection-remote), so I think this would be a "Syncing from cloud.redhat.com" as as an example.
> 
> @fao89 wdyt about adding this?

I agree, I'll add it

pulp_ansible/app/downloaders.py
fao89: I copied from `pulp_container` and forgot to remove it

pulp_ansible/app/models.py
```diff
@@ -227,6 +229,60 @@ class CollectionRemote(Remote):
     TYPE = "collection"
 
     requirements_file = models.TextField(null=True, max_length=255)
+    auth_url = models.CharField(null=True, max_length=255)
```
alikins, line 232: I was thinking this may need to be split info two Remote subclasses. One for the 'v2' and galaxy.ansible.com API, and another for Automation-hub / cloud.redhat.com / galaxy_ng.

Partly because "community" galaxy.ansible.com and cloud.redhat.com need different auth tokens.

cloud.redhat.com needs:
-  the 'auth_url' (pointing at https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token SSO server)
- The value of "token" is a "refresh_token" in from SSO  (the token a user gets from 'GET API TOKEN' on the cloud.redhat.com automation-hub ui).
- The token received from SSO is used in 'Autorization: Bearer <token here>'

community / galaxy.ansible.com / maybe standalone galaxy_ng
- No 'auth_url' is required.   For ansible-galaxy client the presence of a auth_url is what determines if SSO style auth is used and what format the authorization token takes.
- The value of "token" is a drf auth token
- The token is used in an 'Authorization: Token <token here>'  header.

alikins: name wise, I was thinking more or less "CollectionRemote" for v2 / community galaxy.ansible.com remotes since it exists already.

And 'AutomationHubRemote' for Remotes for talking to automation-hub / cloud.redhat.com.

pulp_ansible/app/downloaders.py
alikins: Possible to add the full ca_chain for ci / stage will work as well.

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,74 @@
...
+
+        This method provides the same return object type and documented in
+        :meth:`~pulpcore.plugin.download.BaseDownloader._run`.
+
+        """
+        if not self.remote.token:
+            return await super()._run(extra_data=extra_data)
+
+        token = await self.update_token()
+        headers = {"Authorization": "Bearer {token}".format(token=token)}
```
alikins, line 46: I think this will fail if the CollectionRemote is pointing at galaxy.ansible.com  (and I believe, if pointed against a standalone galaxy_ng that is using drf style 'Authorization: Token <token here>' style auth headers)

fao89:
I wasn't able to make the CI work, because it requires VPN to access https://ci.cloud.redhat.com/:
image
it was not about the token on travis CI

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,77 @@
...
+
+        """
+        if not self.remote.token:
+            return await super()._run(extra_data=extra_data)
+
+        if not self.remote.auth_url:
+            headers = {"Authorization": self.remote.token}
+        else:
+            token = await self.update_token_from_auth_url()
+            headers = {"Authorization": "Bearer {token}".format(token=token)}
```
fao89, line 48: Without `auth_url` will be considered as `community / galaxy.ansible.com / maybe standalone galaxy_ng`
With `auth_url` will be considered as `cloud.redhat.com`

daviddavis: I think this is simpler than splitting the remotes. The only reservation I have is if users will be confused about this. I guess we could make this clear in the serializer help text.

pulp_ansible/app/models.py
bmbouter: The differences in the technologies of the tokens (to me) motivates splitting up the Remote into two.

pulp_ansible/app/downloaders.py
bmbouter: I'm just seeing this comment now, maybe we should leave it as one remote.

fao89: pulp_container does something similar:
https://github.com/pulp/pulp_container/blob/master/pulp_container/app/downloaders.py#L25

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -77,3 +83,96 @@ def test_install_collection(self):
             subprocess.run(cmd.split())
 
             self.assertTrue(path.exists(directory), "Could not find directory {}".format(directory))
+
+    @skip_if(bool, "AH_token", False)
+    def test_install_collection_with_token_from_automation_hub(self):
```
fao89, line 88: Tested locally and it is working! It will only run on pulp org, as travis does not set variables on PRs

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -77,3 +83,96 @@ def test_install_collection(self):
...
+            self.assertTrue(
+                not path.exists(directory), "Directory {} already exists".format(directory)
+            )
+
+            subprocess.run(cmd.split())
+
+            self.assertTrue(path.exists(directory), "Could not find directory {}".format(directory))
+
+    @skip_if(bool, "GH_token", False)
+    def test_install_collection_with_token_from_galaxy(self):
```
fao89, line 136: It is more like a placeholder, as I don't know which collection requires token auth to be synced, I put `pulp.pulp_installer` here.
@bmbouter @ironfroggy @chouseknecht

CHANGES/6540.feature
```diff
@@ -0,0 +1 @@
+Enable token auth sync
```
bmbouter, line 1: Can this get more detail. As a user I would not understand this. Here are a few concrete suggestions:
* Call out the new fields
* Indicate which remote servers these fields can work with, e.g. `galaxy.ansible.com` and `cloud.redhat.com`.
* Relate the new fields to the ansible-galaxy CLI documentation here https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#configuring-the-ansible-galaxy-client

docs/workflows/collections.rst
```diff
@@ -136,6 +136,30 @@ Remote GET Response::
         "url": "https://galaxy-dev.ansible.com/api/v2/collections/testing/ansible_testing_content/",
     }
 
+For content sources token authentication. You can provide the ``token`` and/or ``auth_url``.
```
bmbouter, line 139:
```suggestion
For remote sources that require authentication, tokens can be used. You can provide the ``token`` and/or ``auth_url``.
```

^ still needs line wrapping

Also finding a way to https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#configuring-the-ansible-galaxy-client here would be good too.

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,83 @@
...
+from pulpcore.plugin.download import http_giveup, HttpDownloader
+
+
+log = getLogger(__name__)
+
+
+class TokenAuthHttpDownloader(HttpDownloader):
+    """
+    Custom Downloader that automatically handles Token Based and Basic Authentication.
+
```
bmbouter, line 17: extra line not needed

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+
+    token_lock = asyncio.Lock()
```
bmbouter, line 20: class attributes tend to be in all caps so `/token_lock/TOKEN_LOCK/`

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initialize the downloader.
+        """
+        self.remote = kwargs.pop("remote")
```
bmbouter, line 26: This is definitly how the current ones implement this, but at a pulpcore we were hoping to have it be done differently in the future. This pattern would be replaced with a subclassed DownloadFactory and the kwargs would become positional args passed here instead. I wrote more about this change here: https://pulp.plan.io/issues/6965#note-2

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,83 @@
...
+        if not self.remote.auth_url:
+            # Galaxy Token
+            headers = {"Authorization": self.remote.token}
+        else:
+            token = await self.update_token_from_auth_url()
+            # Keycloak Token
+            headers = {"Authorization": "Bearer {token}".format(token=token)}
+
+        # aiohttps does not allow to send auth argument and auth header together
+        self.session._default_auth = None
```
bmbouter, line 57: This could be better handled in the subclassed DownloadFactory also.

pulp_ansible/app/downloaders.py
```diff
@@ -0,0 +1,83 @@
...
+
+        async with self.session.get(self.url, headers=headers, proxy=self.proxy) as response:
+            response.raise_for_status()
+            to_return = await self._handle_response(response)
+            await response.release()
+
+        if self._close_session_on_finalize:
+            self.session.close()
+        return to_return
```
bmbouter, line 66: You could call super() to receive all of this instead.

pulp_ansible/app/models.py
```diff
@@ -227,6 +229,60 @@ class CollectionRemote(Remote):
...
+            self._download_factory = DownloaderFactory(
+                self,
+                downloader_overrides={
+                    "http": downloaders.TokenAuthHttpDownloader,
+                    "https": downloaders.TokenAuthHttpDownloader,
+                },
+            )
+            return self._download_factory
+
+    def get_downloader(self, remote_artifact=None, url=None, **kwargs):
```
bmbouter, line 262: Using the suggestion to subclass DownloadFactory would eliminate this method also.

pulp_ansible/app/models.py
```diff
@@ -227,6 +229,60 @@ class CollectionRemote(Remote):
...
+
+        Returns:
+            DownloadFactory: The instantiated DownloaderFactory to be used by
+                get_downloader()
+
+        """
+        try:
+            return self._download_factory
+        except AttributeError:
+            self._download_factory = DownloaderFactory(
```
bmbouter, line 253: This is where you would use your subclassed DownloaderFactory instead

pulp_ansible/app/serializers.py
```diff
@@ -95,6 +95,21 @@ class CollectionRemoteSerializer(RemoteSerializer):
         required=False,
         allow_null=True,
     )
+    auth_url = serializers.CharField(
+        help_text=_(
+            "The URL of a Keycloak server ‘token_endpoint’ "
```
bmbouter, line 100:
```suggestion
            "The URL to receive a session token from, e.g. used with Automation Hub."
```

Can we also add a sentence like: `see https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#configuring-the-ansible-galaxy-client for more details.`

pulp_ansible/app/serializers.py
```diff
@@ -95,6 +95,21 @@ class CollectionRemoteSerializer(RemoteSerializer):
...
+            "if using SSO auth (Automation Hub for ex)."
+        ),
+        allow_null=True,
+        required=False,
+        max_length=255,
+    )
+    token = serializers.CharField(
+        help_text=_("The token key to use for authentication."),
```
bmbouter, line 108: Can we also add a sentence like: `See https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#configuring-the-ansible-galaxy-client for more details`.

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -77,3 +83,96 @@ def test_install_collection(self):
...
+        self.addCleanup(self.remote_collection_api.delete, remote.pulp_href)
+
+        # Sync the repository.
+        self.assertEqual(repo.latest_version_href, f"{repo.pulp_href}versions/0/")
+        repository_sync_data = RepositorySyncURL(remote=remote.pulp_href)
+        sync_response = self.repo_api.sync(repo.pulp_href, repository_sync_data)
+        monitor_task(sync_response.task)
+        repo = self.repo_api.read(repo.pulp_href)
+
+        # Create a distribution.
```
bmbouter, line 156: You could probably just check the unit counts at this point instead, but I'm ok w/ the rest of this test also even though it's duplicated in various places.

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -77,3 +83,96 @@ def test_install_collection(self):
...
+        body = gen_distribution()
```
bmbouter, line 110: You could probably just check the unit counts at this point instead, but I'm ok w/ the rest of this test also even though it's duplicated in various places.

pulp_ansible/app/downloaders.py
fao89: I couldn't, pulpcore is different:
```python
async with self.session.get(self.url, proxy=self.proxy, auth=self.auth) as response:
```
https://github.com/pulp/pulpcore/blob/master/pulpcore/download/http.py#L197

PR #327, opened 2020-06-19
Release 0.2.0b14

fao89:
Redmine query of issues to close:
https://pulp.plan.io/issues?set_filter=1&status_id=*&issue_id=6889,5494,6772
Release commit == 04bb171

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 galaxy_importer
 packaging
-pulpcore>=3.0,<3.5
+pulpcore>=3.0
 PyYAML
 semantic_version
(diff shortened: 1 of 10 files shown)


bmbouter:
On the release commit I expected there to be a pulpcore version restriction to < 3.5

requirements.txt
```diff
@@ -1,5 +1,2 @@
 galaxy_importer
 packaging
-pulpcore>=3.0
```
bmbouter, line 3: We still need to depend on pulpcore.

What about these other deps why are they being removed?

fao89: old version of release script

PR #328, opened 2020-06-23
Update installation doc for for the new single role name

mikedep333:
[noissue]

re: #7005
As a user, I can run a single stably-named Ansible role to install Pulp
https://pulp.plan.io/issues/7005

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -4,13 +4,14 @@ Installation
 Install using Ansible
 ---------------------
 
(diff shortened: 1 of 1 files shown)


fao89:
we should remove
```
git clone https://github.com/pulp/pulp_installer.git
```
and use the collection:
```
ansible-galaxy collection install pulp.pulp_installer
```
as we did here:
https://docs.pulpproject.org/en/master/nightly/installation/instructions.html#ansible-installation-recommended

PR #331, opened 2020-06-30
Replace whitelist with allowlist

fao89:
[noissue]


daviddavis:
Thanks for this. The variable names confuse me because they aren't even lists AFAICT. I wonder if it should be ANSIBLE_DEMO_COLLECTION and ANSIBLE_ELASTIC_ROLE?

fao89:
> Thanks for this. The variable names confuse me because they aren't even lists AFAICT. I wonder if it should be ANSIBLE_DEMO_COLLECTION and ANSIBLE_ELASTIC_ROLE?

I did a simple replace, but you are right, most of them are just `namespace.name`

PR #335, opened 2020-07-17
Stop using artifacts when importing collections

fao89:
Required PR: https://github.com/pulp/pulpcore/pull/793

https://pulp.plan.io/issues/6718
closes #6718

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -1,5 +1,4 @@
 import asyncio
-import contextlib
 from gettext import gettext as _
 import json
 import logging
@@ -19,6 +18,7 @@ from pulpcore.plugin.models import (
     ContentArtifact,
     CreatedResource,
     ProgressReport,
+    PulpTemporaryFile,
     Remote,
     Repository,
 )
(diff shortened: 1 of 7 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -118,41 +118,41 @@ def import_collection(
...
+                artifact_file, filename=filename, logger=user_facing_logger
+            )
+            artifact = temp_file.to_artifact()
+            importer_result["artifact_url"] = reverse("artifacts-detail", args=[artifact.pk])
+            collection_version = create_collection_from_importer(importer_result)
 
-                importer_result["artifact_url"] = reverse("artifacts-detail", args=[artifact_pk])
-                collection_version = create_collection_from_importer(importer_result)
+    except ImporterError as exc:
+        log.info(f"Collection processing was not successfull: {exc}")
```
daviddavis, line 136: I wonder if we should delete temp_file if there was an import error?

fao89: oh yeah, I forgot it, thanks!

CHANGES/6718.removal
```diff
@@ -0,0 +1 @@
+Stop using artifacts when importing collections
```
bmbouter, line 1: I think this could be `6718.misc` instead because it's not a change a user could perceive as different.

PR #337, opened 2020-07-22
Handle link between Repositories and Remotes

daviddavis:
fixes #7194


bmbouter:
I am so glad to see this story. A test would be really great.

daviddavis:
@fao89 @bmbouter I added a test in [the pulpcore PR](https://github.com/pulp/pulpcore/pull/788) to test this out. Let me know if you think pulp_ansible should also have a test.

bmbouter:
Thank you!

PR #338, opened 2020-08-04
Hide Galaxy api from pulp api docs

mdellweg:
[noissue]


bmbouter:
I think this is good overall because right now these APIs are a mess in the API schema. I think this would clean up the APIs and if we do want to bring them back we can do so while making sure they look correct at that time.

PR #339, opened 2020-08-05
Revert "Hide Galaxy api from pulp api docs"

fao89:
Reverts pulp/pulp_ansible#338


bmbouter:
This was reverted because it removed the bindings which are used in the CI of other projects, e.g. ansible-lint.

PR #342, opened 2020-08-10
Add modify_ansible_repo_content permission.

newswangerd:
Adds `modify_ansible_repo_content` to the `AnsibleRepository` model.


pulp_ansible/app/models.py
```diff
@@ -256,6 +256,8 @@ class AnsibleRepository(Repository):
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
 
+        permissions = (("modify_ansible_repo_content", "Modify Ansible Repository Content"),)
```
bmbouter, line 259: I didn't realize it when I susggested this name, but there is a naming conventions Permissions use. As is this reads in the list like this:

```
 'Can add ansible repository',
 'Can change ansible repository',
 'Can delete ansible repository',
 'Modify Ansible Repository Content',
 'Can view ansible repository',
 'Can add collection',
 'Can change collection',
 'Can delete collection',
```

So I think the name should be `"Can modify ansible repository content"`.

bmbouter: For the "codename" (the slug version) that list currently reads like:

```
 'change_ansiblerepository',
 'delete_ansiblerepository',
 'modify_ansible_repo_content',
 'view_ansiblerepository',
 'add_collection',
 'change_collection',
 'delete_collection',
```

I think what we have right now is good.

pulp_ansible/app/migrations/0020_auto_20200810_1926.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ansible', '0019_collection_token'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='ansiblerepository',
+            options={'default_related_name': '%(app_label)s_%(model_name)s', 'permissions': (('modify_ansible_repo_content', 'Modify Ansible Repository Content'),)},
```
bmbouter, line 15: This also needs the new string or the db will be incorrect.

PR #348, opened 2020-08-27
Pagination query params according to API versions

fao89:
https://pulp.plan.io/issues/7396
closes #7396

AH
```
GET /api/automation-hub/v3/collections/
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "meta": {
        "count": 61
    },
    "links": {
        "first": "/api/automation-hub/v3/collections/?limit=10&offset=0",
        "previous": null,
        "next": "/api/automation-hub/v3/collections/?limit=10&offset=10",
        "last": "/api/automation-hub/v3/collections/?limit=10&offset=51"
    },
    "data": [
```

Galaxy
```
GET /api/v2/collections/?page=1&page_size=1
HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 595,
    "next": "/api/v2/collections/?page=2&page_size=1",
    "next_link": "/api/v2/collections/?page=2&page_size=1",
    "previous": null,
    "previous_link": null,
    "results": [
```

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -47,6 +47,7 @@ from pulp_ansible.app.models import (
 )
 from pulp_ansible.app.serializers import CollectionVersionSerializer
 from pulp_ansible.app.tasks.utils import (
+    get_api_version,
     get_page_url,
     parse_metadata,
     parse_collections_requirements_file,
@@ -318,7 +319,7 @@ class CollectionSyncFirstStage(Stage):
         remote = self.remote
         collection_info = self.collection_info
 
-        def _get_url(page):
+        def _get_url(page, api_version):
             if collection_info:
                 name, version, source = collection_info[page - 1]
                 namespace, name = name.split(".")
@@ -326,7 +327,7 @@ class CollectionSyncFirstStage(Stage):
                 url = f"{root}/api/v2/collections/{namespace}/{name}"
                 return url
 
-            return get_page_url(remote.url, page)
+            return get_page_url(remote.url, api_version, page)
 
         def _build_url(path_or_url):
             """Check value and turn it into a url using remote.url if it's a relative path."""
@@ -339,11 +340,13 @@ class CollectionSyncFirstStage(Stage):
 
         progress_data = dict(message="Parsing Galaxy Collections API", code="parsing.collections")
         with ProgressReport(**progress_data) as progress_bar:
-            url = _get_url(page_count)
+            api_version = get_api_version(remote.url)
+            url = _get_url(page_count, api_version)
             downloader = remote.get_downloader(url=url)
             initial_data = parse_metadata(await downloader.run())
 
-            count = len(self.collection_info) or initial_data.get("count", 1)
+            _count = initial_data.get("count") or initial_data.get("meta", {}).get("count", 1)
+            count = len(self.collection_info) or _count
             page_count = math.ceil(float(count) / float(PAGE_SIZE))
             progress_bar.total = count
             progress_bar.save()
@@ -351,7 +354,7 @@ class CollectionSyncFirstStage(Stage):
             # Concurrent downloads are limited by aiohttp...
             not_done = set()
             for page in range(1, page_count + 1):
-                downloader = remote.get_downloader(url=_get_url(page))
+                downloader = remote.get_downloader(url=_get_url(page, api_version))
                 not_done.add(downloader.run())
 
             while not_done:
@@ -360,8 +363,10 @@ class CollectionSyncFirstStage(Stage):
                 for item in done:
                     data = parse_metadata(item.result())
 
-                    # v2 uses 'results' as the key while v3 uses 'data'
-                    results = data.get("results") or data.get("data") or [data]
+                    if api_version < 3:
+                        results = data.get("results", [data])
+                    else:
+                        results = data.get("data", [data])
 
                     for result in results:
                         download_url = result.get("download_url")
--- a/pulp_ansible/app/tasks/utils.py
+++ b/pulp_ansible/app/tasks/utils.py
@@ -1,5 +1,6 @@
 from gettext import gettext as _
 import json
+import re
 import yaml
 
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
@@ -9,12 +10,26 @@ from yaml.error import YAMLError
 from pulp_ansible.app.constants import PAGE_SIZE
 
 
-def get_page_url(url, page=1):
+def get_api_version(url):
+    """Get API version."""
+    result = re.findall(r"/api/v(\d)", url)
+    if len(result) != 1:
+        raise RuntimeError("Could not determine Galaxy API version")
+    return int(result[0])
+
+
+def get_page_url(url, api_version, page=1):
     """Get URL page."""
     parsed_url = urlparse(url)
     new_query = parse_qs(parsed_url.query)
-    new_query["page"] = page
-    new_query["page_size"] = PAGE_SIZE
+
+    if api_version < 3:
+        new_query["page"] = page
+        new_query["page_size"] = PAGE_SIZE
+    else:
+        new_query["offset"] = (page - 1) * PAGE_SIZE
+        new_query["limit"] = PAGE_SIZE
+
     return urlunparse(parsed_url._replace(query=urlencode(new_query, doseq=True)))
 
 
--- a/CHANGES/7396.bugfix
+++ b/CHANGES/7396.bugfix
@@ -1,0 +1,3 @@
+Pagination query params according to API versions.
+v1 and v2 - `page` and `page_size`
+v3 or above - `offset` and `limit`
--- a/CHANGES/7435.removal
+++ b/CHANGES/7435.removal
@@ -1,0 +1,1 @@
+Changed V3 pagination to match Galaxy V3 API pagination
--- a/pulp_ansible/app/galaxy/v3/pagination.py
+++ b/pulp_ansible/app/galaxy/v3/pagination.py
@@ -1,0 +1,84 @@
+from rest_framework import pagination
+from rest_framework.response import Response
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+
+
+class LimitOffsetPagination(pagination.LimitOffsetPagination):
+    """
+    Pagination for V3.
+
+    """
+
+    default_limit = 10
+    max_limit = 100
+
+    def get_first_link(self):
+        """First link."""
+        url = self.request.get_full_path()
+        url = replace_query_param(url, self.limit_query_param, self.limit)
+
+        return replace_query_param(url, self.offset_query_param, 0)
+
+    def get_last_link(self):
+        """Last link."""
+        url = self.request.get_full_path()
+        url = replace_query_param(url, self.limit_query_param, self.limit)
+
+        offset = self.count - self.limit if (self.count - self.limit) >= 0 else 0
+        return replace_query_param(url, self.offset_query_param, offset)
+
+    def get_next_link(self):
+        """Next link."""
+        if self.offset + self.limit >= self.count:
+            return None
+
+        url = self.request.get_full_path()
+        url = replace_query_param(url, self.limit_query_param, self.limit)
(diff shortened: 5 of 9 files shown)


pulp_ansible/app/tasks/utils.py
```diff
@@ -13,8 +13,13 @@ def get_page_url(url, page=1):
     """Get URL page."""
     parsed_url = urlparse(url)
     new_query = parse_qs(parsed_url.query)
-    new_query["page"] = page
-    new_query["page_size"] = PAGE_SIZE
+
+    if "v1" in url or "v2" in url:
```
bmbouter, line 17: Can the sync code deduplicate this check with the other one @daviddavis linked to? That way all the checking is done on a variable that checks the url once.

bmbouter, line 17: Also can the check include `/api/v2` and `/api/v3` instead? this will avoid complications when collection names have v2 in them for example

fao89: it would not work with `/api/automation-hub/v3/collections/`

fao89: I don't know what is the check you are talking about

daviddavis: Here you go:

https://github.com/pulp/pulp_ansible/blob/8e4168dca96ab00d329ef8f5e0e9886218ef1302/pulp_ansible/app/tasks/collections.py#L363-L364

pulp_ansible/app/tasks/utils.py
```diff
@@ -9,12 +9,18 @@
 from pulp_ansible.app.constants import PAGE_SIZE
 
 
-def get_page_url(url, page=1):
+def get_page_url(url, page=1, api_version=None):
```
bmbouter, line 12: `api_version` I think should be a required param.

fao89: `get_page_url` is used by roles task, I made it optional because I wasn't sure if I have to change roles task

fao89: https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/tasks/roles.py#L133-L145

pulp_ansible/app/tasks/collections.py
```diff
@@ -343,15 +343,22 @@ def _build_url(path_or_url):
             downloader = remote.get_downloader(url=url)
             initial_data = parse_metadata(await downloader.run())
 
-            count = len(self.collection_info) or initial_data.get("count", 1)
+            api_version = None
+            if "results" in initial_data.keys():
```
bmbouter, line 347: I think inspecting the remote's url would give us a more reliable case when future versions use one of these two response styles and yet we need to detect v4 for example.

Also can we make api_version and integer so simplicity.

That would make this block:

```
api_version = None
if "/api/v2/" in remote.url:
    api_version = 2
if "/api/v3/" in remote.url:
    api_version

Another option would be to regexparse the url itself and match on the integer.

```
import regex
galaxy_version_regex_compiled = re.compile('/api/v(\d)')
match_result = galaxy_version_regex_compiled.match(remote.url)
if match_result:
    api_version = int(match_result.groups()[0])
else:
    raise RuntimeError('Could not determine Galaxy API version')
```

Comment (fao89) [reply pulp_ansible/app/tasks/collections.py:347]: I thought the deduplication idea was to avoid  these string comparisons such as `"/api/v2/"`

Comment (daviddavis) [reply pulp_ansible/app/tasks/collections.py:347]: I believe the idea was to reuse the api_version after parsing it from the url. So this:

```python
# v2 uses 'results' as the key while v3 uses 'data'
results = data.get("results") or data.get("data") or [data]
```

Would become:

```python
if api_version == 2:
        results = data.get("results", [data])
else:
    results = data.get("data", [data])
```

fao89: I think the current code is more pythonic

bmbouter: My believe is that python prefers explicitness over implicitness. For example the need to add an additional comment suggests the code itself isn't very readable on its own basis.

pulp_ansible/app/tasks/utils.py
```diff
@@ -9,12 +10,26 @@
 from pulp_ansible.app.constants import PAGE_SIZE
 
 
-def get_page_url(url, page=1):
+def get_api_version(url):
+    """Get API version."""
+    result = re.findall(r"/api/v(\d)", url)
+    if len(result) != 1:
+        raise RuntimeError("Could not determine Galaxy API version")
```
bmbouter, line 17: Niiice

pulp_ansible/app/tasks/collections.py
fao89: the URL approach fails for `PulpToPulpSyncCase`:
```
(pulp) (pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ nc 127.0.0.1 4444
> /home/vagrant/devel/pulp_ansible/pulp_ansible/app/tasks/collections.py(345)_fetch_collections()
-> url = _get_url(page_count, api_version)
(Pdb) api_version
2
(Pdb) remote.url
'https://galaxy.ansible.com/api/v2/collections/testing/k8s_demo_collection'
(Pdb) c




^C
(pulp) (pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ nc 127.0.0.1 4444
Ncat: Connection refused.
(pulp) (pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ nc 127.0.0.1 4444
> /home/vagrant/devel/pulp_ansible/pulp_ansible/app/tasks/collections.py(345)_fetch_collections()
-> url = _get_url(page_count, api_version)
(Pdb) remote.url
'http://pulp3-source-fedora31.lan/pulp_ansible/galaxy/55ec1549-1a44-49ec-b10f-b96cd27c31da/api/v3/collections/testing/k8s_demo_collection'
(Pdb) api_version
3
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -360,8 +369,10 @@ def _build_url(path_or_url):
                 for item in done:
                     data = parse_metadata(item.result())
 
-                    # v2 uses 'results' as the key while v3 uses 'data'
-                    results = data.get("results") or data.get("data") or [data]
+                    if api_version < 3 or "results" in data.keys():
```
fao89, line 372: Need to check `results` because:
```
(pulp) (pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ nc 127.0.0.1 4444
> /home/vagrant/devel/pulp_ansible/pulp_ansible/app/tasks/collections.py(345)_fetch_collections()
-> url = _get_url(page_count, api_version)
(Pdb) api_version
2
(Pdb) remote.url
'https://galaxy.ansible.com/api/v2/collections/testing/k8s_demo_collection'
(Pdb) c




^C
(pulp) (pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ nc 127.0.0.1 4444
Ncat: Connection refused.
(pulp) (pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ nc 127.0.0.1 4444
> /home/vagrant/devel/pulp_ansible/pulp_ansible/app/tasks/collections.py(345)_fetch_collections()
-> url = _get_url(page_count, api_version)
(Pdb) remote.url
'http://pulp3-source-fedora31.lan/pulp_ansible/galaxy/55ec1549-1a44-49ec-b10f-b96cd27c31da/api/v3/collections/testing/k8s_demo_collection'
(Pdb) api_version
3
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -339,19 +340,27 @@ def _build_url(path_or_url):
...
+            url = _get_url(page_count, api_version)
             downloader = remote.get_downloader(url=url)
             initial_data = parse_metadata(await downloader.run())
 
-            count = len(self.collection_info) or initial_data.get("count", 1)
+            if api_version == 3 and "results" in initial_data.keys():
+                api_version = 2
+                url = _get_url(page_count, api_version)
+                downloader = remote.get_downloader(url=url)
+                initial_data = parse_metadata(await downloader.run())
```
fao89, line 352: For assuring the page size is what we expect

pulp_ansible/app/tasks/collections.py
bmbouter: I'm confused about why. The `/api/v2/` and `/api/v3/` is shown in both urls.

bmbouter: My takeaway is that the `_get_url()` implementation is not parsing correctly in these cases.

fao89: `_get_url()` is ok, looking the `TestCase`, we can see it syncs from v2: https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/v3/test_sync.py#L39
but it exposes as v3 https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/v3/test_sync.py#L66

The problem is, we don't expose v3 the same way galaxy does, we always expose as `results`, never as `data`
```
(pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ nc 127.0.0.1 4444
> /home/vagrant/devel/pulp_ansible/pulp_ansible/app/tasks/collections.py(374)_fetch_collections()
-> if api_version < 3: 
(Pdb) item.result()
DownloadResult(url='http://pulp3-source-fedora31.lan/pulp_ansible/galaxy/31f91a4e-6fa0-4e44-9d74-467a39f64cc6/api/v3/collections/testing/k8s_demo_collection/versions/', artifact_attributes={
'size': 318, 'sha512': 'f0618dd61223fe837d4b2bddbec40612a129e06d10e3587fc128cd729febef274ca62caf9eb170da40daa12294290412efd787c66e24bffd93c3ff321dacfc75', 'sha384': 'e18d39765ff12f443e452f71
68a0b8fe8a0e00511cdb1f6178a2973263b96beff6b71943014ac376ecaf0b63040c2fb7', 'sha256': '999fddf130260adbe24176aedf19f75b71e56fa4e0b207d7aaa08fa829dc01ab', 'sha224': '46aaf28f9ba5092629e9f90a78
7b12ef5a0fb4e0721c02c047ccf647', 'sha1': '1fab81a787871bd8d3f095360028d91bd437b57f', 'md5': '4655e17a0333c7a38bac259162cca1c9'}, path='/var/lib/pulp/tmp/63412@pulp3-source-fedora31.lan/7e313
7e2-6df8-4d9a-b066-5386c86a7445/tmpc9twrbz0', headers=<CIMultiDictProxy('Date': 'Mon, 31 Aug 2020 19:10:05 GMT', 'Server': 'gunicorn/20.0.4', 'Content-Type': 'application/json', 'Vary': 'Acc
ept', 'Allow': 'GET, HEAD, OPTIONS', 'X-Frame-Options': 'SAMEORIGIN', 'Content-Length': '318')>)
(Pdb) data
{'count': 1, 'next': None, 'previous': None, 'results': [{'version': '0.0.3', 'certification': 'needs_review', 'href': '/pulp_ansible/galaxy/31f91a4e-6fa0-4e44-9d74-467a39f64cc6/api/v3/colle
ctions/testing/k8s_demo_collection/versions/0.0.3/', 'created_at': '2020-08-31T19:10:01.766252Z', 'updated_at': '2020-08-31T19:10:01.766271Z'}]}

```

fao89: we need https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/pagination.py#L46

bmbouter:
This PR looks great. It fixes a second issue too though, the Galaxy V3 API not having the same response format as the galaxy_ng project defined. As one final thing would you be willing to:

1) open a bug against pulp_ansible describing the difference?
2) add a `CHANGES/xxxx.bugfix` release note describing the bug as being fixed
3) add a `CHANGES/xxxx.removal` release note describing the change as backwards incompatible

PR #351, opened 2020-08-31
Add endpoint to show docs_blob for a CollectionVersion

awcrosby:
ref #7397

https://pulp.plan.io/issues/7397

### DIFF
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -186,3 +186,13 @@ class CollectionVersionSerializer(CollectionVersionListSerializer):
         filename_path = self.context["content_artifact"].relative_path.lstrip("/")
         download_url = f"{host}/{distro_base_path}/{filename_path}"
         return download_url
+
+
+class CollectionVersionDocsSerializer(serializers.ModelSerializer):
+    """A serializer to display the docs_blob of a CollectionVersion."""
+
+    docs_blob = serializers.JSONField()
+
+    class Meta:
+        fields = ("docs_blob",)
+        model = models.CollectionVersion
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -241,6 +242,14 @@ def set_certified(self, request, *args, **kwargs):
         return Response({})
 
 
+class CollectionVersionDocsViewSet(CollectionVersionViewSet):
```
bmbouter, line 245: Is this viewset expected to support list() operation?

awcrosby: no, just retrieve()... the urls.py only has the get()/retrieve()

bmbouter: Sorry I missed this reply until now. Then I think we need to have this be its own viewset because by inheriting from `CollectionVersionViewSet` it also receives other action endpoints like `list` for example.

If their are common parts, then those could be moved to a mixin and shared between both classes.

awcrosby: Updated per above.

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -186,3 +186,12 @@ def get_download_url(self, obj):
         filename_path = self.context["content_artifact"].relative_path.lstrip("/")
         download_url = f"{host}/{distro_base_path}/{filename_path}"
         return download_url
+
+
+class CollectionVersionDocsSerializer(CollectionVersionSerializer):
```
bmbouter, line 191: Can we split this from the `CollectionVersionSerializer` also? Maybe it would just be:

```
class CollectionVersionDocsSerializer(serializers.ModelSerializer):
    """A serializer to display the docs_blob of a CollectionVersion."""

    docs_blob = serializers.JSONField()

    class Meta:
        fields = ("docs_blob",)
        model = models.CollectionVersion
```

PR #352, opened 2020-09-01
Allow requirements file field to be longer

daviddavis:
fixes #7434


rochacbruno:
@daviddavis 👍🏻 thanks!

PR #353, opened 2020-09-02
Build collections URL according to requirements.yml

fao89:
https://pulp.plan.io/issues/7412
closes #7412

### DIFF
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -132,7 +132,20 @@ class CollectionRemoteSerializer(RemoteSerializer):
         data = super().validate(data)
 
         if data.get("requirements_file"):
-            parse_collections_requirements_file(data["requirements_file"])
+            parsed_file = parse_collections_requirements_file(data["requirements_file"])
+            urls_to_validate = [data["url"]]
+            for name, version, source in parsed_file:
+                if source:
+                    urls_to_validate.append(source)
+
+            for url in urls_to_validate:
+                if not url.endswith("/v2/collections") and not url.endswith("/v3/collections"):
+                    raise serializers.ValidationError(
+                        _(
+                            "Invalid URL {url}. ".format(url=data["url"])
+                            + "Ensure the URL ends in either '/v2/collections' or '/v3/collections'"
+                        )
+                    )
 
         if data["url"].endswith("/"):
             raise serializers.ValidationError(_("url should not end with '/'"))
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -324,7 +324,7 @@ class CollectionSyncFirstStage(Stage):
                 name, version, source = collection_info[page - 1]
                 namespace, name = name.split(".")
                 root = source or remote.url
-                url = f"{root}/api/v2/collections/{namespace}/{name}"
+                url = f"{root.rstrip('/')}/{namespace}/{name}"
                 return url
 
             return get_page_url(remote.url, api_version, page)
--- a/pulp_ansible/tests/functional/constants.py
+++ b/pulp_ansible/tests/functional/constants.py
@@ -82,7 +82,7 @@ ANSIBLE_COLLECTION_REQUIREMENT = """
 collections:
 - name: testing.ansible_testing_content
   version: ">=1.0.0,<=2.0.0"
-  source: https://galaxy-dev.ansible.com
+  source: https://galaxy-dev.ansible.com/api/v2/collections
 - testing.k8s_demo_collection
 """
 
--- a/CHANGES/7412.bugfix
+++ b/CHANGES/7412.bugfix
@@ -1,0 +1,1 @@
+Build collections URL according to requirements.yml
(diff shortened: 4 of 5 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -324,7 +324,7 @@ def _get_url(page, api_version):
                 name, version, source = collection_info[page - 1]
                 namespace, name = name.split(".")
                 root = source or remote.url
-                url = f"{root}/api/v2/collections/{namespace}/{name}"
+                url = urljoin(root, f"/api/v{api_version}/collections/{namespace}/{name}")
```
fao89, line 327: it fixes the issue only for galaxy V2, but it won't work for syncing from Galaxy V3, AH or Pulp

fao89: @bmbouter @daviddavis suggestions?

bmbouter: Because those sources use the V3 api?

fao89: I changed the code, it is enough to solve the situation described on the issue, but I don't like this solution

fao89: @newswangerd maybe it helps to unblock your tests

bmbouter:
@fao89 I looked at this problem and I think we should do something like this: https://pulp.plan.io/issues/7412#note-5 instead. wdyt?

pulp_ansible/app/serializers.py
```diff
@@ -132,7 +133,18 @@ def validate(self, data):
...
+            urls_to_validate = [data["url"]]
+            for name, version, source in parsed_file:
+                if source:
+                    urls_to_validate.append(source)
+
+            for url in urls_to_validate:
+                result = re.findall(r"/v(\d)/collections", url)
+                if not result:
+                    raise serializers.ValidationError(
+                        _("Invalid URL {url}".format(url=data["url"]))
```
fao89, line 146: I couldn't think a better error message

fao89:
> @fao89 I looked at this problem and I think we should do something like this: https://pulp.plan.io/issues/7412#note-5 instead. wdyt?

this is much better now, thank you!

pulp_ansible/app/serializers.py
```diff
@@ -132,7 +133,18 @@ def validate(self, data):
         data = super().validate(data)
 
         if data.get("requirements_file"):
-            parse_collections_requirements_file(data["requirements_file"])
+            parsed_file = parse_collections_requirements_file(data["requirements_file"])
```
bmbouter, line 136: I'm confused, I didn't think we would be validating the requirements file since it doesn't contain urls typically. https://docs.ansible.com/ansible/devel/galaxy/user_guide.html#install-multiple-collections-with-a-requirements-file

I thought when a requirements file is being used, the `remote.url` would be validated.

fao89: I'm validating `source` which is a URL

pulp_ansible/tests/functional/constants.py
```diff
@@ -82,7 +82,7 @@
 collections:
 - name: testing.ansible_testing_content
   version: ">=1.0.0,<=2.0.0"
-  source: https://galaxy-dev.ansible.com
+  source: https://galaxy-dev.ansible.com/api/v2/collections
```
fao89, line 85: @bmbouter see this URL would break

pulp_ansible/app/tasks/collections.py
```diff
@@ -324,7 +324,7 @@ def _get_url(page, api_version):
                 name, version, source = collection_info[page - 1]
                 namespace, name = name.split(".")
                 root = source or remote.url
-                url = f"{root}/api/v2/collections/{namespace}/{name}"
+                url = f"{root}/{namespace}/{name}"
```
fao89, line 327: it picks source or the remote.url, so I had to validate it

pulp_ansible/app/serializers.py
bmbouter: Maybe `Inavlid URL {url}. Ensure the URL ends in either '/v2/collections/' or '/v3/collections/'.`

pulp_ansible/app/serializers.py
```diff
@@ -132,7 +133,18 @@ def validate(self, data):
...
+            urls_to_validate = [data["url"]]
+            for name, version, source in parsed_file:
+                if source:
+                    urls_to_validate.append(source)
+
+            for url in urls_to_validate:
+                result = re.findall(r"/v(\d)/collections", url)
```
bmbouter, line 143: What do we do about the optional trailing slash or not? I think we either need to require it or not, but not allow both.

bmbouter: Also what about cases where it's something like /something/api/v2/collection/more/stuff/v2/collectionmine/. This would also validate. So I think validating at the end is key.

bmbouter: So maybe this check would instead be:

`if url.endswith('/v2/collections/`) or url.endswith('/v3/collections/`):`

or possibly `if url.endswith('/v2/collections`) or url.endswith('/v3/collections`):` depending on how the trailing slash is handled.

pulp_ansible/app/tasks/collections.py
bmbouter: I see how this correctly handles the case of forming the url for the requirements file, but in the case of no requirements file, where does it form those urls from?

fao89: if it does not have a requirements file, it does not fall on this if

bmbouter: Cool, so to confirm, we are able to form the right urls when not using a requirements file?

fao89: yes, we are, we have lots of tests doing that

bmbouter (approved):
This looks right to me, thanks @fao89 !

PR #354, opened 2020-09-02
Fix get_api_version  regex

fao89:
[noissue]

```
In [1]: import re

In [2]: re.findall(r"/api(?:/automation-hub)?/v(\d)", "https://galaxy.ansible.com/api/v2/collections")
Out[2]: ['2']

In [3]: re.findall(r"/api(?:/automation-hub)?/v(\d)", "https://cloud.redhat.com/api/automation-hub/v3/collections")
Out[3]: ['3']
```

### DIFF
--- a/pulp_ansible/app/tasks/utils.py
+++ b/pulp_ansible/app/tasks/utils.py
@@ -12,8 +12,8 @@ from pulp_ansible.app.constants import PAGE_SIZE
 
 def get_api_version(url):
     """Get API version."""
-    result = re.findall(r"/api/v(\d)", url)
-    if len(result) != 1:
+    result = re.findall(r"/v(\d)/", url)
+    if len(result) == 0:
         raise RuntimeError("Could not determine Galaxy API version")
     return int(result[0])
 


bmbouter:
@newswangerd I thought the galaxy API spec had `/api/v2/` or `/api/v3/`. Do you know why this url is this way? `https://cloud.redhat.com/api/automation-hub/v3/collections`

bmbouter:
@fao89 what I've learned from @newswangerd on chat is that the API beings with the `v#` itself. What this means for this PR is that the reges shoiuld look for the `v\(d+)\` itself.

fao89:
> @fao89 what I've learned from @newswangerd on chat is that the API beings with the `v#` itself. What this means for this PR is that the reges shoiuld look for the `v\(d+)\` itself.

how about when a name or namespace contains v(some number)?

bmbouter:
> > @fao89 what I've learned from @newswangerd on chat is that the API beings with the `v#` itself. What this means for this PR is that the reges shoiuld look for the `v\(d+)\` itself.
> 
> how about when a name or namespace contains v(some number)?

I expect the regex to match multiple times, and our version detection should always use the first one which we know will come first in the string.

pulp_ansible/app/tasks/utils.py
```diff
@@ -12,7 +12,7 @@
 
 def get_api_version(url):
     """Get API version."""
-    result = re.findall(r"/api/v(\d)", url)
+    result = re.findall(r"v(\d+)", url)
```
bmbouter, line 15: Actually I should not have recommended the `+`, sorry for this. Also it needs slashes on the front and back that @newswangerd suggested on irc, but we didn't bring to you here.

```suggestion
    result = re.findall(r"/v(\d)/", url)
```

pulp_ansible/app/tasks/utils.py
```diff
@@ -12,7 +12,7 @@
 
 def get_api_version(url):
     """Get API version."""
-    result = re.findall(r"/api/v(\d)", url)
+    result = re.findall(r"/v(\d)/", url)
     if len(result) != 1:
```
bmbouter, line 16: Actually I think this should be `if len(result) == 0` because in the case when there is a v2 in the collection name this will be 2+.

bmbouter (approved):
Thank you!

PR #355, opened 2020-09-04
Sync docs_blob information for collection versions

fao89:
https://pulp.plan.io/issues/7439
closes #7439

### DIFF
--- a/pulp_ansible/app/downloaders.py
+++ b/pulp_ansible/app/downloaders.py
@@ -6,10 +6,29 @@ import json
 from aiohttp import BasicAuth
 from aiohttp.client_exceptions import ClientResponseError
 
-from pulpcore.plugin.download import http_giveup, DownloaderFactory, HttpDownloader
+from pulpcore.plugin.download import http_giveup, DownloaderFactory, FileDownloader, HttpDownloader
 
 
 log = getLogger(__name__)
+
+
+class AnsibleFileDownloader(FileDownloader):
+    """
+    FileDownloader that strips out Ansible's custom http downloader arguments.
+
+    This is unfortunate, but there isn't currently a good pattern for customizing the downloader
+    factory machinery such that certain types of arguments only apply to certain downloaders,
+    so passing a kwarg into get_downloader() will pass it to constructor for any downloader.
+
+    TODO: https://pulp.plan.io/issues/7352
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initialize the downloader.
+        """
+        kwargs.pop("silence_errors_for_response_status_codes", None)
+        super().__init__(*args, **kwargs)
 
 
 class TokenAuthHttpDownloader(HttpDownloader):
@@ -19,13 +38,35 @@ class TokenAuthHttpDownloader(HttpDownloader):
 
     TOKEN_LOCK = asyncio.Lock()
 
-    def __init__(self, url, auth_url, token, **kwargs):
+    def __init__(
+        self, url, auth_url, token, silence_errors_for_response_status_codes=None, **kwargs
+    ):
         """
         Initialize the downloader.
         """
         self.ansible_auth_url = auth_url
         self.ansible_token = token
+        if silence_errors_for_response_status_codes is None:
+            silence_errors_for_response_status_codes = set()
+        self.silence_errors_for_response_status_codes = silence_errors_for_response_status_codes
         super().__init__(url, **kwargs)
+
+    def raise_for_status(self, response):
+        """
+        Raise error if aiohttp response status is >= 400 and not silenced.
+
+        Raises:
+            FileNotFoundError: If aiohttp response status is 404 and silenced.
+            aiohttp.ClientResponseError: If the response status is 400 or higher and not silenced.
+
+        """
+        silenced = response.status in self.silence_errors_for_response_status_codes
+
+        if not silenced:
+            response.raise_for_status()
+
+        if response.status == 404:
+            raise FileNotFoundError()
 
     @backoff.on_exception(backoff.expo, ClientResponseError, max_tries=10, giveup=http_giveup)
     async def _run(self, extra_data=None):
@@ -55,7 +96,7 @@ class TokenAuthHttpDownloader(HttpDownloader):
             headers = {"Authorization": "Bearer {token}".format(token=token)}
 
         async with self.session.get(self.url, headers=headers, proxy=self.proxy) as response:
-            response.raise_for_status()
+            self.raise_for_status(response)
             to_return = await self._handle_response(response)
             await response.release()
 
@@ -99,6 +140,7 @@ class AnsibleDownloaderFactory(DownloaderFactory):
             downloader_overrides = {
                 "http": TokenAuthHttpDownloader,
                 "https": TokenAuthHttpDownloader,
+                "file": AnsibleFileDownloader,
             }
         super().__init__(remote, downloader_overrides)
 
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -228,7 +228,7 @@ def _update_highest_version(collection_version):
 
 class AnsibleDeclarativeVersion(DeclarativeVersion):
     """
-    Subclassed Declarative version creates a custom pipeline for RPM sync.
+    Subclassed Declarative version creates a custom pipeline for Ansible sync.
     """
 
     def pipeline_stages(self, new_version):
@@ -251,6 +251,7 @@ class AnsibleDeclarativeVersion(DeclarativeVersion):
             ArtifactDownloader(),
             ArtifactSaver(),
             QueryExistingContents(),
+            DocsBlobDownloader(),
             CollectionContentSaver(),
             RemoteArtifactSaver(),
             ResolveContentFutures(),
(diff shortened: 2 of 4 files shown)


pulp_ansible/app/downloaders.py
```diff
@@ -6,27 +6,65 @@
...
-from pulpcore.plugin.download import http_giveup, DownloaderFactory, HttpDownloader
+from pulpcore.plugin.download import http_giveup, DownloaderFactory, FileDownloader, HttpDownloader
 
 
 log = getLogger(__name__)
 
 
+class AnsibleFileDownloader(FileDownloader):
+    """
+    FileDownloader that strips out RPM's custom http downloader arguments.
```
bmbouter, line 17: s/RPM/Ansible/

pulp_ansible/app/tasks/collections.py
```diff
@@ -296,13 +297,17 @@ async def run(self):
...
 
+                extradata = dict(docs_blob_url=metadata["docs_blob_url"])
+
                 d_artifact = DeclarativeArtifact(
                     artifact=Artifact(sha256=artifact["sha256"], size=artifact["size"]),
                     url=url,
                     relative_path=collection_version.relative_path,
                     remote=self.remote,
                     deferred_download=self.deferred_download,
+                    extra_data=extradata,
```
bmbouter, line 308: Since the docs blob data follows the `CollectionVersion` I believe this extra_data should live on `DeclarativeVersion` with [this](https://github.com/pulp/pulpcore/blob/b94abd64d76ea4554e6750ff38ce458eaa888cc8/pulpcore/plugin/stages/models.py#L117) instead.

bmbouter: This would basically move this to L#310

pulp_ansible/app/tasks/collections.py
```diff
@@ -378,12 +385,50 @@ def _build_url(path_or_url):
...
+    """
+
+    async def _handle_content_unit(self, d_content):
+        """Handle one content unit.
+
+        Returns:
+            The number of downloads
+        """
+        downloaders_for_content = []
+        for d_artifact in d_content.d_artifacts:
```
bmbouter, line 418: The move of extradata to `DeclarativeContent` I believe will remove the need for this loop.

pulp_ansible/app/tasks/collections.py
```diff
@@ -378,12 +385,50 @@ def _build_url(path_or_url):
...
+            docs_blob = d_artifact.extra_data.get("docs_blob_url")
+            if docs_blob:
+                downloader = d_artifact.remote.get_downloader(
+                    url=docs_blob, silence_errors_for_response_status_codes={404}
+                )
+                downloaders_for_content.append(downloader.run())
+
+        if downloaders_for_content:
+            await asyncio.gather(*downloaders_for_content)
```
bmbouter, line 427: Since you're downloading one thing itself, you probably don't need the `gather` here to "parallelize" multiple downloaders for a single content unit.

pulp_ansible/app/tasks/collections.py
```diff
@@ -378,12 +385,50 @@ def _build_url(path_or_url):
...
+                            }
 
                         if download_url:
+                            metadata = additional_metadata.get(_build_url(data["href"]), {})
+                            data["docs_blob_url"] = metadata.get("docs_blob_url")
                             yield data
                             progress_bar.increment()
 
 
+class DocsBlobDownloader(ArtifactDownloader):
```
bmbouter, line 399: First off, it's really great you've not duplicated the code of `ArtifactDownloader`. I can see now that pulpcore.plugin would do well to make this functionality more re-usable. I think the only issue I see is that the progress reporting from `ArtifactDownloader` is not customizable [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/plugin/stages/artifact_stages.py#L139).

So I wrote this for us to solve https://pulp.plan.io/issues/7464

I don't think the "incorrect progress reporting" problem should stop us from merging this and releasing 0.3.0. We should fix https://pulp.plan.io/issues/7464 with pulpcore 3.7 and switch `DocsBlogDownloader` to use `GenericDownloader` with the correct error message once 3.7 is GA.

bmbouter:
What's confusing to me here is that I can't see where `ArtifactDownloader` is saving the downloaded binary data to the in-memory Artifact. Similarly, I can't see where the `DocsBlobDownloader` is saving the response data into the `CollectionVersion` data. I expect the data would attached to the in-memory `DocsBlobDownloader` so that the `CollectionContentSaver` stage would have it and save it.

bmbouter:
Also does this code actually run as part of the Pulp-to-Pulp sync which is really the only place it could be tested?

pulp_ansible/app/tasks/collections.py
fao89: I would need to get the `DeclarativeArtifact` the same way, as the downloader comes from there

fao89:
> Also does this code actually run as part of the Pulp-to-Pulp sync which is really the only place it could be tested?

it runs for all syncs

bmbouter:
Do we have a cloud.redhat.com staging server sync test? It's behind the firewall so it would need to be `SkipTest` by default, but for testing this code it would be very useful. If we don't have one, would be willing to add one that is `SkipTest` by default and sync's from? https://ci.cloud.redhat.com/api/automation-hub/v3/collections/

pulp_ansible/app/tasks/collections.py
```diff
@@ -378,12 +388,54 @@ def _build_url(path_or_url):
...
+            remote = d_content.d_artifacts[0].remote
+            docs_blob = d_content.extra_data.get("docs_blob_url")
+            if docs_blob:
+                downloaded = 1
+                downloader = remote.get_downloader(
+                    url=docs_blob, silence_errors_for_response_status_codes={404}
+                )
+                try:
+                    download_result = await downloader.run()
+                    d_content.extra_data["docs_blob_file"] = download_result.path
```
bmbouter, line 431:
```suggestion
                    d_content.extra_data["docs_blob_file_path"] = download_result.path
```

For clarity when it's a string path I put `_path` in the name to disambiguate versus a file handle.

fao89:
Here is how I tested it:
```diff --git a/pulp_ansible/app/tasks/collections.py b/pulp_ansible/app/tasks/collections.py
index 3360913..f0b4a1b 100644
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -295,10 +295,10 @@ class CollectionSyncFirstStage(Stage):
                     version=metadata["version"],
                 )
 
-                artifact = metadata["artifact"]
+                # artifact = metadata["artifact"]
 
                 d_artifact = DeclarativeArtifact(
-                    artifact=Artifact(sha256=artifact["sha256"], size=artifact["size"]),
+                    artifact=Artifact(),
                     url=url,
                     relative_path=collection_version.relative_path,
                     remote=self.remote,
@@ -395,6 +395,7 @@ class CollectionSyncFirstStage(Stage):
                         if download_url:
                             metadata = additional_metadata.get(_build_url(data["href"]), {})
                             data["docs_blob_url"] = metadata.get("docs_blob_url")
+                            data["download_url"] = "https://galaxy.ansible.com/download/testing-k8s_demo_collection-0.0.3.tar.gz"
                             yield data
                             progress_bar.increment()
 
diff --git a/pulp_ansible/tests/functional/cli/test_collection_install.py b/pulp_ansible/tests/functional/cli/test_collection_install.py
index 08224b2..e648ea7 100644
--- a/pulp_ansible/tests/functional/cli/test_collection_install.py
+++ b/pulp_ansible/tests/functional/cli/test_collection_install.py
@@ -81,7 +81,7 @@ class InstallCollectionTestCase(unittest.TestCase):
             )
 
             subprocess.run(cmd.split())
-
+            import pdb; pdb.set_trace()
             self.assertTrue(path.exists(directory), "Could not find directory {}".format(directory))
 
     def test_install_collection(self):
```
```
(pulp) [vagrant@pulp3-source-fedora31 pulp_ansible]$ pytest pulp_ansible/tests/functional/cli/test_collection_install.py::InstallCollectionTestCase::test_install_collection_with_token
_from_ci_automation_hub
===================================== test session starts =====================================
platform linux -- Python 3.7.9, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /home/vagrant/devel/pulp_ansible
collected 1 item                                                                              

pulp_ansible/tests/functional/cli/test_collection_install.py 
>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/vagrant/devel/pulp_ansible/pulp_ansible/tests/functional/cli/test_collection_install.py(85)create_install_scenario()
-> self.assertTrue(path.exists(directory), "Could not find directory {}".format(directory))
(Pdb) 
```
```
In [1]: CollectionVersion.objects.all()
Out[1]: <QuerySet [<CollectionVersion: k8s_demo_collection>]>

In [2]: CollectionVersion.objects.first().docs_blob
Out[2]: 
{'docs_blob': {'contents': [],
  'collection_readme': {'html': '<p>This is collection_reqs_test collection</p>',
   'name': 'README.md'},
  'documentation_files': []}}
```

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -99,6 +100,21 @@ def test_install_collection_with_token_from_automation_hub(self):
         )
         self.create_install_scenario(body, TOKEN_DEMO_COLLECTION)
 
+    @skip_if(bool, "CI_AH_token", False)
```
bmbouter, line 103: Thank you for adding this.

bmbouter (approved):
Thanks @fao89 ! I think this is as good as we can test it given that no server-side environment can provide the data this sync is supposed to fetch.

PR #358, opened 2020-09-15
Sync collection deprecation status

fao89:
closes #7504

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -305,7 +305,10 @@ class CollectionSyncFirstStage(Stage):
                     deferred_download=self.deferred_download,
                 )
 
-                extradata = dict(docs_blob_url=metadata["docs_blob_url"])
+                extradata = dict(
+                    docs_blob_url=metadata["docs_blob_url"],
+                    deprecated=metadata["deprecated"],
+                )
 
                 d_content = DeclarativeContent(
                     content=collection_version,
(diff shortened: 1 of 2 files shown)


bmbouter:
@fao89 is there any public API that we can sync against that has this data today (for test assertion?)

pulp_ansible/app/tasks/collections.py
```diff
@@ -460,8 +474,9 @@ async def _pre_save(self, batch):
                 continue
 
             info = d_content.content.natural_key_dict()
+            deprecated = d_content.extra_data.get("deprecated", False)
             collection, created = Collection.objects.get_or_create(
-                namespace=info["namespace"], name=info["name"]
+                namespace=info["namespace"], name=info["name"], deprecated=deprecated
```
bmbouter, line 479: This code is right, but this `_pre_save` stage only saves when the unit is created the first time. This means that a situation like

1. Pulp sync's a collection and deprecated=False
2. time passes and the deprecation flag is marked as True
3. This code will never execute to change deprecated=False because the unit is already in Pulp

The issue is that content is immutable in Pulp and this content type is expecting mutability.

fao89: I can do a if..else to check if it is the first time creation

bmbouter: This issue description is the first problem, it's signalling a deeper issue an if/else won't fix. For example, if we did get it "updating" an existing unit with the deprecated flag once changed, that would be changed for every repository everywhere, which versions are intended to be immutable.

fao89:
closing in favor of: https://github.com/pulp/pulp_ansible/pull/366

PR #364, opened 2020-09-18
Deprecation status by repo

fao89:
closes #7504


pulp_ansible/app/models.py
```diff
@@ -253,6 +253,8 @@ class AnsibleRepository(Repository):
     CONTENT_TYPES = [Role, CollectionVersion]
     REMOTE_TYEPES = [RoleRemote, CollectionRemote]
 
+    extra_data = psql_fields.JSONField(default=dict, editable=False)
```
bmbouter, line 256: I had imagined this differently. I'll write it out here, but just to prompt some discussion. This proposal is designed to do more in the db and less in Python parsing that data.

1. Create a ManytoMany table that relates Collection and AnsibleRepository.
2. Have the table be "sparse" in the sense that if this table does not relate a Collection to that repository, it is assumed as deprecated=False. Conversly, if there is a relationship between Collection and a given Repository, it is deprecated=True.

The idea here is to rely on the DB's foreign key to take a Collection in the viewset and effeciently say if it's deprecated or not in the Galaxy serializer [here](https://github.com/pulp/pulp_ansible/blob/06c738ea17fcb0ebbf05a5ba3e16ace6ddb3d53a/pulp_ansible/app/galaxy/v3/serializers.py#L16). It leverages the assumption that there will be man fewer deprecations than there will be deprecations.

The reading of the remote API and the setting of this data would happen in FirstStage similarly to how this PR does it.

bmbouter:
Also this PR I think would remove the Collection.deprecated all together from everywhere. Perhaps we would have a data migration also that would create the new db records. Then again maybe not, and just let the sync data handle keeping it up to date...

fao89:
> Also this PR I think would remove the Collection.deprecated all together from everywhere. Perhaps we would have a data migration also that would create the new db records. Then again maybe not, and just let the sync data handle keeping it up to date...

I did not remove it, because I'm not sure about the regular viewset (the one that is not galaxy v1,v2, or v3), as it does not have the `base_path`

pulp_ansible/app/models.py
fao89: my idea was to use json field, for not requiring migration in the future, in case we need more mutable fields

fao89: based on your idea: https://github.com/pulp/pulp_ansible/pull/366

fao89:
closing in favor of https://github.com/pulp/pulp_ansible/pull/366

PR #366, opened 2020-09-22
Deprecation - collection x repository relation

fao89:


pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -48,6 +49,14 @@ def get_href(self, obj):
             kwargs={"path": self.context["path"], "namespace": obj.namespace, "name": obj.name},
         )
 
+    def get_deprecated(self, obj):
+        """Get href."""
+        distro = get_object_or_404(models.AnsibleDistribution, base_path=self.context["path"])
+        deprecation = distro.repository.cast().collection_memberships.filter(
+            collection_id=obj.collection.pk
+        )
+        return bool(deprecation.count())
```
bmbouter, line 58: I believe `exists()` https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.exists would be more efficient.

pulp_ansible/app/models.py
```diff
@@ -259,6 +257,20 @@ class Meta:
         permissions = (("modify_ansible_repo_content", "Can modify ansible repository content"),)
 
 
+class AnsibleRepositoryCollection(BaseModel):
```
bmbouter, line 260: You've basically done exactly what I was describing. The relating of a Collection to an `AnsibleRepository` seems right, but the name would be more appropriate if it was maybe `AnsibleCollectionDeprecated`. The presence of this record affects *only* the deprecated status.

### Thought experiment (this may be a terrible idea)

What happens if we get another piece of repo-level metadata, we'd end up creating another model type, and then another, etc. To avoid that architecture maybe we want to [use this pattern](https://docs.djangoproject.com/en/dev/topics/db/models/#extra-fields-on-many-to-many-relationships)

1) Leave the name as-is
2) Add the `deprecated` value defaulting to False

The major downside of this pattern is that you'll end up creating these relationships for every collection in every repo which to store an extremely sparse number of `deprecated=True` could be a bad idea performance wise.

I wanted to write this here, but likely we should just rename this model to `AnsibleCollectionDeprecated`. What do you think?

fao89: I think we can rename it for now, but in the future, if we need another repo-level metadata I would prefer to rename this table and add the deprecated field

pulp_ansible/app/viewsets.py
```diff
@@ -112,7 +112,6 @@ class CollectionVersionFilter(ContentFilter):
     name = filters.CharFilter(field_name="name")
     is_highest = filters.BooleanFilter(field_name="is_highest", method="get_highest")
     certification = MultipleChoiceFilter(choices=CollectionVersion.CERTIFICATION_CHOICES)
-    deprecated = filters.BooleanFilter(field_name="collection__deprecated")
```
bmbouter, line 115: I think this needs a `.removal` chanelog entry.

pulp_ansible/app/serializers.py
```diff
@@ -333,12 +333,6 @@ class CollectionVersionSerializer(SingleArtifactContentSerializer, ContentChecks
 
     version = serializers.CharField(help_text=_("The version of the collection."), max_length=32)
 
-    deprecated = serializers.BooleanField(
```
bmbouter, line 336: I think this needs a `.removal` chanelog entry.

pulp_ansible/app/models.py
```diff
@@ -259,6 +257,20 @@ class Meta:
         permissions = (("modify_ansible_repo_content", "Can modify ansible repository content"),)
 
 
+class AnsibleCollectionDeprecated(BaseModel):
+    """
+    A relationship between a Collection and an AnsibleRepository.
```
bmbouter, line 262:
```suggestion
    A model that represents if a Collection is `deprecated` for a given AnsibleRepository.
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -463,8 +495,16 @@ async def _pre_save(self, batch):
             collection, created = Collection.objects.get_or_create(
                 namespace=info["namespace"], name=info["name"]
             )
+            deprecated = d_content.extra_data.get("deprecated", False)
+            if deprecated:
```
newswangerd, line 499: This doesn't seem to account for cases where a collection gets undeprecated.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -108,6 +108,29 @@ class CollectionViewSet(
...
+        repository = distro.repository.cast()
+        deprecated_pks = repository.collection_memberships.values_list("collection_id", flat=True)
+
+        if deprecated_pks and deprecated:
+            queryset = queryset.filter(collection_id__in=deprecated_pks)
+        elif deprecated_pks:
+            queryset = queryset.exclude(collection_id__in=deprecated_pks)
+
+        return queryset
+
```
newswangerd, line 133: I'm worried about performance here. It looks like we have to basically pull all the deprecated collections and paste them into the query. Not sure how django handles this, but if we end up lots of deprecated collections, this seems like it could be a big drag on the database.

mdellweg:
Looking as an outsider here.
It seems to me that this "deprecation" is affecting a Collection in all (current and future) Versions of the repository. So my question is, whether a deprecation should create a new repoversion where the collection is marked while staying intact in the previous one.

fao89:
Closing, as we need to work more on the design: https://hackmd.io/@pulp/deprecation

PR #368, opened 2020-09-22
Derive ANSIBLE_CONTENT_HOSTNAME from CONTENT_ORIGIN

fao89:
https://pulp.plan.io/issues/7368
closes #7368


CHANGES/7368.misc
```diff
@@ -0,0 +1 @@
+Derive ANSIBLE_CONTENT_HOSTNAME from CONTENT_ORIGIN
```
bmbouter, line 1: The default of a variable I think needs a more user-facing changelog. Maybe even a removal type.

bmbouter:
Also the docs need to be updated here https://pulp-ansible.readthedocs.io/en/latest/settings.html#ansible-api-hostname

docs/settings.rst
```diff
@@ -21,9 +21,9 @@ ANSIBLE_CONTENT_HOSTNAME
    The origin, e.g. "http://example.com" that will instruct the client how to find the Pulp content
    app. This URL is formed in various Galaxy APIs (V1, V2, V3) responses.
 
-   This defaults to http on your fqdn, which is usable with the Ansible installer's default Nginx
-   configuration. By default it includes the ``pulp/content`` subpath. So if `example.com` is your
-   fqdn, this would default to "http://example.com/pulp/content".
+   This defaults to `CONTENT_ORIGIN <https://docs.pulpproject.org/pulpcore/settings.html?#content-origin>`_.
+   By default it includes the ``pulp/content`` subpath. So if `example.com` is your
```
bmbouter, line 25: Since CONTENT_ORIGIN is an origin, it also needs a protocol handler at the front in the example. It should also have a port, but since https is used the port is implied by the protocol handler. Origin is a technical portion of a URI

```suggestion
   By default it includes the ``pulp/content`` subpath. So if `https://example.com` is your
```

bmbouter (approved):
This looks right, thanks!

PR #369, opened 2020-09-22
Remove 'certification' flag from CollectionVersion

fao89:
https://pulp.plan.io/issues/6715
closes #6715


bmbouter:
This looks good to me. @newswangerd and @awcrosby can you verify this so we can merge it. This would come as a removal change in the 0.5.0 release.

bmbouter (approved):
Thank you @fao89 ! I saw on slack an lgtm from galaxy_ng also.

PR #370, opened 2020-09-23
Use amazon.aws collection to create a bucket

fao89:
This is my best guess for: https://travis-ci.com/github/pulp/pulp_ansible/jobs/389450601#L871

### DIFF
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -103,13 +103,9 @@ fi
 
 
 # Intall requirements for ansible playbooks
-pip install docker netaddr boto3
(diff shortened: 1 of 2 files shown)


mdellweg (approved):
Whenever you get that pass the CI. I am all for it.

.travis/before_install.sh
```diff
@@ -105,11 +105,7 @@ fi
 # Intall requirements for ansible playbooks
 pip install docker netaddr boto3
```
mdellweg, line 106: should this not install ansible, too?

bmbouter (approved):
Thank you for this!

PR #382, opened 2020-10-05
Deprecation - collection x repository_version relation

fao89:
closes #7504


fao89:
An alternative to: https://github.com/pulp/pulp_ansible/pull/366

PR #383, opened 2020-10-05
Only update bearer token after expiring the previous one

fao89:
https://pulp.plan.io/issues/7643
closes #7643

### DIFF
--- a/pulp_ansible/app/downloaders.py
+++ b/pulp_ansible/app/downloaders.py
@@ -37,6 +37,7 @@ class TokenAuthHttpDownloader(HttpDownloader):
     """
 
     TOKEN_LOCK = asyncio.Lock()
+    registry_auth = {"bearer": None}
 
     def __init__(
         self, url, auth_url, token, silence_errors_for_response_status_codes=None, **kwargs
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/downloaders.py
```diff
@@ -51,25 +52,31 @@ def __init__(
...
+
+        elif not silenced:
             response.raise_for_status()
 
         if response.status == 404:
             raise FileNotFoundError()
 
     @backoff.on_exception(backoff.expo, ClientResponseError, max_tries=10, giveup=http_giveup)
-    async def _run(self, extra_data=None):
+    async def _run(self, handle_401=True, extra_data=None):
```
bmbouter, line 79: Is there any place where the code calls with `handle_401=False`? Unless this param is in use I'd suggest to have this be the way the downloader works until configurability is needed.

bmbouter: Oh I see it is used further down...

pulp_ansible/app/downloaders.py
```diff
@@ -51,25 +52,31 @@ def __init__(
...
         Raises:
             FileNotFoundError: If aiohttp response status is 404 and silenced.
             aiohttp.ClientResponseError: If the response status is 400 or higher and not silenced.
 
         """
         silenced = response.status in self.silence_errors_for_response_status_codes
 
-        if not silenced:
+        if handle_401 and response.status == 401:
+            self.registry_auth["bearer"] = None
```
bmbouter, line 70: This shouldn't really be set here. Due to separation of concerns this should handle the silencing but not what happens if it's silenced. I gotta look closer to suggest what to do instead.

pulp_ansible/app/downloaders.py
```diff
@@ -91,12 +101,15 @@ async def _run(self, extra_data=None):
...
+            token = self.registry_auth["bearer"]
             # Keycloak Token
             headers = {"Authorization": "Bearer {token}".format(token=token)}
 
         async with self.session.get(self.url, headers=headers, proxy=self.proxy) as response:
-            self.raise_for_status(response)
+            self.raise_for_status(response, handle_401)
+            if response.status == 401:
+                await self.update_token_from_auth_url()
+                return await self._run(handle_401=False)
```
bmbouter, line 112: This could run a little unexpectedly in cases where a lot of other downloaders are running L#111 waits a long time, and then this one is the first one to wake up and it won't re-acquire the token even though it should.

fao89:
closing in favor of https://github.com/pulp/pulp_ansible/pull/406

PR #385, opened 2020-10-07
Collection Metadata Content - Deprecation

fao89:
closes #7504


fao89:
Downsides:
- needs a data migration for creating CollectionRepoMetadata
- the existing RepositoryVersions would not have CollectionRepoMetadata as content

PR #391, opened 2020-10-14
Improve regex to determine the API version

fao89:
The regex was expecting a slash after the API version, which may conflict with  trailing slash validation

closes #7711


fao89:
instead of `/api/v2` it should be `/api/v2/collections`

PR #394, opened 2020-10-26
Increase collection version size

daviddavis:
fixes #7745

PR #395, opened 2020-10-26
Fix ansible collection version > 32 chars

mcsSteam:
Ansible collections version references might be bigger
than 32 characters. As example the ansible community.
general version can be checked
https://galaxy.ansible.com/api/v2/collections/community/general/versions

closes #7745
https://pulp.plan.io/issues/7745


mcsSteam:
Is duplicate of #394

PR #396, opened 2020-10-26
Fix CI error by not downloading collection dependencies

fao89:
Ref: https://pulp.plan.io/issues/7751


fao89:
Pros:
- fixes the CI

Cons:
- does not fix the dependency issue

* with ansible 2.10, the use of dependencies will be more common, this is exactly what triggered this error.
`pulp.pulp_installer` collection didn't have dependency before and now it depends on `ansible.posix`

Alternative: https://github.com/pulp/pulp_ansible/pull/398

daviddavis (approved):

PR #403, opened 2020-10-29
Added validation and fixed handling of trailing slashes in URLs

daviddavis:
fixes #7767

### DIFF
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -116,6 +116,7 @@ class CollectionRemoteSerializer(RemoteSerializer):
         """
         Validate collection remote data.
 
+            - url: to ensure it ends with slashes.
             - requirements_file: to ensure it is a valid yaml file.
 
         Args:
@@ -128,10 +129,23 @@ class CollectionRemoteSerializer(RemoteSerializer):
             rest_framework.serializers.ValidationError: If the url or requirements_file is invalid
 
         """
+
+        def _validate_url(url):
+            """Based on https://git.io/JTMAA."""
+            if not url.endswith("/") and url != "https://galaxy.ansible.com":
+                raise serializers.ValidationError(
+                    _("Invalid URL {url}. Ensure the URL ends '/'.").format(url=data["url"])
+                )
+
         data = super().validate(data)
 
+        _validate_url(data["url"])
+
         if data.get("requirements_file"):
-            parse_collections_requirements_file(data["requirements_file"])
+            collections = parse_collections_requirements_file(data["requirements_file"])
+            for collection in collections:
+                if collection[2]:
+                    _validate_url(collection[2])
 
         return data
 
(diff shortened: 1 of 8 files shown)


pulp_ansible/app/serializers.py
```diff
@@ -128,10 +129,23 @@ def validate(self, data):
             rest_framework.serializers.ValidationError: If the url or requirements_file is invalid
 
         """
+
+        def _validate_url(url):
+            """Based on https://git.io/JTMAA."""
+            if not url.endswith("/") and url != "https://galaxy.ansible.com":
```
fao89, line 135: maybe we need a constant with URLs, I believe we need to add https://galaxy-dev.ansible.com

pulp_ansible/app/tasks/collections.py
```diff
@@ -336,18 +336,29 @@ async def _fetch_collections(self):
...
-                root = f"{root}/api/"
+                if root.endswith("/api/"):
+                    raise
+
+                root = urljoin(root, "api/")
                 downloader = remote.get_downloader(url=root)
                 api_data = parse_metadata(await downloader.run())
 
+            if "available_versions" not in api_data:
+                raise RuntimeError(_("Could not find 'available_versions' at {}").format(root))
```
fao89, line 360: this error message is pretty helpful, I like it!

pulp_ansible/tests/functional/api/collection/v2/test_sync.py
```diff
@@ -45,6 +44,33 @@ def setUpClass(cls):
...
         cls.client = api.Client(cls.cfg)
 
+    def test_remote_urls(self):
+        """This tests that the remote url ends with a "/".
+
+        This test targets https://pulp.plan.io/issues/7767
+
+        ansible-galaxy requires a trailing slash: https://git.io/JTMAA
+
+        It does make an exception for "https://galaxy.ansible.com": https://git.io/JTMpk
```
fao89, line 54: awesome!

fao89 (commented):
LGTM! Just waiting for my vagrant provisioning so I can test AH

pulp_ansible/app/serializers.py
daviddavis: In order to do that, I'd have to make [this logic](https://github.com/pulp/pulp_ansible/pull/403/files#diff-e6f9543dbfb3c4a3c4c71d8068b87a2388ef384e16d8b19fc61ec7c5e03d7919R344-R345) more complex. I don't think many users will use dev galaxy and if they do, they can use `https://galaxy-dev.ansible.com/api/` which is a link to the actual api.

daviddavis: This line is just sort of a nasty hack that I have to carry over from ansible-galaxy. See https://github.com/ansible/ansible/blob/a51a6f4a259b45593c3f803737c6d5d847258a83/lib/ansible/galaxy/api.py#L52-L53.

fao89 (approved):
Just tested and it fixes the CI

PR #414, opened 2020-11-04
Remove unnecessary is_highest filter method

fao89:
[noissue]

### DIFF
--- a/pulp_ansible/app/viewsets.py
+++ b/pulp_ansible/app/viewsets.py
@@ -1,6 +1,4 @@
-from collections import defaultdict
 from gettext import gettext as _
-from packaging.version import parse
 
 from django.contrib.postgres.search import SearchQuery
 from django.db.models import fields as db_fields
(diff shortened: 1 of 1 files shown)


pulp_ansible/app/viewsets.py
```diff
@@ -113,7 +111,7 @@ class CollectionVersionFilter(ContentFilter):
 
     namespace = filters.CharFilter(field_name="namespace")
     name = filters.CharFilter(field_name="name")
-    is_highest = filters.BooleanFilter(field_name="is_highest", method="get_highest")
+    is_highest = filters.BooleanFilter(field_name="is_highest")
```
bmbouter, line 114: Is it correct to say that we don't need this because the model contains it in the DB: https://github.com/pulp/pulp_ansible/blob/f4164bd007adc19e2dae5edb28d0dfc5ade885ce/pulp_ansible/app/models.py#L144

fao89: yes, we did a data migration in the past to assure that:
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/migrations/0005_collectionversion_is_highest.py

bmbouter (approved):
Thank you

PR #415, opened 2020-11-04
Field updated_at to use latest instead of highest version

fao89:
https://pulp.plan.io/issues/7775
closes #7775

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -5,6 +5,7 @@ import semantic_version
 
 from django.db.models import Exists, OuterRef
 from django.shortcuts import get_object_or_404
+from django.utils import timezone
 from django.utils.datastructures import MultiValueDictKeyError
 from django.utils.dateparse import parse_datetime
 from drf_spectacular.utils import OpenApiParameter, extend_schema
@@ -44,7 +45,9 @@ from pulp_ansible.app.galaxy.v3.pagination import LimitOffsetPagination
 from pulp_ansible.app.viewsets import CollectionVersionFilter
 
 
-CollectionTuple = namedtuple("CollectionTuple", ["namespace", "name", "version", "pulp_created"])
+CollectionTuple = namedtuple(
+    "CollectionTuple", ["namespace", "name", "version", "pulp_created", "updated_at"]
+)
 
 
 class AnsibleDistributionMixin:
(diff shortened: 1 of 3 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -200,6 +200,8 @@ def get_serializer_context(self, *args, **kwargs):
         super_data = super().get_serializer_context()
         super_data["highest_versions"] = self.highest_versions_context
```
bmbouter, line 201: The `highest_version` (and its computation) should be removed entirely also.

bmbouter: oh actually no it's needed for is_highest nevermind...

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -200,6 +200,8 @@ def get_serializer_context(self, *args, **kwargs):
...
         super_data["lowest_versions"] = self.lowest_versions_context
+        repo_version = self.get_repository_version(self.kwargs["path"])
+        super_data["repo_version_last_updated"] = repo_version.pulp_last_updated
```
bmbouter, line 204: This timestamp will move forward with each repo version published. That is what we want, but it also will move forward even when the CollectionVersions for this collection have not changed in between two repo versions. Is that ok? I'm a little unclear on how to query for the latter off the top of my head...

fao89: yes, I couldn't think of a good way to check the CollectionVersion change between RepositoryVersions

fao89: I think it would be too costly to make it precise, 
- latest_repo_version contains a new collection version? 
if no, get latest_repo_version - 1, ...

bmbouter:
I think this will correctly handle the case when content is added, but will it handle the case when content is removed? The queryset returning `pulp_last_updated` will disinclude those `CollectionVersion` objects that have been removed from the repository which also constitutes a change.

For example, say we have a single repo version published at time 1111 with two CollectionVersions of one Collection in it. Then someone removes one of these two CollectionVersions in a new repo version at 2222. I think the updated timestamp will still show 1111.

Unfortunately I think it'll need another query to go back to the db for specifically this info because the one I built only considers content in the current repo version, not all content that was ever in the repo version history...

fao89:
closing in favor of: https://github.com/pulp/pulp_ansible/pull/416

PR #416, opened 2020-11-05
Field updated_at to use latest instead of highest version

fao89:
https://pulp.plan.io/issues/7775
closes #7775

### DIFF
--- a/CHANGES/7775.bugfix
+++ b/CHANGES/7775.bugfix
@@ -1,0 +1,1 @@
+Field ``updated_at`` from Galaxy v3 Collections endpoint using latest instead of highest version
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -52,8 +52,10 @@ class CollectionSerializer(serializers.ModelSerializer):
 
     def get_updated_at(self, obj):
         """Get the timestamp of the highest version CollectionVersion's created timestamp."""
-        collection = self.context["highest_versions"][obj.pk]
-        return collection.pulp_created
+        if obj.repo_version_added_at and obj.repo_version_removed_at:
+            return max(obj.repo_version_added_at, obj.repo_version_removed_at)
+
+        return obj.repo_version_added_at or obj.repo_version_removed_at
 
     @extend_schema_field(OpenApiTypes.OBJECT)
     def get_highest_version(self, obj):
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from gettext import gettext as _
 import semantic_version
 
-from django.db.models import Exists, OuterRef
+from django.db.models import Exists, OuterRef, Subquery, Max
 from django.shortcuts import get_object_or_404
 from django.utils.datastructures import MultiValueDictKeyError
 from django.utils.dateparse import parse_datetime
@@ -17,7 +17,7 @@ from rest_framework import status as http_status
 from rest_framework import viewsets
 
 from pulpcore.plugin.exceptions import DigestValidationError
-from pulpcore.plugin.models import PulpTemporaryFile, Content, ContentArtifact
+from pulpcore.plugin.models import PulpTemporaryFile, Content, ContentArtifact, RepositoryVersion
 from pulpcore.plugin.serializers import AsyncOperationResponseSerializer
 
 from pulp_ansible.app.galaxy.v3.exceptions import ExceptionHandlerMixin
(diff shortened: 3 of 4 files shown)


pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -52,8 +54,22 @@ def get_created_at(self, obj):
 
     def get_updated_at(self, obj):
         """Get the timestamp of the highest version CollectionVersion's created timestamp."""
-        collection = self.context["highest_versions"][obj.pk]
-        return collection.pulp_created
+        added = Q(
```
bmbouter, line 57: One of the new requirements from galaxy_ng for performance is to not perform queries here, but instead in the get_queryset method. This can be done with the `annotate` approach that is in use there. The issue with doing it here is that if you have N items in the list view it goes to the database N times, which they no longer want. I learned about this while doing the `deprecated` work.

CHANGES/7775.bugfix
```diff
@@ -0,0 +1 @@
+Field updated_at from Galaxy v3 Collections endpoint using latest instead of highest version
```
bmbouter, line 1:
```suggestion
Field ``updated_at`` from Galaxy v3 Collections endpoint using latest instead of highest version
```

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -52,8 +52,7 @@ def get_created_at(self, obj):
 
     def get_updated_at(self, obj):
         """Get the timestamp of the highest version CollectionVersion's created timestamp."""
-        collection = self.context["highest_versions"][obj.pk]
-        return collection.pulp_created
+        return obj.repo_version_updated_at
```
bmbouter, line 55: :+1:

bmbouter (commented):
@fao89 would you be willing to write a test for this?

Maybe one that creates a repository version and sync's from a source that brings in two collectionVersions of a collection. Then use the /modify/ endpoint to remove one of them and assert the updated_at increases from repoversion 1 to repoversion 2?

fao89:
> @fao89 would you be willing to write a test for this?
> 
> Maybe one that creates a repository version and sync's from a source that brings in two collectionVersions of a collection. Then use the /modify/ endpoint to remove one of them and assert the updated_at increases from repoversion 1 to repoversion 2?

Thank you for suggesting the test, I could catch the bug on my query

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
```diff
@@ -0,0 +1,94 @@
+# coding=utf-8
```
bmbouter, line 1: This line should be removed. Python3 knows source files are utf-8

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
```diff
@@ -0,0 +1,94 @@
...
+"""Tests related to sync ansible plugin collection content type."""
```
bmbouter, line 2: Maybe this docstring gets updated to focus on serializers?

fao89: I forgot the docstrings, I'll fix it

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -153,8 +153,26 @@ def get_queryset(self):
...
+            complete=True,
+            added_memberships__content__in=OuterRef("versions__pk"),
+        ).values("pulp_last_updated")
+        removed_repo_version_query = RepositoryVersion.objects.filter(
+            repository=repo_version.repository,
+            complete=True,
+            removed_memberships__content__in=OuterRef("versions__pk"),
+        ).values("pulp_last_updated")
+
+        collections = (
```
bmbouter, line 167: This line reads a bit confusing because I keep thinking it is a tuple definition but it's not. It's valid Python though. Consider maybe:

```
collections_qs = Collection.objects.filter(versions__in=repo_version.content).distinct()
collections_qs =  collection_qs.annotate(
    repo_version_added_at=Max(Subquery(added_repo_version_query)),
    repo_version_removed_at=Max(Subquery(removed_repo_version_query)),
    deprecated=Exists(deprecated_query),
)
```

```
collections_qs = Collection.objects.filter(versions__in=repo_version.content).distinct()
collections_qs_annotated =  collection_qs.annotate(
    repo_version_added_at=Max(Subquery(added_repo_version_query)),
    repo_version_removed_at=Max(Subquery(removed_repo_version_query)),
    deprecated=Exists(deprecated_query),
)
```

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
```diff
@@ -0,0 +1,93 @@
...
+
+        collections = self.collections_api.list(self.base_path)
+        highest_version = collections.data[0].highest_version["version"]
+        updated_at = datetime.strptime(collections.data[0].updated_at, "%Y-%m-%dT%H:%M:%S.%fZ")
+
+        total_versions = self.collections_versions_v3api.list(
+            "pulp_installer", "pulp", self.base_path
+        ).meta["count"]
+
+        self.assertEqual(highest_version, original_highest_version)
```
bmbouter, line 91: I think the assertion for `highest_version` should move to its own test or somehow be removed from this test. This test is labelled as only being about `updated_at` but really deals with both fields. Rather than relabelling it, having each test deal with just one thing I think would be better.

bmbouter (changes requested):
@fao89 This code works perfectly, great job. I requested one readability change and one test assertion change. Wdyt about those?

pulp_ansible/app/galaxy/v3/views.py
fao89: I also didn't like it, it was black, I'll improve it

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
fao89: I put it here to assure `updated_at` didn't change because of a new `highest_version`

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -153,8 +153,23 @@ def get_queryset(self):
...
+            complete=True,
+            removed_memberships__content__in=OuterRef("versions__pk"),
+        ).values("pulp_last_updated")
+
+        collections_qs_annotated = Collection.objects.annotate(
+            repo_version_added_at=Max(Subquery(added_repo_version_query)),
+            repo_version_removed_at=Max(Subquery(removed_repo_version_query)),
+            deprecated=Exists(deprecated_query),
+        )
+        collections = collections_qs_annotated.filter(versions__in=repo_version.content).distinct()
```
fao89, line 172: When filter comes first the test fails, I believe it is because the subqueries use collection versions that are not on the current repo version.

bmbouter: Yup that makes sense. Well breaking it up to not use parens I think would be better because of readability.

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
bmbouter: Ok

bmbouter (approved):
Thanks!

PR #418, opened 2020-11-05
Adds validation to `CollectionRemoteSerializer`

bmbouter:
When using either the `token` or `auth_url` attributes, ensure the user
is using both.

closes #7802

### DIFF
--- a/pulp_ansible/app/downloaders.py
+++ b/pulp_ansible/app/downloaders.py
@@ -49,7 +49,7 @@ class TokenAuthHttpDownloader(HttpDownloader):
         Initialize the downloader.
         """
         self.ansible_auth_url = auth_url
-        self.ansible_token = token
+        self.token = token
         if silence_errors_for_response_status_codes is None:
             silence_errors_for_response_status_codes = set()
         self.silence_errors_for_response_status_codes = silence_errors_for_response_status_codes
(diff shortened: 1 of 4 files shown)


fao89 (approved):
Thank you!

newswangerd:
@bmbouter @fao89 I don't mean to be a party pooper here, but this change will prevent hub from syncing from other hub instances or from authenticating to galaxy.ansible.com. The ONLY place the `auth_url` is used for is cloud.redhat.com. Token is a requirement for `auth_url`, but you can use a token without specifying `auth url`

pulp_ansible/app/downloaders.py
```diff
@@ -83,18 +83,10 @@ async def _run(self, extra_data=None):
         This method provides the same return object type and documented in
         :meth:`~pulpcore.plugin.download.BaseDownloader._run`.
 
-        Ansible token reference:
-            https://github.com/ansible/ansible/blob/devel/lib/ansible/galaxy/token.py
-
         """
-        if not self.ansible_token:
+        if not self.token:
```
fao89, line 87: I believe doing 
```python
if not self.ansible_auth_url:
```
would cover @newswangerd comment.
We need to file an issue for this

bmbouter:
Those use cases make sense. No party pooperness perceived.

PR #423, opened 2020-11-10
Refactor CollectionSync to use asyncio more

bmbouter:
[noissue]

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -3,18 +3,20 @@ from aiohttp.client_exceptions import ClientResponseError
 from gettext import gettext as _
 import json
 import logging
-import math
 import tarfile
-from urllib.parse import urljoin, urlparse, urlunparse
+from urllib.parse import urljoin
 
+from async_lru import alru_cache
 from django.db import transaction
 from django.db.models import Q
 from django.urls import reverse
 from galaxy_importer.collection import import_collection as process_collection
 from galaxy_importer.collection import CollectionFilename
 from galaxy_importer.exceptions import ImporterError
+from pkg_resources import Requirement
 from rq.job import get_current_job
 
+from pulpcore.plugin.constants import TASK_STATES
 from pulpcore.plugin.models import (
     Artifact,
     ContentArtifact,
@@ -50,9 +52,9 @@ from pulp_ansible.app.models import (
 )
 from pulp_ansible.app.serializers import CollectionVersionSerializer
 from pulp_ansible.app.tasks.utils import (
-    get_page_url,
     parse_metadata,
     parse_collections_requirements_file,
+    RequirementsFileEntry,
 )
 
 
@@ -293,6 +295,8 @@ class CollectionSyncFirstStage(Stage):
 
         """
         super().__init__()
+        msg = _("Parsing CollectionVersion Metadata")
+        self.parsing_metadata_progress_bar = ProgressReport(message=msg, code="parsing.metadata")
         self.remote = remote
         self.collection_info = parse_collections_requirements_file(remote.requirements_file)
         self.deprecations = Q()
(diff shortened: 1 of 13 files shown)


bmbouter:
My testing of a repo mirror that has 5 collections in it shows the new implementation is 9.2% faster. I expect it to perform better the more collections there are for it to fetch because the API fetching occurs more in parallel in this implementation.

pulp_ansible/app/tasks/collections.py
```diff
@@ -293,206 +295,219 @@ def __init__(self, remote):
...
+        self.parsing_metadata_progress_bar.increment()
+        await self.put(d_content)
 
-            if "v3" in api_data.get("available_versions", {}):
-                self.api_version = 3
-            elif "v2" in api_data.get("available_versions", {}):
-                self.api_version = 2
+    def _collection_versions_list_downloader(
+        self, api_version, collection_endpoint, namespace, name, page_num, page_size
+    ):
```
fao89, line 387: sweet!

pulp_ansible/app/tasks/collections.py
```diff
@@ -293,206 +295,219 @@ def __init__(self, remote):
...
+        else:
+            offset = (page_num - 1) * page_size
+            versions_list_url = f"{url_without_get_params}?limit={page_size}&offset={offset}"
+        return self.remote.get_downloader(url=versions_list_url)
+
+    async def _fetch_collection_metadata(self, requirements_entry):
+        if requirements_entry.version == "*":
+            requirement_version = Requirement.parse("collection")
+        else:
+            requirement_version = Requirement.parse(f"collection{requirements_entry.version}")
```
fao89, line 400: pretty cool!

pulp_ansible/app/tasks/collections.py
```diff
@@ -293,206 +295,219 @@ def __init__(self, remote):
...
+            page_num = page_num + 1
+
+        await asyncio.gather(*tasks)
+
+    @staticmethod
+    def _get_response_next_value(api_version, response):
+        if api_version == 2:
+            return response["next"]
+        else:
+            return response["links"]["next"]
```
fao89, line 448: wondering if it can raise KeyError

pulp_ansible/app/tasks/utils.py
```diff
@@ -39,6 +40,9 @@ def parse_metadata(download_result):
         return json.load(fd)
 
 
+RequirementsFileEntry = namedtuple("RequirementsFileEntry", ["name", "version", "source"])
```
fao89, line 43: cool!

pulp_ansible/tests/functional/utils.py
```diff
@@ -150,6 +157,21 @@ def wait_tasks():
...
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.client = gen_ansible_client()
+        cls.repo_api = RepositoriesAnsibleApi(cls.client)
+        cls.remote_collection_api = RemotesCollectionApi(cls.client)
+        cls.distributions_api = DistributionsAnsibleApi(cls.client)
+        cls.cv_api = ContentCollectionVersionsApi(cls.client)
+        cls.collections_v3api = PulpAnsibleGalaxyApiCollectionsApi(cls.client)
+        cls.collections_versions_v3api = PulpAnsibleGalaxyApiV3VersionsApi(cls.client)
```
fao89, line 172: Nice!

fao89 (approved):
This is great!
This PR makes it easier to read the sync code, thank you @bmbouter

pulp_ansible/app/tasks/collections.py
bmbouter: I don't expect it to for V2 or V3. If it does though, I think not catching it is right because it would be fatal.

fao89: I agree it would be fatal, but I wonder if the error message would be clear enough

bmbouter: Those keys are expected for both V2 and V3 so I don't think we need to take special care for such an exceptional situation. Generally I think exception catching is for situations where you expect to experience that exception.

daviddavis (approved):
LGTM :+1:

bmbouter:
I added some changelog, a better git commit log, and rebased to one commit

PR #424, opened 2020-11-11
Fix CollectionRemote auth_url validation

fao89:
https://pulp.plan.io/issues/7821
closes #7821

### DIFF
--- a/CHANGES/7821.bugfix
+++ b/CHANGES/7821.bugfix
@@ -1,0 +1,1 @@
+Ensure that when creating a ``CollectionRemote`` you can use ``token`` without specifying ``auth_url``
--- a/pulp_ansible/app/downloaders.py
+++ b/pulp_ansible/app/downloaders.py
@@ -84,9 +84,11 @@ class TokenAuthHttpDownloader(HttpDownloader):
         :meth:`~pulpcore.plugin.download.BaseDownloader._run`.
 
         """
-        if not self.token:
-            # No Token
+        if not self.token and not self.ansible_auth_url:
             return await super()._run(extra_data=extra_data)
+        elif self.token:
+            headers = {"Authorization": "Bearer {token}".format(token=self.token)}
+            return await self._run_with_additional_headers(headers)
         else:
             return await self._run_with_token_refresh_and_401_retry()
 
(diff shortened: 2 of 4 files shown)


pulp_ansible/app/downloaders.py
```diff
@@ -84,8 +84,8 @@ async def _run(self, extra_data=None):
         :meth:`~pulpcore.plugin.download.BaseDownloader._run`.
 
         """
-        if not self.token:
-            # No Token
+        if not self.ansible_auth_url:
```
bmbouter, line 87: If there is no auth_url, but there is a token I believe we need to use that. I think that would be a call to _run_with_additional_headers because the token is the extra Authorization header that is needed.

If there is no token and no auth_url then call return await super()._run(extra_data=extra_data).

If there is both a token and auth_url call self._run_with_token_refresh_and_401_retry()

pulp_ansible/tests/functional/api/collection/test_remote.py
```diff
@@ -23,10 +23,11 @@ def test_remote_with_url_only_is_allowed(self):
         remote = self.remote_collection_api.create(body)
         self.addCleanup(self.remote_collection_api.delete, remote.pulp_href)
 
-    def test_token_requires_auth_url(self):
+    def test_token_only_is_allowed(self):
         """Assert that a `CollectionRemote` with `token` and no `auth_url` can't be created."""
```
bmbouter, line 27: This line needs updating

pulp_ansible/app/downloaders.py
```diff
@@ -84,9 +84,12 @@ async def _run(self, extra_data=None):
         :meth:`~pulpcore.plugin.download.BaseDownloader._run`.
 
         """
-        if not self.token:
-            # No Token
+        if not self.token and not self.ansible_auth_url:
+            # No Token and no auth_url
```
bmbouter, line 88: This comment can probably be taken out

bmbouter:
The implementation looks great, I left two small comment requests

CHANGES/7821.bugfix
```diff
@@ -0,0 +1 @@
+Ensure that when creating a ``CollectionRemote`` with ``auth_url`` that you use ``token`` together.
```
bmbouter, line 1: Sorry one more thing I didn't see. I think this should read differently.
```suggestion
Ensure that when creating a ``CollectionRemote`` you can use ``token`` without specifying ``auth_url``.
```

bmbouter (approved):
This looks great, thank you!

PR #426, opened 2020-11-13
Fix semver comparison

fao89:
https://pulp.plan.io/issues/7826
closes #7826

### DIFF
--- a/CHANGES/7826.bugfix
+++ b/CHANGES/7826.bugfix
@@ -1,0 +1,2 @@
+Fix version comparisons during sync and upload when comparing the same version with different build
+numbers.
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -35,7 +35,7 @@ from pulpcore.plugin.stages import (
     QueryExistingArtifacts,
     QueryExistingContents,
 )
-import semantic_version as semver
+from semantic_version import Version
 
 from pulp_ansible.app.constants import PAGE_SIZE
 from pulp_ansible.app.models import (
@@ -220,7 +220,7 @@ def _update_highest_version(collection_version):
     if not last_highest:
         collection_version.is_highest = True
         return None
-    if semver.compare(collection_version.version, last_highest.version) > 0:
+    if Version(collection_version.version) > Version(last_highest.version):
         last_highest.is_highest = False
         collection_version.is_highest = True
         last_highest.save()
--- a/pulp_ansible/tests/functional/constants.py
+++ b/pulp_ansible/tests/functional/constants.py
@@ -84,4 +84,5 @@ collections:
   source: https://galaxy-dev.ansible.com/
 - testing.k8s_demo_collection
 - rshad.collection_demo
+- brightcomputing.bcm
 """


pulp_ansible/tests/functional/constants.py
```diff
@@ -83,4 +83,5 @@
   version: ">=1.0.0,<=2.0.0"
   source: https://galaxy-dev.ansible.com/
 - testing.k8s_demo_collection
+- brightcomputing.bcm
```
bmbouter, line 86: Where is this used?

bmbouter: Oh wait this is what is being sync'd to ensure the code is working, is that right?

bmbouter:
We need to think a bit more about this I believe. I think using semver as a library is right actually because technically semver is the controlling versioning scheme. Also the "build number" portion of semver (after the plus) cannot be used to indicate newer versus older builds, see the `have the same precedence` section [here](https://semver.org/#spec-item-10). That is why we get this:

```
In [1]: import semantic_version as semver

In [2]: semver.compare("9.1.0-dev.1+40688.git966d0b3", "9.1.0-dev.1+40390.gita455009")
Out[2]: NotImplemented
```

So if comparison is not allowed, then I expect one to not be "higher" than the other and that means either one can be is_highest arbitrarily. They are literally the same verison just with different build metadata.

My suggestion then is to:
1) keep using the semver parsing library
2) Catch the `NotImplemented` error case and arbitrarily pick one to mark as is_highest.

pulp_ansible/tests/functional/constants.py
fao89: exactly! to assure semver comparison is working

fao89:
@bmbouter with pulp_installer we noticed galaxy was not handling our bugfixes releases well,
so it motivated me to use LooseVersion
```python
In [1]: import semantic_version as semver

In [2]: semver.Version("3.8.0") < semver.Version("3.8.0-1")
Out[2]: False

In [3]: semver.Version("3.8.0") > semver.Version("3.8.0-1")
Out[3]: True

In [4]: from distutils.version import LooseVersion

In [5]: LooseVersion("3.8.0") < LooseVersion("3.8.0-1")
Out[5]: True

In [6]: LooseVersion("3.8.0") > LooseVersion("3.8.0-1")
Out[6]: False

```

pulp_ansible/app/tasks/collections.py
```diff
@@ -220,7 +220,7 @@ def _update_highest_version(collection_version):
     if not last_highest:
         collection_version.is_highest = True
         return None
-    if semver.compare(collection_version.version, last_highest.version) > 0:
+    if Version(collection_version.version) > Version(last_highest.version):
```
fao89, line 223:
```python
In [1]: import semantic_version as semver

In [2]: semver.compare("9.1.0-dev.1+40688.git966d0b3", "9.1.0-dev.1+40390.gita455009")
Out[2]: NotImplemented

In [3]: semver.Version("9.1.0-dev.1+40688.git966d0b3") > semver.Version("9.1.0-dev.1+40390.gita455009")
Out[3]: False

In [4]: semver.Version("9.1.0-dev.1+40688.git966d0b3") < semver.Version("9.1.0-dev.1+40390.gita455009")
Out[4]: False

In [5]: semver.Version("9.1.0-dev.1+40688.git966d0b3") == semver.Version("9.1.0-dev.1+40390.gita455009")
Out[5]: False
```

bmbouter: This is looking good to me. Now we're asking it a question it can answer!

CHANGES/7826.bugfix
```diff
@@ -0,0 +1 @@
+Fix semver comparison
```
bmbouter, line 1: Can this describe the sitaution more?

fao89: Fix semver comparison for build numbers?

bmbouter: Maybe:     `Fix version comparisons during sync and upload when comparing the same version with different build numbers.`

bmbouter (approved):
This looks great to me. Thanks!

PR #427, opened 2020-11-13
Improve MANIFEST.json handling and provide better error message

fao89:
https://pulp.plan.io/issues/5745
closes #5745

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -597,7 +597,15 @@ class CollectionContentSaver(ContentSaver):
                     fileobj=artifact_file, mode="r"
                 ) as tar:
                     log.info(_("Reading MANIFEST.json from {path}").format(path=artifact.file.name))
-                    file_obj = tar.extractfile("MANIFEST.json")
+                    for manifest in ["MANIFEST.json", "./MANIFEST.json"]:
+                        try:
+                            file_obj = tar.extractfile(manifest)
+                        except KeyError:
+                            file_obj = None
+                        else:
+                            break
+                    if not file_obj:
+                        raise FileNotFoundError("MANIFEST.json not found")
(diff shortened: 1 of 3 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -597,7 +597,15 @@ async def _post_save(self, batch):
                     fileobj=artifact_file, mode="r"
                 ) as tar:
                     log.info(_("Reading MANIFEST.json from {path}").format(path=artifact.file.name))
-                    file_obj = tar.extractfile("MANIFEST.json")
+                    for manifest in ["MANIFEST.json", "./MANIFEST.json"]:
```
bmbouter, line 600: What case causes us to also check `./MANIFEST.json` ?

daviddavis: If you look at the bug, there's one collection version tar that has `./MANIFEST.json`:

```
❯ tar -ztvf rshad-collection_demo-1.0.2.tar.gz                               
-rw-r--r-- rashz/rashz   10373 2019-11-15 11:14 ./CHANGELOG.md
-rw-r--r-- rashz/rashz   34129 2019-11-15 12:11 ./FILES.json
-rw-r--r-- rashz/rashz   24736 2019-11-15 11:14 ./LICENSE
-rw-r--r-- rashz/rashz     876 2019-11-15 12:11 ./MANIFEST.json
...
```

fao89: idk how tar can have `./MANIFEST.json`, but I thought it would not be cool to throw an error for this case

daviddavis: If you run tar with a path like `./*` (eg `tar -czf my.tgz ./*`) then the filename inside the tar get prepended with `./`.

bmbouter (approved):
This looks great, thank you

PR #428, opened 2020-11-16
Avoid to download docs-blob when content is already saved

fao89:
https://pulp.plan.io/issues/7831
closes #7831


bmbouter (approved):
This looks great!

PR #429, opened 2020-11-16
Stop making requests to docs-blob endpoint on Galaxy v2

fao89:
https://pulp.plan.io/issues/7830
closes #7830


bmbouter (approved):
This looks right to me. I'll also make this change in my new sync code too.

PR #430, opened 2020-11-18
Ensure deprecation status is synced even when no content changes

fao89:
https://pulp.plan.io/issues/7834
closes #7834

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -41,7 +41,7 @@ from pulp_ansible.app.serializers import (
 
 from pulp_ansible.app.galaxy.mixins import UploadGalaxyCollectionMixin
 from pulp_ansible.app.galaxy.v3.pagination import LimitOffsetPagination
-from pulp_ansible.app.viewsets import CollectionVersionFilter
+from pulp_ansible.app.viewsets import CollectionFilter, CollectionVersionFilter
 
 
 CollectionTuple = namedtuple("CollectionTuple", ["namespace", "name", "version", "pulp_created"])
@@ -123,6 +123,7 @@ class CollectionViewSet(
     authentication_classes = []
     permission_classes = []
     serializer_class = CollectionSerializer
+    filterset_class = CollectionFilter
     pagination_class = LimitOffsetPagination
 
     def filter_queryset(self, queryset):
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -234,7 +234,7 @@ class CollectionViewSet(
             AnsibleCollectionDeprecated.objects.filter(
                 repository_version=repo_version, collection=collection
             ).delete()
-        return Response()
+        return Response(serializer.data)
 
 
 class CollectionUploadViewSet(
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -142,7 +142,8 @@ class CollectionRemoteSerializer(RemoteSerializer):
 
         data = super().validate(data)
 
-        _validate_url(data["url"])
+        if data.get("url"):
+            _validate_url(data["url"])
 
         if data.get("requirements_file"):
             collections = parse_collections_requirements_file(data["requirements_file"])
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -8,6 +8,7 @@ import tarfile
 from urllib.parse import urljoin, urlparse, urlunparse
 
 from django.db import transaction
+from django.db.models import Q
 from django.urls import reverse
 from galaxy_importer.collection import import_collection as process_collection
 from galaxy_importer.collection import CollectionFilename
@@ -80,6 +81,23 @@ def sync(remote_pk, repository_pk, mirror):
     first_stage = CollectionSyncFirstStage(remote)
     d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
     d_version.create()
+
+    repo_version = repository.latest_version()
+    to_deprecate = []
+    if first_stage.deprecations:
+        collections_deprecate_true_qs = Collection.objects.filter(first_stage.deprecations)
+        for collection in collections_deprecate_true_qs:
+            to_deprecate.append(
+                AnsibleCollectionDeprecated(repository_version=repo_version, collection=collection)
+            )
+
+        AnsibleCollectionDeprecated.objects.bulk_create(to_deprecate, ignore_conflicts=True)
+
+    non_deprecated_qs = Collection.objects.exclude(first_stage.deprecations)
+    AnsibleCollectionDeprecated.objects.filter(
+        repository_version=repo_version,
+        collection__pk__in=non_deprecated_qs,
+    ).delete()
 
 
 def import_collection(
@@ -277,6 +295,7 @@ class CollectionSyncFirstStage(Stage):
         super().__init__()
         self.remote = remote
         self.collection_info = parse_collections_requirements_file(remote.requirements_file)
+        self.deprecations = Q()
 
         # Interpret download policy
         self.deferred_download = self.remote.policy != Remote.IMMEDIATE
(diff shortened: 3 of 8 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -123,6 +123,7 @@ class CollectionViewSet(
     authentication_classes = []
     permission_classes = []
     serializer_class = CollectionSerializer
+    filterset_class = CollectionFilter
```
bmbouter, line 126: what motivated this change?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -234,7 +235,7 @@ def update(self, request, *args, **kwargs):
             AnsibleCollectionDeprecated.objects.filter(
                 repository_version=repo_version, collection=collection
             ).delete()
-        return Response()
+        return Response(serializer.data)
```
bmbouter, line 238: what motivated this change?

pulp_ansible/app/serializers.py
```diff
@@ -142,7 +142,8 @@ def _validate_url(url):
 
         data = super().validate(data)
 
-        _validate_url(data["url"])
+        if data.get("url"):
```
bmbouter, line 145: what motivated this change?

pulp_ansible/app/galaxy/v3/views.py
fao89: the bindings client was raising a validation error because it was expecting a serialized collection - https://pulp-ansible.readthedocs.io/en/latest/restapi.html#operation/pulp_ansible_galaxy_api_v3_collections_update

pulp_ansible/app/galaxy/v3/views.py
fao89: I wanted to filter by name/namespace on my test

pulp_ansible/app/serializers.py
fao89: when doing a PATCH it was breaking because no URL was being sent

pulp_ansible/app/tasks/collections.py
```diff
@@ -81,6 +81,26 @@ def sync(remote_pk, repository_pk, mirror):
     d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
     d_version.create()
 
+    repo_version = repository.latest_version()
+    collections = Collection.objects.filter(versions__in=repo_version.content).distinct()
+    to_deprecate = []
+    to_undeprecate = []
+    for collection in collections:
```
bmbouter, line 88: I had imagined this part differently. Instead of iterating through the db's collection's you could filter them with a `Q(namespace==..., name=...)` and || those all together to just get the subset of collections that need to have `AnsibleCollectionDeprecated` objects created, then use the bulk_create with ignore_conflicts to create them.

For deletion I was thinking take the set of collections used in the creation, say that is `collections_deprecate_true_qs` then do something like `AnsiblecollectionDeprecated.objects.exclude(collection__pk__in=collections_deprecate_true_qs).delete()`.

pulp_ansible/app/tasks/collections.py
```diff
@@ -277,6 +297,7 @@ def __init__(self, remote):
         super().__init__()
         self.remote = remote
         self.collection_info = parse_collections_requirements_file(remote.requirements_file)
+        self.deprecations = {}
```
bmbouter, line 300: :+1:

pulp_ansible/app/galaxy/v3/views.py
bmbouter: That makes sense, but then this would be introducing a new feature in the z-stream. Also it needs a changelog when introduced too.

pulp_ansible/app/galaxy/v3/views.py
bmbouter: probably a changelog for this. :+1:

fao89:
@bmbouter after reading your comments, I'll split into two commits: bugfix, features for facilitating the test, so we would cherry-pick the fix to 0.5, and the features would only stay in master.

pulp_ansible/app/tasks/collections.py
fao89: I tested locally, doing:
```python
AnsiblecollectionDeprecated.objects.exclude(collection__pk__in=collections_deprecate_true_qs).delete()
```
would not cover the case where no collection was deprecated and a collection from the previous sync was undeprecated

pulp_ansible/app/tasks/collections.py
```diff
@@ -81,6 +82,32 @@ def sync(remote_pk, repository_pk, mirror):
     d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
     d_version.create()
 
+    repo_version = repository.latest_version()
+    to_deprecate = []
+    to_deprecate_query = Q()
+    to_undeprecate_query = Q()
+    for collection, deprecated in first_stage.deprecations.items():
```
bmbouter, line 89: I see one way this could be even higher performing without much effort. Try having deprecations contain the pre-built `Q()` objects, and only store those where `deprecated=True`. This improves several things:

1. It doesn't have to re-loop through every collection here, just those that are deprecated. The number of `deprecated = True` will likely be few.
2. It doesn't have to join them with a '.' only to resplit them again later

pulp_ansible/app/tasks/collections.py
```diff
@@ -81,6 +82,32 @@ def sync(remote_pk, repository_pk, mirror):
...
+        collections_deprecate_true_qs = Collection.objects.filter(to_deprecate_query)
+        for collection in collections_deprecate_true_qs:
+            to_deprecate.append(
+                AnsibleCollectionDeprecated(repository_version=repo_version, collection=collection)
+            )
+
+        AnsibleCollectionDeprecated.objects.bulk_create(to_deprecate, ignore_conflicts=True)
+
+    if to_undeprecate_query:
+        collections_deprecate_false_qs = Collection.objects.filter(to_undeprecate_query)
```
bmbouter, line 106: Instead of constructing `to_undeprecate_query` using `Q()` objects, use the `Collection.objects.exclude(to_deprecate_query).

pulp_ansible/app/tasks/collections.py
bmbouter: I suspect you're onto something, but can you help me understand the case a bit more? I don't follow.

pulp_ansible/app/tasks/collections.py
```diff
@@ -81,6 +82,27 @@ def sync(remote_pk, repository_pk, mirror):
     d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
     d_version.create()
 
+    repo_version = repository.latest_version()
+    to_deprecate = []
+    to_deprecate_query = Q()
+    for collection, deprecated in first_stage.deprecations.items():
+        namespace, name = collection.split(".")
+        to_deprecate_query |= Q(namespace=namespace, name=name)
```
bmbouter, line 90: This aggregation can happen at L#482 instead `self.deprecations[f"{namespace}.{name}"] = result["deprecated"]`. It can just build and `|=` the Q objects there.

pulp_ansible/app/tasks/collections.py
```diff
@@ -81,6 +82,27 @@ def sync(remote_pk, repository_pk, mirror):
...
+        for collection in collections_deprecate_true_qs:
+            to_deprecate.append(
+                AnsibleCollectionDeprecated(repository_version=repo_version, collection=collection)
+            )
+
+        AnsibleCollectionDeprecated.objects.bulk_create(to_deprecate, ignore_conflicts=True)
+
+    AnsibleCollectionDeprecated.objects.filter(
+        repository_version=repo_version,
+        collection__pk__in=Collection.objects.exclude(to_deprecate_query),
```
bmbouter, line 103: for readability I would move `Collection.objects.exclude(to_deprecate_query)` to a line earlier and be:

`non_deprecated_qs = Collection.objects.exclude(to_deprecate_query)`.

bmbouter (approved):
I think this is hella great. I'll refactor the tests to use the common machinery in my branch to shorten them some when I rebase. Thank you!!!

PR #434, opened 2020-11-20
Increase Collection Tag length to 64 chars

bmbouter:
Increases collection tag field length from 32 to 64, which allows
sync to work for longer tag names used on galaxy.ansible.com.

closes #7827


CHANGES/7827.bugfix
```diff
@@ -0,0 +1,2 @@
+Increases collection tag field length from 32 to 64, which allows sync to work for longer tag names
```
daviddavis, line 1: nitpick - should use past tense.

daviddavis (approved):

PR #440, opened 2020-11-23
Enable AH test

fao89:
[noissue]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,10 @@ jobs:
           .github/workflows/scripts/before_script.sh
           echo "HTTPIE_CONFIG_DIR=$GITHUB_WORKSPACE/.github" >> $GITHUB_ENV
 
+      - name: Setting env vars
+        if: github.event_name != 'pull_request'
+        run: echo 'AUTOMATION_HUB_TOKEN_AUTH='${{ secrets.AUTOMATION_HUB_TOKEN_AUTH }} >> $GITHUB_ENV
+
       - name: Script
         run: .github/workflows/scripts/script.sh
         shell: bash


.github/workflows/ci.yml
```diff
@@ -105,6 +105,8 @@ jobs:
           echo "HTTPIE_CONFIG_DIR=$GITHUB_WORKSPACE/.github" >> $GITHUB_ENV
 
       - name: Script
+        env:
+          AUTOMATION_HUB_TOKEN_AUTH: ${{ secrets.AUTOMATION_HUB_TOKEN_AUTH }}
```
fao89, line 109: @daviddavis @dralley @mdellweg we need to find a better way to set env vars, otherwise plugin-template would remove it

mdellweg: This is a gha variable, so we are stuck with this file, right?
We could create something similar to how we feed settings from the template_config.

mdellweg: Alternatively, we collect all such information in a vault and unlock it with the plugin_template.
OTOH, you should not put any secrets in the workflows that run on pull_requests. They can be leaked.

daviddavis: > OTOH, you should not put any secrets in the workflows that run on pull_requests. They can be leaked.

I've tested and confirmed that secrets are not shared with certain builds such as those triggered by PRs. It's why the PR processor is a cron job.

daviddavis: Another option would be to define a new workflow to set this variable and run the `AutomationHubV3SyncCase` test.

mdellweg: > > OTOH, you should not put any secrets in the workflows that run on pull_requests. They can be leaked.
> 
> I've tested and confirmed that secrets are not shared with certain builds such as those triggered by PRs. It's why the PR processor is a cron job.

That means the test needing that token will never succeed on a pr.

daviddavis: > That means the test needing that token will never succeed on a pr.

Yes. If you look at the logs, it's failing because it's empty.

daviddavis: > Alternatively, we collect all such information in a vault and unlock it with the plugin_template.

I guess we could encrypt a script that exports a bunch of variables and then store the encryption key, and the CI would unecrypt and source this script?

mdellweg: That would basically be a vault.
But still the problem with this PR is that we cannot use any credential whatsoever in PR runs.

fao89: this workflow is at the same time for PRs and pushes to master, we need to conditionally set this variable only for master

mdellweg: So the test will be skipped if the token is absent?

fao89: GHA basically advises to do:
```
echo MYVAR=VAR_VALUE >> $GITHUB_ENV
```
I believe $GITHUB_ENV is kind of a .env file
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

fao89:
Closing in favor of https://github.com/pulp/plugin_template/pull/311

PR #445, opened 2020-12-01
Release 0.6.0

fao89:
Redmine query of issues to close:
https://pulp.plan.io/issues?set_filter=1&status_id=*&issue_id=7873,7741,7777,7739,7827
Release commit == eb44302

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ async_lru
 galaxy_importer
 jsonschema>=3.0
 packaging
-pulpcore>=3.7
+pulpcore>=3.7,<3.10
 PyYAML
 semantic_version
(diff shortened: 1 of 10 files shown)


requirements.txt
```diff
@@ -2,6 +2,6 @@ async_lru
 galaxy_importer
 jsonschema>=3.0
 packaging
-pulpcore>=3.7
+pulpcore>=3.7,<3.10
```
fao89: is this okay? Or should we support 3.10?

bmbouter: I think this is right. Plugins should plan to support the pulpcore current version + 1, so the current version is 3.8 so yes.

bmbouter (approved):
Looks great, thanks!

PR #449, opened 2020-12-03
Slow down sync when token is required

fao89:
https://pulp.plan.io/issues/7917
closes #7917

### DIFF
--- a/CHANGES/7917.bugfix
+++ b/CHANGES/7917.bugfix
@@ -1,0 +1,1 @@
+Avoid rate limiting by slowing down sync when token is required
--- a/pulp_ansible/app/downloaders.py
+++ b/pulp_ansible/app/downloaders.py
@@ -128,6 +128,7 @@ class TokenAuthHttpDownloader(HttpDownloader):
             DownloadResult: Contains information about the result. See the DownloadResult docs for
                  more information.
         """
+        await asyncio.sleep(0.2)
         async with self.session.get(self.url, headers=headers, proxy=self.proxy) as response:
             self.raise_for_status(response)
             to_return = await self._handle_response(response)


CHANGES/7917.bugfix
```diff
@@ -0,0 +1 @@
+Slow down sync when token is required
```
bmbouter, line 1: Can we reword this to focus on what it's fixing instead? Perhaps `Avoid rate limiting by slowing down sync when token is required`.

bmbouter:
Also will we be putting this on master too? If so, it probably should go there first and then be brought back as a backport.

fao89:
> Also will we be putting this on master too? If so, it probably should go there first and then be brought back as a backport.

I was thinking: hardcoded on 0.5 and using a setting on master

bmbouter:
If we're going to make a "slow down" feature we should implement leaky bucket and have the requests per second rate be on the Remote. It would be much more usable for example if we know the remote site has 12 req / sec then someone could set it to 11 and know it's the right value.

fao89:
I agree, I'll change this PR to the master branch then

pulp_ansible/app/downloaders.py
```diff
@@ -128,6 +128,7 @@ async def _run_with_additional_headers(self, headers):
             DownloadResult: Contains information about the result. See the DownloadResult docs for
                  more information.
         """
+        await asyncio.sleep(0.2)
```
fao89, line 131: @newswangerd do you think 0.2 is good?
ref: https://hackmd.io/@fao89/rate_limiting

chouseknecht (approved):

bmbouter (approved):
Thank you!

pulp_ansible/app/downloaders.py
newswangerd: Sorry I'm late to the party. I would have suggested something more like 0.5. The rationale being that we know the limit is 15 requests per second and concurrent connections defaults to 10, so that would give us a theoretical upper bound of 20 requests per second, assuming all requests are instant (which they aren't). That should roughly guarantee that nobody runs into this issue.

0.2 is probably fine though.

PR #455, opened 2020-12-04
Increase interval between requests when token is required

fao89:
https://pulp.plan.io/issues/7929
closes #7929


chouseknecht (approved):

chouseknecht:
I really don’t that we’re having to do this, but it seems like the only way to get galaxy_ng 4.2.1 release out. Hopefully we can simply the sync process soon, and remove the need for this.

PR #458, opened 2020-12-04
Make pulp_ansible downloader reuse connections

alikins:
NOTE: This is completely untested and is provided as a potential fix/workaround.
It isn't DRY, etc. But mostly wanted eyes to see if this approach would make sense.

This cuts and pasts from pulpcore.plugin.download.DownloadFactory.

It overrides the _make_aiohttp_session_from_remote() method
used to create the ClientSession for http* Downloaders.

The method is changed slightly to disable 'force_close'
and to set a 'limit_per_host' of 8 for the TCPConnection

https://pulp.plan.io/issues/7929
closes #7929

### DIFF
--- a/pulp_ansible/app/downloaders.py
+++ b/pulp_ansible/app/downloaders.py
@@ -1,14 +1,23 @@
 import asyncio
 import backoff
 import json
+import ssl
+from tempfile import NamedTemporaryFile
 
 from gettext import gettext as _
 from logging import getLogger
 
+import aiohttp
 from aiohttp import BasicAuth
 from aiohttp.client_exceptions import ClientResponseError
 
-from pulpcore.plugin.download import http_giveup, DownloaderFactory, FileDownloader, HttpDownloader
+from pulpcore.download.factory import user_agent
+from pulpcore.plugin.download import (
+    http_giveup,
+    DownloaderFactory,
+    FileDownloader,
+    HttpDownloader,
+)
 
 
 log = getLogger(__name__)
@@ -35,6 +44,7 @@ class AnsibleFileDownloader(FileDownloader):
 
 AUTH_TOKEN = None
 TOKEN_LOCK = asyncio.Lock()
+AUTOHUB_LIMIT_PER_HOST = 2
 
 
 class TokenAuthHttpDownloader(HttpDownloader):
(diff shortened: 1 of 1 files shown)


alikins:
> discussion from slack
> 
> pulpcore.download.factory.DownloaderFactory() doc string claims `Also for http and https urls, even though HTTP 1.1 is used, the TCP connection is setup and
>     closed with each request. This is done for compatibility reasons due to various issues related
>     to session continuation implementation in various servers.`
> 
> That feels like it could contribute to triggering rate limiting. Don't know if that gets overridden elsewhere, but if not, maybe turning it on could help.  I'm assuming akamai supports http 1.1 and keep-alive or whatever
> 
> esp since akamai would be doing the tls termination for those connections
> Has anyone reproduced hitting the rate limit with any other http client tools?
> 
> https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L57-L59
> 
> https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L88-L92
> 
> In theory, the HttpDownloader may be overriding the one-connect-per-request session mentioned above https://github.com/pulp/pulpcore/blob/master/pulpcore/download/http.py#L41-L44
> 
> In theory, aiohttp supports a connection limit per host 'limit_per_host' https://docs.aiohttp.org/en/stable/client_advanced.html#limiting-connection-pool-size
> 
> Some quick grepping suggests that isn't currently used
> 
> Think you could set it in https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L122   though doing it without pulpcore changes may mean providing an alternative ClientSession object

alikins:
ping @fao89 @bmbouter @ironfroggy

pulp_ansible/app/downloaders.py
```diff
@@ -35,6 +44,7 @@ def __init__(self, *args, **kwargs):
 
 AUTH_TOKEN = None
 TOKEN_LOCK = asyncio.Lock()
+AUTOHUB_LIMIT_PER_HOST = 8
```
fao89, line 47: I was rate limited with 8, then I changed to 2 and I could sync.
We need more tests to find the best number

bmbouter:
> > discussion from slack
> > pulpcore.download.factory.DownloaderFactory() doc string claims `Also for http and https urls, even though HTTP 1.1 is used, the TCP connection is setup and closed with each request. This is done for compatibility reasons due to various issues related to session continuation implementation in various servers.`
> > That feels like it could contribute to triggering rate limiting. Don't know if that gets overridden elsewhere, but if not, maybe turning it on could help.  I'm assuming akamai supports http 1.1 and keep-alive or whatever
> > esp since akamai would be doing the tls termination for those connections
> > Has anyone reproduced hitting the rate limit with any other http client tools?
> > https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L57-L59
> > https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L88-L92
> > In theory, the HttpDownloader may be overriding the one-connect-per-request session mentioned above https://github.com/pulp/pulpcore/blob/master/pulpcore/download/http.py#L41-L44
> > In theory, aiohttp supports a connection limit per host 'limit_per_host' https://docs.aiohttp.org/en/stable/client_advanced.html#limiting-connection-pool-size
> > Some quick grepping suggests that isn't currently used
> > Think you could set it in https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L122   though doing it without pulpcore changes may mean providing an alternative ClientSession object

How can we establish the relationship between the force close and rate limiting? The `Also for http and https urls, even though HTTP 1.1 is used, the TCP connection is setup and closed with each request. This is done for compatibility reasons due to various issues related to session continuation implementation in various servers.` comment came out of a roughly 2-week investigation that made sync's unstable in a variety of plugins. If the servers pulp_ansible uses don't have that problem then let's change it, but we need to be sure we're actually improving the rate limiting problem. What testing can be done to be sure?

bmbouter:
Since this PR is not in a state for merging please change it to draft also.

pulp_ansible/app/downloaders.py
```diff
@@ -1,14 +1,23 @@
...
 from logging import getLogger
 
+import aiohttp
 from aiohttp import BasicAuth
 from aiohttp.client_exceptions import ClientResponseError
 
-from pulpcore.plugin.download import http_giveup, DownloaderFactory, FileDownloader, HttpDownloader
+from pulpcore.plugin.download import (
+    http_giveup,
+    user_agent,
```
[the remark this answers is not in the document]
alikins: Won't a import from pulpcore trigger the CI test that checks for pulpcore imports?

pulp_ansible/app/downloaders.py
alikins: Yup, 8 was just a guess. Needs tuning.

pulp_ansible/app/downloaders.py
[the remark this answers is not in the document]
alikins: Ah, never mind. Fixing.

alikins:
Updated the import statement. Still more or less completely untested ;->

pulp_ansible/app/downloaders.py
alikins: Updated to default to 2.

fao89:
closing in favor of https://github.com/pulp/pulp_ansible/pull/496

PR #464, opened 2020-12-09
Allow updating auth_url on CollectionRemote when token is already set

fao89:
https://pulp.plan.io/issues/7957
closes #7957


daviddavis:
Would it be easy to add a test?

fao89:
> Would it be easy to add a test?

Yes it is, let me work on it

daviddavis (approved):

PR #466, opened 2020-12-09
Expose the ansible supported version for each collection version

fao89:
https://pulp.plan.io/issues/7949
closes #7949


fao89:
http://pulp3-source-fedora31.localhost.example.com/pulp_ansible/galaxy/c09fb051-3eb3-412e-ad27-6a4ef6e872be/api/v3/collections/
image


http://pulp3-source-fedora31.localhost.example.com/pulp_ansible/galaxy/c09fb051-3eb3-412e-ad27-6a4ef6e872be/api/v3/collections/cisco/nxos/versions/1.3.2-dev2/
image

pulp_ansible/app/migrations/0028_collectionversion_requires_ansible.py
```diff
@@ -0,0 +1,39 @@
...
+        with artifact.file.open() as artifact_file, tarfile.open(fileobj=artifact_file, mode="r") as tar:
+            try:
+                file_obj = tar.extractfile("meta/runtime.yml")
+            except KeyError:
+                continue
+
+            meta_runtime = yaml.safe_load(file_obj)
+            if meta_runtime.get("requires_ansible"):
+                collection_version.requires_ansible = meta_runtime.get("requires_ansible")
+                collection_version.save()
```
fao89, line 21: we can merge this with the data migration from https://github.com/pulp/pulp_ansible/pull/468

pulp_ansible/app/tasks/collections.py
```diff
@@ -146,10 +169,16 @@ def import_collection(
...
+        with temp_file.file.open() as artifact_file, tarfile.open(
+            fileobj=artifact_file, mode="r"
+        ) as tar:
             importer_result = process_collection(
                 artifact_file, filename=filename, logger=user_facing_logger
             )
+            runtime = _get_file_obj(tar, "meta/runtime.yml", temp_file.file.name, False)
+            if runtime:
+                meta_runtime = yaml.safe_load(runtime)
+                importer_result["requires_ansible"] = meta_runtime.get("requires_ansible")
```
fao89, line 181: @awcrosby does the importer handle `requires_ansible`?

awcrosby: @fao89 currently it does not, I made issue https://issues.redhat.com/browse/AAH-231 to track this. We may end up making it a top-level key of the importer result as you have here.

awcrosby: @fao89 to clarify, we will make the `galaxy-importer` provide this as a top-level key as you have in this PR, so this looks good

fao89: thank you @awcrosby

pulp_ansible/app/tasks/collections.py
```diff
@@ -203,6 +226,7 @@ def create_collection_from_importer(importer_result):
         collection_version = CollectionVersion(
             collection=collection,
             **collection_info,
+            requires_ansible=importer_result.get("requires_ansible"),
```
fao89, line 229: `None` for now, but it will get values once `galaxy-importer` is released

bmbouter: So `galaxy-importer` will carry the implementation and pulp_ansible will return `None` until `galaxy-importer` releases it? I'm ok with that, but I wanted to confirm my understanding.

awcrosby: yes that is correct - we have merged in the galaxy-importer implementation here: https://github.com/ansible/galaxy-importer/pull/110

pulp_ansible/app/tasks/collections.py
```diff
@@ -61,6 +62,28 @@
 log = logging.getLogger(__name__)
 
 
+def _get_file_obj(tar, file_path, artifact_path, raise_exc=True):
```
bmbouter, line 65: Can this docstring have its signature variables added to it please.

pulp_ansible/app/tasks/collections.py
```diff
@@ -611,17 +635,13 @@ async def _post_save(self, batch):
...
-                        try:
-                            file_obj = tar.extractfile(manifest)
-                        except KeyError:
-                            file_obj = None
-                        else:
-                            break
-                    if not file_obj:
-                        raise FileNotFoundError("MANIFEST.json not found")
-                    manifest_data = json.load(file_obj)
+                    runtime = _get_file_obj(tar, "meta/runtime.yml", artifact.file.name, False)
```
bmbouter, line 638: Please pass the kwarg argument as a kwarg. (the `False`).

bmbouter, line 638:
```suggestion
                    runtime_metadata = _get_file_obj(tar, "meta/runtime.yml", artifact.file.name, raise_exc=False)
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -611,17 +635,13 @@ async def _post_save(self, batch):
...
+                    if runtime:
+                        meta_runtime = yaml.safe_load(runtime)
```
bmbouter, line 640:
```suggestion
                        runtime_yaml = yaml.safe_load(runtime)
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -45,7 +45,9 @@
 from pulp_ansible.app.viewsets import CollectionFilter, CollectionVersionFilter
 
 
-CollectionTuple = namedtuple("CollectionTuple", ["namespace", "name", "version", "pulp_created"])
+CollectionTuple = namedtuple(
+    "CollectionTuple", ["namespace", "name", "version", "pulp_created", "requires_ansible"]
```
bmbouter, line 49: What motivated this change?

CHANGES/7949.feature
```diff
@@ -0,0 +1 @@
+Expose the ansible supported version for each collection version
```
bmbouter, line 1:
```suggestion
Adds the ``requires_ansible`` attribute to the Galaxy V3 CollectionVersion APIs. This documents the version of Ansible required to use the collection.
```

bmbouter:
I don't think we have any tests that assert on the Galaxy APIs directly do we? I think we only have tests that assert the CLI can correctly operate with those APIs. If it's true that there are no galaxy API tests currently, let's not add them just for this one change, but we should try to add comprehensive, direct tests for the Galaxy APIs in the future.

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -70,7 +70,11 @@ def get_highest_version(self, obj):
                 "version": collection.version,
             },
         )
-        return {"href": href, "version": str(collection.version)}
+        return {
+            "href": href,
+            "version": str(collection.version),
+            "requires_ansible": collection.requires_ansible,
```
fao89, line 76: @bmbouter I added `requires_ansible` on the NamedTuple to get it for the highest version and display it at collection level:
image

pulp_ansible/app/galaxy/v3/views.py
fao89: https://github.com/pulp/pulp_ansible/pull/466/files#r567994874

fao89:
> I don't think we have any tests that assert on the Galaxy APIs directly do we? I think we only have tests that assert the CLI can correctly operate with those APIs. If it's true that there are no galaxy API tests currently, let's not add them just for this one change, but we should try to add comprehensive, direct tests for the Galaxy APIs in the future.

we have this:
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/v3/test_collection.py

bmbouter:
> > I don't think we have any tests that assert on the Galaxy APIs directly do we? I think we only have tests that assert the CLI can correctly operate with those APIs. If it's true that there are no galaxy API tests currently, let's not add them just for this one change, but we should try to add comprehensive, direct tests for the Galaxy APIs in the future.
> 
> we have this:
> https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/v3/test_collection.py

Then maybe we should build out some tests for this there? Wdyt?

pulp_ansible/app/galaxy/v3/serializers.py
bmbouter: I only expected the `requires_ansible` to be on the serializer for `collections/<str:namespace>/<str:name>/versions/<str:version>/`. Maybe I didn't understand the requirements though. What motivated delivering this data there?

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
```diff
@@ -221,6 +221,10 @@ def test_collection_version(artifact, pulp_client, collection_detail):
     assert "updated_at" in version
     assert "created_at" in version
 
+    assert "files" in version
```
bmbouter, line 224: Is there some assertion that can be made for the data being returned? If we have a test that sync's a specific version of a collection then we could assert on the data populated during import.

Actually we could also do one to upload a collection and then assert there is on data currently and then when galaxy-importer releases it'll fail and we can update the test to match then.

Wdyt?

fao89: actually `requires_ansible` is optional, even with galaxy-importer release it probably won't fail
(requires_ansible is rarely used)

fao89: the importer is already released: https://github.com/ansible/galaxy-importer/blob/master/CHANGES.rst#galaxy-importer-0214-2021-01-28

fao89:
```
pulp_ansible/tests/functional/api/collection/v3/test_collection.py ...
>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/vagrant/devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/v3/test_collection.py(224)test_collection_version()
-> assert "files" in version
(Pdb) version["files"]
{}
(Pdb) version["manifest"]
{}
(Pdb) version["requires_ansible"]
(Pdb) version
{'version': '1.0.0', 'href': '/pulp_ansible/galaxy/automation-hub/api/v3/collections/orionuser1/collection_dep_a_arwtsrzb/versions/1.0.0/', 'created_at': '2021-02-02T16:15:16.450378Z', 'updated_at': '2021-02-02T16:15:16.450397Z', 'manifest': {}, 'files': {}, 'requires_ansible': None, 'artifact': {'filename': 'orionuser1-collection_dep_a_arwtsrzb-1.0.0.tar.gz', 'sha256': '35ce4f9e17a9d973939e044b7eedd4355f93ee827dd36104bbb6890afed44eb7', 'size': 1520}, 'collection': {'id': 'cda47603-fcbb-4b84-87e2-d9631979a33d', 'name': 'collection_dep_a_arwtsrzb', 'href': '/pulp_ansible/galaxy/automation-hub/api/v3/collections/orionuser1/collection_dep_a_arwtsrzb/'}, 'download_url': 'http://pulp3-source-fedora33.localhost.example.com/pulp/content/automation-hub/orionuser1-collection_dep_a_arwtsrzb-1.0.0.tar.gz', 'name': 'collection_dep_a_arwtsrzb', 'namespace': {'name': 'orionuser1'}, 'metadata': {'authors': ['Orion User 1'], 'contents': [], 'dependencies': {}, 'description': 'a collection with some deps on other collections', 'documentation': '', 'homepage': '', 'issues': '', 'license': ['GPL-3.0-or-later'], 'repository': 'http://github.example.com/orionuser1/skeleton', 'tags': ['collectiontest']}}
```

bmbouter: Yes so what I'm saying is that in this response we're asserting on the key and not the value. The test data has `'files': {}` so asserting on `{}` isn't worth much. What would be better is to have the test data populate the `files` `manifest` and `requires_ansible` responses with actual data and then the test code assert on the values (not just the presence of the keys).

bmbouter: Oh I see a test further down that seems like it's doing just that. Let me check that out.

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
```diff
@@ -32,61 +30,62 @@ def setUpClass(cls):
...
-        # Create a distribution.
-        body = gen_distribution()
-        body["repository"] = repo.pulp_href
-        distribution_create = self.distributions_api.create(body)
-        created_resources = monitor_task(distribution_create.task).created_resources
-        distribution = self.distributions_api.read(created_resources[0])
-        self.base_path = distribution.base_path
-        self.addCleanup(self.distributions_api.delete, distribution.pulp_href)
-
     @unittest.skip("FIXME: Re-enable later")
```
bmbouter, line 33: This test is being skipped do we still want this skipped?

bmbouter: Oh is this what is enabled in the other PR actually?

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
```diff
@@ -32,61 +30,62 @@ def setUpClass(cls):
...
 
         self.assertEqual(highest_version, original_highest_version)
         self.assertEqual(original_total_versions, total_versions + 1)
         self.assertGreater(updated_at, original_updated_at)
+
+    def test_v3_collection_version(self):
+        """Test Collection Versions V3 endpoint fields."""
+        body = gen_ansible_remote(
+            url="https://galaxy.ansible.com",
+            requirements_file="collections:\n  - name: cisco.nxos\n    version: ==1.4.0",
```
bmbouter, line 76: :+1:

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
```diff
@@ -32,61 +30,62 @@ def setUpClass(cls):
...
-            "pulp_installer", "pulp", self.base_path
-        ).meta["count"]
+            "pulp_installer", "pulp", distribution.base_path
+        ).meta.count
 
         self.assertEqual(highest_version, original_highest_version)
         self.assertEqual(original_total_versions, total_versions + 1)
         self.assertGreater(updated_at, original_updated_at)
+
+    def test_v3_collection_version(self):
```
bmbouter, line 72: This test is great. It's really only specific to data populated by sync so maybe:

```suggestion
    def test_v3_collection_version_from_synced_data(self):
```

bmbouter: What would be really great is if we could also use the CLI to publish a collection and then make similar assertions. That codepath flows through the importer so only a test like that can assert on our integration code with the importer. Wdyt?

pulp_ansible/app/migrations/0030_collectionversion_requires_ansible.py
```diff
@@ -0,0 +1,43 @@
...
+
+import json
+import tarfile
+import yaml
+from django.db import migrations, models
+
+from pulp_ansible.app.tasks.utils import get_file_obj_from_tarball
+
+
+def set_requires_ansible_and_manifest_and_files_json(apps, schema_editor):
```
bmbouter, line 11: This is pretty slick consolidating the migrations like this for performance reasons. :+1:

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
fao89: yes, this is re-enabled on `update_at/uqery performance` PR

pulp_ansible/app/migrations/0030_collectionversion_requires_ansible.py
```diff
@@ -0,0 +1,43 @@
...
+                tar, "meta/runtime.yml", artifact.file.name, raise_exc=False
+            )
+            if runtime_metadata:
+                runtime_yaml = yaml.safe_load(runtime_metadata)
+                collection_version.requires_ansible = runtime_yaml.get("requires_ansible")
+            collection_version.manifest = json.load(
+                get_file_obj_from_tarball(tar, "MANIFEST.json", artifact.file.name)
+            )
+            collection_version.files = json.load(get_file_obj_from_tarball(tar, "FILES.json", artifact.file.name))
+            collection_version.save()
```
bmbouter, line 26: One idea I had was to wrap this whole section in a general try/except blog that logs and continues? I'm not sure if that's a great idea, but it would be better to continue if a fatal exception was experienced due to one or a small number of data quality issues (e.b. a missing FILES.json, invalid yaml, missing MANIFEST.json). It would make this migration more "defensive". Wdyt about this?

fao89: I think we can address it by passing `raise_exc=False`

bmbouter: We can but what about the invalid yaml case?

bmbouter: Oh and also I think we want this to be logged to the user, which that wouldn't do.

pulp_ansible/tests/functional/api/collection/v3/test_serializers.py
bmbouter: Actually maybe the existing CLI upload test could just use a collection that has all this data and then make assertions in that existing test.

pulp_ansible/tests/functional/cli/test_collection_upload.py
```diff
@@ -69,3 +72,57 @@ def test_upload_collection(self):
...
+
+        repo = self.repo_api.read(repo.pulp_href)
+        repo_version = self.repo_versions_api.read(repo.latest_version_href)
+        self.assertEqual(repo_version.number, 1)  # We uploaded 1 collection
+
+        version = self.collections_versions_v3api.read(
+            collection_name, "pulp", distribution.base_path, "1.0.0"
+        )
+
+        self.assertEqual(version.requires_ansible, ">=2.9.10,<2.11")
```
bmbouter, line 128: This is great!

pulp_ansible/tests/functional/cli/test_collection_upload.py
```diff
@@ -69,3 +72,57 @@ def test_upload_collection(self):
...
+            runtime_path = "{}{}/meta".format(temp_dir, "/pulp/" + collection_name)
+            cmd = "mkdir -p {}".format(runtime_path)
+            subprocess.run(cmd.split())
+            with open(f"{runtime_path}/runtime.yml", "wb") as runtime_yml:
+                runtime_yml.write(http_get(runtime_url))
+
+            cmd = "ansible-galaxy collection build --output-path {} {}{}".format(
+                temp_dir, temp_dir, "/pulp/" + collection_name + "/"
+            )
+            subprocess.run(cmd.split())
```
bmbouter, line 109: Is building this once and checking the tarball as an asset in the tests area an option? If that tarball was checked in all of this code could be removed which would simplify things overall.

pulp_ansible/tests/functional/cli/test_collection_upload.py
```diff
@@ -69,3 +73,37 @@ def test_upload_collection(self):
...
+        # Create a distribution.
+        body = gen_distribution()
+        body["repository"] = repo.pulp_href
+        distribution_create = self.distributions_api.create(body)
+        created_resources = monitor_task(distribution_create.task).created_resources
+        distribution = self.distributions_api.read(created_resources[0])
+
+        self.addCleanup(self.distributions_api.delete, distribution.pulp_href)
+        colletion_path = os.path.join(
+            os.getcwd(), "pulp_ansible/tests/assets/collections/pulp-testing_asset-1.0.0.tar.gz"
```
bmbouter, line 92: Perfect!

pulp_ansible/tests/functional/constants.py
```diff
@@ -86,3 +86,6 @@
 - rshad.collection_demo
 - brightcomputing.bcm
 """
+
+RAW_COLLECTIONS = "https://raw.githubusercontent.com/ansible-collections/"
+RUNTIME_URL = RAW_COLLECTIONS + "cisco.nxos/1.4.0/meta/runtime.yml"
```
bmbouter, line 91: I think these can be removed.

bmbouter (approved):
This is excellent! Thank you! :+1: :rocket:

PR #468, opened 2020-12-10
Expose MANIFEST.json and FILES.json at CollectionVersion endpoint

fao89:
https://pulp.plan.io/issues/7572
closes #7572


fao89:
http://pulp3-source-fedora31.localhost.example.com/pulp_ansible/galaxy/e991b94f-a732-42e5-96d2-d685ed80bd18/api/v3/collections/testing/k8s_demo_collection/versions/0.0.3/

image

pulp_ansible/app/migrations/0028_manifest_and_files_json_fields.py
```diff
@@ -0,0 +1,52 @@
...
+def set_manifest_and_files_json(apps, schema_editor):
+    CollectionVersion = apps.get_model("ansible", "CollectionVersion")
+    for collection_version in CollectionVersion.objects.all():
+        artifact = collection_version.contentartifact_set.get().artifact
+        with artifact.file.open() as artifact_file, tarfile.open(fileobj=artifact_file, mode="r") as tar:
+            collection_version.manifest = json.load(
+                _get_file_obj(tar, "MANIFEST.json", artifact.file.name)
+            )
+            collection_version.files = json.load(_get_file_obj(tar, "FILES.json", artifact.file.name))
+            collection_version.save()
```
fao89, line 30: we can merge this with the data migration from https://github.com/pulp/pulp_ansible/pull/466

pulp_ansible/app/tasks/collections.py
```diff
@@ -596,6 +596,28 @@ async def _post_save(self, batch):
                 :class:`~pulpcore.plugin.stages.DeclarativeContent` objects to be saved.
 
         """
+
+        def _get_file_obj(tar, file_path, artifact_path):
```
newswangerd, line 600: @awcrosby I'd like you to take a look at this. It seems like opening the tarball and extracting the manifest should be something that happens during the import instead of as a separate function that pulp performs when a collection is added.

The one snag here is that this would also require the sync logic to be updated to pull down the manifest and files list.

What do you think?

awcrosby: when `pulp-ansible` calls `galaxy-importer` the `MANIFEST.json` is extracted and validated, as long as this step occurs after that validation then it should be ok

it looks like pulp-ansible was already untaring in `_post_save()` and this PR makes it so data gets saved to the `CollectionVersion`

CHANGES/7572.feature
```diff
@@ -0,0 +1 @@
+Expose MANIFEST.json and DIGEST.json at CollectionVersion endpoint
```
awcrosby, line 1: This and title should be updated from DIGEST.json to FILES.json, I'm not aware of any DIGEST.json

pulp_ansible/app/tasks/collections.py
fao89: yes, this is what is blocking us from doing on-demand sync, we always open the tarball during sync,
the importer is only used when uploading collections.

pulp_ansible/app/migrations/0029_manifest_and_files_json_fields.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 2.2.17 on 2020-12-10 15:04
+import json
+import tarfile
+import django.contrib.postgres.fields.jsonb
+from django.db import migrations
+
+
+def _get_file_obj(tar, file_path, artifact_path):
```
bmbouter, line 8: With this being present in a few places I think the real implementation should be moved to a utility area.

bmbouter (changes requested):
This PR is lgtm, but what should be done to de-duplicate this utility funciton? If you move it to a utils area, docstring it, and add the raise_exc kwarg I would lgtm this PR.

bmbouter (approved):
This looks good to me. Thanks!

PR #477, opened 2020-12-22
Optimize is_highest

fao89:
[noissue]

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -1,8 +1,8 @@
-from collections import defaultdict, namedtuple
 from datetime import datetime
 from gettext import gettext as _
 import semantic_version
 
+from django.contrib.postgres.aggregates import ArrayAgg
 from django.db.models import Exists, OuterRef
 from django.shortcuts import get_object_or_404
 from django.utils.datastructures import MultiValueDictKeyError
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -155,32 +156,10 @@ def get_queryset(self):
...
-        collections = collections.annotate(deprecated=Exists(deprecated_query))
-
-        versions_qs = CollectionVersion.objects.filter(pk__in=repo_version.content).values_list(
-            "collection_id",
-            "namespace",
-            "name",
-            "version",
-            "pulp_created",
+        collections = collections.annotate(
+            deprecated=Exists(deprecated_query), available_versions=ArrayAgg("versions__version")
```
fao89, line 160: @cutwater I think we can solve the semver comparison with this, we bring all available versions and sort the values at the serializer
```python
version = sorted(
            obj.available_versions, key=lambda ver: semantic_version.Version(ver), reverse=True
        )[0]
```

fao89:
Only PoC for Array Agg:
```sql
SELECT DISTINCT ...,
              ARRAY_AGG("ansible_collectionversion"."version" ) AS "available_versions"
  FROM "ansible_collection"
 INNER JOIN "ansible_collectionversion"
    ON ("ansible_collection"."pulp_id" = "ansible_collectionversion"."collection_id")
...
```

fao89:
Just a PoC

PR #478, opened 2020-12-23
Openapi test

fao89:


pulp_ansible/tests/functional/api/test_schema.py
```diff
@@ -0,0 +1,24 @@
...
+cfg = config.get_config()
+base_url = cfg.get_base_url()
+schema = schemathesis.from_uri(f"{base_url}/pulp/api/v3/docs/api.json?plugin=pulp_ansible")
+
+
+@schema.parametrize()
+def test_time(case):
+    """Test response time."""
+    response = case.call()
+    assert response.elapsed.total_seconds() < 1.1
```
Stranger6667, line 14: Hi @fao89! Just a quick note on this. The default timeout for a single test is 500 ms, but as I see it, there could be longer responses that are still ok if less than 1100 ms. I suggest adding the `hypothesis.settings` decorator with a customized `deadline` setting to avoid having `DeadlineExceeded` exceptions. Something like this:

```python
from hypothesis import settings

...
@schema.parametrize()
@settings(deadline=1500)
def test_time(case):
    ...
```

Also, you might be interested in Schemathesis trying to cause a longer response time by using `hypothesis.target`:

```python
from hypothesis import target

...
def test_time(case):
    response = case.call()
    target(response.elapsed.total_seconds())
    ...
```

Note, the above examples imply using the latest `hypothesis` version.

fao89: thank you @Stranger6667

Stranger6667: You are welcome! :) Just a quick note - there still should be an assertion after the `target` call so the test will fail in case of long responses :)

PR #479, opened 2021-01-04
OpenAPI validation

fao89:


fao89:
@daviddavis could you please review this PR and merge https://github.com/pulp/pulpcore/pull/1065 ?

daviddavis (dismissed):

daviddavis:
Was there a bug fixed here and if so would it be possible to get a changelog entry?

fao89:
> Was there a bug fixed here and if so would it be possible to get a changelog entry?

I can do it

CHANGES/8037.bugfix
```diff
@@ -0,0 +1 @@
+Fix OpenAPI schema pagination
```
daviddavis, line 1: Any chance you could provide a little more context? I'm not sure I understand the problem that was fixed.

pulp_ansible/app/galaxy/v3/pagination.py
```diff
@@ -73,27 +73,33 @@ def get_paginated_response_schema(self, schema):
             "type": "object",
             "properties": {
                 "meta": {
-                    "count": {
-                        "type": "integer",
-                        "example": 123,
+                    "type": "object",
```
fao89, line 76: @daviddavis customized pagination for v3 had an issue with the schema, it was not following OpenAPI standard

daviddavis: So like `Fixed v3 schema pagination to match OpenAPI standard` would be great.

daviddavis (approved):

PR #489, opened 2021-01-11
Introduce a new  v3/ endpoint returning publication time

fao89:
https://pulp.plan.io/issues/7939
closes #7939

### DIFF
--- a/pulp_ansible/tests/functional/api/collection/v3/test_collection.py
+++ b/pulp_ansible/tests/functional/api/collection/v3/test_collection.py
@@ -2,6 +2,7 @@
 import hashlib
 import logging
 import re
+from datetime import datetime
 from urllib.parse import urljoin
 
 import pytest
@@ -62,15 +63,25 @@ def artifact():
     return artifact
 
 
+def get_metadata_published(pulp_client, pulp_dist):
+    """Return published datetime."""
+    metadata = pulp_client.using_handler(api.json_handler).get(
+        get_galaxy_url(pulp_dist["base_path"], "/v3/")
+    )
+    return datetime.strptime(metadata["published"], "%Y-%m-%dT%H:%M:%S.%fZ")
+
+
 @pytest.fixture(scope="session")
 def collection_upload(pulp_client, artifact, pulp_dist):
     """Publish a new collection and return the processed response data."""
     UPLOAD_PATH = get_galaxy_url(pulp_dist["base_path"], "/v3/artifacts/collections/")
-
+    published_before_upload = get_metadata_published(pulp_client, pulp_dist)
     logging.info(f"Uploading collection to '{UPLOAD_PATH}'...")
     collection = {"file": (ANSIBLE_COLLECTION_FILE_NAME, open(artifact.filename, "rb"))}
 
     response = pulp_client.using_handler(upload_handler).post(UPLOAD_PATH, files=collection)
+    published_after_upload = get_metadata_published(pulp_client, pulp_dist)
+    assert published_after_upload > published_before_upload
     return response
 
 
--- a/CHANGES/7939.feature
+++ b/CHANGES/7939.feature
@@ -1,0 +1,1 @@
+Introduce a new ``v3/`` endpoint returning publication time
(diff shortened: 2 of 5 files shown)


newswangerd (approved):

bmbouter (changes requested):
What do you think about adding a test or tests for this endpoint?

fao89:
> What do you think about adding a test or tests for this endpoint?

AFAIK this endpoint is already tested at https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/cli/test_collection_install.py

bmbouter:
But where are the assertions on the new data returned?

fao89:
> But where are the assertions on the new data returned?

it does not introduce new data, just replaced the way the serializer gets the `ContentArtifact`

bmbouter:
> > But where are the assertions on the new data returned?
> 
> it does not introduce new data, just replaced the way the serializer gets the `ContentArtifact`

Maybe this comment was for the other PR?

fao89:
> > > But where are the assertions on the new data returned?
> > 
> > 
> > it does not introduce new data, just replaced the way the serializer gets the `ContentArtifact`
> 
> Maybe this comment was for the other PR?

my bad!
Actually all the comments were for the other PR

bmbouter:
> > > > But where are the assertions on the new data returned?
> > > 
> > > 
> > > it does not introduce new data, just replaced the way the serializer gets the `ContentArtifact`
> > 
> > 
> > Maybe this comment was for the other PR?
> 
> my bad!
> Actually all the comments were for the other PR

There is so much going on ... totally understandable!

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
```diff
@@ -62,15 +63,25 @@ def artifact():
...
     """Publish a new collection and return the processed response data."""
     UPLOAD_PATH = get_galaxy_url(pulp_dist["base_path"], "/v3/artifacts/collections/")
-
+    published_before_upload = get_metadata_published(pulp_client, pulp_dist)
     logging.info(f"Uploading collection to '{UPLOAD_PATH}'...")
     collection = {"file": (ANSIBLE_COLLECTION_FILE_NAME, open(artifact.filename, "rb"))}
 
     response = pulp_client.using_handler(upload_handler).post(UPLOAD_PATH, files=collection)
+    published_after_upload = get_metadata_published(pulp_client, pulp_dist)
+    assert published_after_upload > published_before_upload
```
fao89, line 84: @bmbouter is this enough or should I do a specific test?

bmbouter: This is ok, but you need to use the test library to assert and not `assert` which could be disabled depending on how Python is compiled.

fao89: so this entire file is affected, all the assertions are done like this

bmbouter: yes none of them should use `assert`.

bmbouter (changes requested):
This tested nicely for me. If you can fix the assert statement I'll lgtm.

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
fao89: apparently pytest rewrites the assertions: https://docs.pytest.org/en/2.1.0/assert.html#advanced-assertion-introspection

bmbouter: Wow that's pretty amazing. Still though do we want to rely on this? I'm ok if so but I want to ask.

bmbouter: @fao89 how about we merge this now since the other code uses this style already and would you be willing to put it on the next pulp_ansible weekly agenda?

fao89: sounds good to me

bmbouter (approved):
This looks great. Thank you @fao89 !

PR #502, opened 2021-01-21
Test galaxy

fao89:
[noissue]


fao89:
@daviddavis wdyt about testing galaxy on nightly s3 jobs?

daviddavis:
Good question. I'd say no for now unless/until we can identify its value? I don't have a strong opinion though (maybe galaxy does?).

fao89:
closing for now

PR #503, opened 2021-01-25
Export/Import for pulp_ansible

gerrod3:
fixes: #6738

### DIFF
--- a/pulp_ansible/app/modelresource.py
+++ b/pulp_ansible/app/modelresource.py
@@ -1,0 +1,103 @@
+from import_export import fields
+from import_export.widgets import ManyToManyWidget
+from pulpcore.plugin.importexport import BaseContentResource, QueryModelResource
+from pulp_ansible.app.models import (
+    Role,
+    Collection,
+    Tag,
+    CollectionVersion,
+)
+
+
+class RoleContentResource(BaseContentResource):
+    """
+    Resource for import/export of ansible_role-content entities.
+    """
+
+    def set_up_queryset(self):
+        """
+        :return: Role content specific to a specified repo-version.
+        """
+        return Role.objects.filter(pk__in=self.repo_version.content)
+
+    class Meta:
+        model = Role
+        import_id_fields = model.natural_key_fields()
+
+
+class CollectionVersionContentResource(BaseContentResource):
+    """
+    Resource for import/export of ansible_collectionversion-content entities.
+    """
+
+    tags = fields.Field(
+        column_name="tags", attribute="tags", widget=ManyToManyWidget(Tag, field="name")
+    )
+
+    def before_import_row(self, row, **kwargs):
+        """
+        Finds and sets collection using name and namespace.
+
+        Args:
+            row (tablib.Dataset row): incoming import-row representing a single CollectionVersion.
+            kwargs: args passed along from the import() call.
+        """
+        col = Collection.objects.get(name=row["name"], namespace=row["namespace"])
+        row["collection"] = str(col.pk)
+
+    def set_up_queryset(self):
+        """
+        :return: CollectionVersion content specific to a specified repo-version.
+        """
+        return CollectionVersion.objects.filter(pk__in=self.repo_version.content)
+
+    class Meta:
+        model = CollectionVersion
+        import_id_fields = model.natural_key_fields()
+
+
+class CollectionResource(QueryModelResource):
+    """
+    Resource for import/export of ansible_collection entities.
+    """
+
+    def set_up_queryset(self):
+        """
+        :return: Collections specific to a specified repo-version.
+        """
+        collection_versions = CollectionVersion.objects.filter(pk__in=self.repo_version.content)
+        return Collection.objects.filter(
+            pk__in=collection_versions.values_list("collection", flat=True)
+        )
+
+    class Meta:
+        model = Collection
+        import_id_fields = ("namespace", "name")
+
+
+class TagResource(QueryModelResource):
+    """
+    Resource for import/export of ansible_tag entities.
+    """
+
+    def set_up_queryset(self):
+        """
+        :return: Tags specific to a specified repo-version.
+        """
+        collection_versions = CollectionVersion.objects.filter(pk__in=self.repo_version.content)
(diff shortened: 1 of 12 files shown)


pulp_ansible/tests/functional/utils.py
```diff
@@ -1,4 +1,5 @@
 """Utilities for tests for the ansible plugin."""
+from dynaconf import Dynaconf
```
fao89, line 2: Our CI tests run client-side (outside the container), pointing to the server-side (container). 
`dynaconf` is installed server-side, so you need to add `dynaconf` here: https://github.com/pulp/pulp_ansible/blob/master/functest_requirements.txt to have it installed client-side

gerrod3: What is the file "/etc/pulp/settings.py" for the CI? I need the directory "/tmp" to be in ALLOWED_EXPORT_PATHS and ALLOWED_IMPORT_PATHS

fao89: it comes from template_config.yml:
https://github.com/pulp/pulp_file/blob/c0bc20b5e8d2aba354cb8c7a246816cd7c7c99ef/template_config.yml#L34
after adding it, you need to use plugin-template:
```
./plugin-template --github pulp_ansible
```
plugin-template converts settings from template_config.yml into CI scripts

daviddavis: We set this here: https://github.com/pulp/pulp_ansible/blob/master/.ci/ansible/settings.py.j2#L9-L13

Comes from: https://github.com/pulp/pulpcore/blob/master/template_config.yml#L32-L42

pulp_ansible/app/modelresource.py
```diff
@@ -0,0 +1,97 @@
...
+        import_id_fields = model.natural_key_fields()
+
+
+class CollectionVersionContentResource(BaseContentResource):
+    """
+    Resource for import/export of ansible_collectionversion-content entities.
+    """
+
+    tags = fields.Field(
+        column_name="tags", attribute="tags", widget=ManyToManyWidget(Tag, field="name")
```
daviddavis, line 34: :+1:

pulp_ansible/app/modelresource.py
```diff
@@ -0,0 +1,97 @@
...
+    )
+    collection = fields.Field(
+        column_name="collection",
+        attribute="collection",
+        widget=ForeignKeyWidget(Collection, "name"),
```
daviddavis, line 39: Thinking through this some: I don't think collection name is unique. I think we also need namespace.

pulp_ansible/tests/functional/api/test_import.py
```diff
@@ -0,0 +1,152 @@
...
+        delete_orphans(cls.cfg)
+
+        cli_client = cli.Client(cls.cfg)
+        cmd = ("rm", "-rf", cls.exporter.path)
+        cli_client.run(cmd, sudo=True)
+
+    def _clean_database(self):
+        cli_client = cli.Client(self.cfg)
+        cmd = get_psql_smash_cmd(TRUNCATE_TABLES_QUERY_BASH)
+        cli_client.run(cmd, sudo=True)
```
daviddavis, line 72: :rocket:

pulp_ansible/tests/functional/api/test_import.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def test_import(self):
+        """Test an import."""
+        self._setup_import()
+        importer = self._create_importer()
+        task_group = self._perform_import(importer)
+        self.assertEqual(len(self.import_repos) + 1, task_group.completed)
+        for repo in self.import_repos:
+            repo = self.repo_api.read(repo.pulp_href)
+            self.assertEqual(f"{repo.pulp_href}versions/1/", repo.latest_version_href)
```
daviddavis, line 137: Can you test that come of the content/tags/etc got imported?

func_test_settings.py
```diff
@@ -0,0 +1 @@
+DATABASES = {"default": {"NAME": "pulp", "USER": "pulp", "PASSWORD": "pulp"}}
```
daviddavis, line 1: Can this be moved out of the root dir? Maybe into pulp_ansible/tests/assets or something?

daviddavis:
What about a workflow doc outlining a simple export/import? Doesn't have to be very long.

gerrod3:
> What about a workflow doc outlining a simple export/import? Doesn't have to be very long.

Is import available on the CLI, I don't think it is? I would like to have all of it using the CLI, so I think I'm going to file an issue to create the workflow once the CLI has import. If you want it in this PR using a mix of CLI and httpie commands I can add it.

.ci/ansible/Containerfile.j2
```diff
@@ -10,7 +10,7 @@ ADD {{ item.source }} {{ item.source }}
 # Install python packages
 # Hacking botocore (https://github.com/boto/botocore/pull/1990)
 
-RUN pip3 install \
+RUN pip3 install --use-feature=2020-resolver \
```
fao89, line 13: it seems your plugin-template is outdated

daviddavis:
> I would like to have all of it using the CLI, so I think I'm going to file an issue to create the workflow once the CLI has import.

:+1:

pulp_ansible/tests/functional/api/test_import.py
```diff
@@ -0,0 +1,169 @@
...
+            repo_ver = self.repo_ver_api.read(repo.latest_version_href)
+            added = repo_ver.content_summary.added
+            if "ansible.role" in added.keys():
+                roles = self.croles_api.list()
+                self.assertEqual(roles.count, added["ansible.role"]["count"])
+            elif "ansible.collection_version" in added.keys():
+                cvs = self.cv_api.list()
+                self.assertEqual(cvs.count, added["ansible.collection_version"]["count"])
```
daviddavis, line 152: It looks like this code is checking that the content that got added to the repo version is equal to the number of content is the system? It seems like it would be more useful to compare the content count from the exported repo(s) to repo_ver.content_summary.added.

daviddavis:
I tested this out and it worked great :1st_place_medal:

daviddavis (approved):

PR #488, opened 2021-01-08
New endpoint returning all collections unpaginated

fao89:
https://pulp.plan.io/issues/7940
closes #7940

PR #505, opened 2021-01-26
Create a new endpoint to return collection level dependencies

bmclaughlin:
Create a new ``collections/ns/name/versions/version/dependencies/`` endpoint that returns collection level dependencies.

https://pulp.plan.io/issues/8132
closes #8132

### DIFF
--- a/CHANGES/8132.feature
+++ b/CHANGES/8132.feature
@@ -1,0 +1,1 @@
+Create a new ``collections/ns/name/versions/version/dependencies/``endpoint that returns collection level dependencies.
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -99,6 +99,29 @@ class CollectionVersionListSerializer(serializers.ModelSerializer):
         )
 
 
+class CollectionVersionDependencySerializer(CollectionSerializer):
+    """A serializer for returning a CollectionVersion's dependencies."""
+
+    dependencies = serializers.JSONField()
+
+    class Meta:
+        fields = ("version", "dependencies")
+        model = models.CollectionVersion
+
+    def get_href(self, obj):
+        """
+        Get href.
+        """
+        kwargs = {
+            "path": self.context["path"],
+            "namespace": obj.namespace,
+            "name": obj.name,
+            "version": obj.version,
+        }
+
+        return self._get_href("collection-version-dependency", **kwargs)
+
+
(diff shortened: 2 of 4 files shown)


bmbouter (changes requested):
Please add tests that drive this endpoint and assert on its results.

bmbouter:
We plan to release 0.7 on Feb 9th so this needs to be merged by Feb 8th if you want it in that release.

CHANGES/8132.feature
```diff
@@ -0,0 +1 @@
+Create a new ``collections/ns/name/versions/version/dependencies/``endpoint that returns collection level dependencies.
```
webknjaz, line 1:
```suggestion
Create a new ``collections/ns/name/versions/version/dependencies/`` endpoint that returns all collection transitive dependencies.
```

bmclaughlin:
Closing in favor of the implementation at https://github.com/ansible/ansible/pull/72591.  :+1:

jborean93:
I agree with the comments made here there are a lot of things that need to happen on the client side and I think that having the API resolve the dependencies will only solve some of those cases and not all making it hard to implement this new API in the cli. What might be better is exposing an API were a client can get a manifest of data in 1 go that includes things like

* All the collections
* Their versions
* Collection info like
    * Dependencies
    * Plugin information

So instead of making tens to hundreds of API calls to get the collection information the client only gets this manifest info and now it has all the data it needs to resolve the collections itself in whatever way it wishes. It can regularly get the updated manifest. I'm no API expert but it seems like tools like `dnf`/`apt` provide something like this to make their tools more efficient.

bmbouter:
@jborean93 fyi check out these new galaxy V3 apis that just got added for exactly that reason: https://github.com/pulp/pulp_ansible/pull/488/files

jborean93:
Fantastic that should help the cli a lot more, any chance of it being done to the `galaxy.ansible` v2 endpoint?

fao89:
@jborean93 do you mean v2 endpoints on pulp side? 
If so, yes we can do it, but if you mean https://galaxy.ansible.com/api/v2/ we can't do it, since it is not pulp code -  https://github.com/ansible/galaxy

jborean93:
I mean the latter, the bulk of the `ansible-galaxy` cli users will be people talking to galaxy.ansible.com so even if we improve the experience on Automation Hub it won't really help the tool overall as Galaxy is still using the slower API. Understandable you can't do much because it doesn't use Pulp but maybe it's a topic to bring up with the Galaxy team to get it to use the v2 Pulp API.

PR #506, opened 2021-01-26
Roles and Collections workflows use Pulp-CLI commands

gerrod3:
fixes: #8076

### DIFF
--- a/CHANGES/8076.doc
+++ b/CHANGES/8076.doc
@@ -1,0 +1,1 @@
+Updated Roles and Collections workflows to use Pulp-CLI commands
--- a/docs/_scripts/distribution_repo_version.sh
+++ b/docs/_scripts/distribution_repo_version.sh
@@ -1,14 +1,3 @@
-# Distributions are created asynchronously. Create one, and specify the publication that will
+# Distributions are created asynchronously. Create one, and specify the repository version that will
 # be served at the base path specified.
-export TASK_HREF=$(http POST $BASE_ADDR/pulp/api/v3/distributions/ansible/ansible/ \
-  name='baz' \
-  base_path='my_content' \
-  repository_version=${REPO_HREF}versions/1/ | jq -r '.task')
-
-# Poll the task (here we use a function defined in docs/_scripts/base.sh)
-# When the task is complete, it gives us the href for our new Distribution
-wait_until_task_finished $BASE_ADDR$TASK_HREF
-export DIST_PATH=${CREATED_RESOURCE[0]}
-
-# Lets inspect the Distribution
-http $BASE_ADDR$DIST_PATH
+pulp ansible distribution create --name "baz" --base-path "my_content" --repository "foo" --version 1
--- a/docs/_scripts/setup.sh
+++ b/docs/_scripts/setup.sh
@@ -1,33 +1,18 @@
-#!/usr/bin/env bash
-echo "Setting environment variables for default hostname/port for the API and the Content app"
-export BASE_ADDR=${BASE_ADDR:-http://localhost:24817}
-export CONTENT_ADDR=${CONTENT_ADDR:-http://localhost:24816}
+# Install from PyPI
+pip install pulp-cli[pygments]  # colorized output
+pip install pulp-cli  # no color output
 
-# Necessary for `django-admin`
-export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+# Install from source
+git clone https://github.com/pulp/pulp-cli.git # or your fork url
+cd pulp-cli
+pip install -e .
+cd ..
 
-# Poll a Pulp task until it is finished.
-wait_until_task_finished() {
-    echo "Polling the task until it has reached a final state."
-    local task_url=$1
-    while true
-    do
-        local response=$(http $task_url)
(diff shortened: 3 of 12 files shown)


docs/_scripts/setup.sh
```diff
@@ -1,33 +1,8 @@
...
-                echo "Still waiting..."
-                sleep 2
-                ;;
-        esac
-    done
-}
+# Install from source
+git clone https://github.com/pulp/pulp-cli.git # or your fork url
+cd pulp-cli
+pip install -e .
```
fao89, line 8:
```suggestion
pip install -e .
cd ..
```

fao89:
the CI creates a .netrc file:
https://github.com/pulp/pulp_ansible/blob/master/.github/workflows/scripts/post_before_script.sh
not sure how can it affect the CLI

gerrod3:
> the CI creates a .netrc file:
> https://github.com/pulp/pulp_ansible/blob/master/.github/workflows/scripts/post_before_script.sh
> not sure how can it affect the CLI

The netrc file should be fine. The CLI only uses that file for username/password config and that file you link has the same defaults. I need to add a config file so it knows where to find the Pulp server.

CHANGES/8076.doc
```diff
@@ -0,0 +1,2 @@
+Updated Roles and Collections workflows to use Pulp-CLI commands
+
```
bmbouter, line 2: nitpick: the extra newline isn't necessary it'll create a gap in the docs.

docs/_scripts/distribution_repo_version.sh
```diff
@@ -1,14 +1,3 @@
-# Distributions are created asynchronously. Create one, and specify the publication that will
```
bmbouter, line 1: Is there value in keeping these `.sh` files with only one command each? I was thinking of having them embedded in the docs directly and escapped as bash? wdyt?

bmbouter: Like maybe we're testing with these somehow?

gerrod3: I don't think that's a bad idea. I only kept them in the scripts because the CI runs them to ensure they work, but the CLI already has those tests, so it's probably redundant.

bmbouter:
I am very excited about this change! Besides my question about keeping the `.sh` files I think this PR looks good to merge (to me).

docs/_scripts/distribution_repo_version.sh
bmbouter: I've seen projects test their command snippets by having a sphinx plugin handle it. I'm in favor of dropping all .sh file usage. @fao89 and @daviddavis wdyt?

docs/_scripts/sync-collection.sh
```diff
@@ -0,0 +1,9 @@
+# Sync repository foo using remote cbar
+pulp ansible repository sync --name "foo" --remote "cbar"
+
+# Use the -b option to have the sync task complete in the background
+# e.g. pulp -b ansible repository sync --name "foo" --remote "cbar"
+
+# After the task is complete, it gives us a new repository version
+# Inspecting new repository version
+pulp ansible repository version --repository "foo" show --version 1
```
fao89, line 9: is it possible to display all versions? (if not, maybe we need different repos for testing roles/collections)
since we sync roles first, version 1 is displaying role content on CI

docs/_scripts/distribution_repo_version.sh
fao89: pulp_ansible is the first plugin to move docs to CLI, maybe we need to discuss it on the open floor?

docs/_scripts/sync-collection.sh
gerrod3: Yes, there is a list command that would show all of them. I kept this at version 1 so that users could copy and paste the commands if they are following the collection workflow, but since I didn't create a new repo the CI would be showing the wrong version.  Maybe I should create a PR for the CLI to allow the show command to accept no version and default to showing the latest.

docs/_scripts/distribution_repo_version.sh
daviddavis: So the commands would live in the rst docs and then the plugin would extract + run them? Is that what you're proposing?

docs/_scripts/sync-collection.sh
fao89: +1 to showing the latest command!

bmbouter:
I think we should consider moving away from these the CI testing the `.sh` files and replace them with https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html usage.

docs/_scripts/distribution_repo_version.sh
bmbouter: Yes using doctest hopefully.

daviddavis (approved):

bmbouter (approved):
I am very excited about this!

gerrod3:
Created an issue to keep track of @bmbouter suggestion: https://pulp.plan.io/issues/8170. Will create a future PR to implement.

fao89 (approved):
Thank you Gerrod!

PR #507, opened 2021-01-27
Use rate_limit from pulpcore

fao89:
Required PR: https://github.com/pulp/pulpcore/pull/1091
[noissue]


bmbouter (approved):
This looks right to me. Thanks @fao89 !

newswangerd:
@bmbouter which version of pulpcore includes rate limiting?

fao89:
> @bmbouter which version of pulpcore includes rate limiting?

it will be 3.10

PR #509, opened 2021-01-27
New endpoint returning all collection_versions unpaginated

fao89:
https://pulp.plan.io/issues/7941
closes #7941

### DIFF
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -182,7 +182,7 @@ class CollectionVersionSerializer(CollectionVersionListSerializer):
         """
         Get atrifact summary.
         """
-        return ArtifactRefSerializer(self.context["content_artifact"]).data
+        return ArtifactRefSerializer(obj.contentartifact_set.get()).data
 
     def get_download_url(self, obj) -> str:
         """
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -90,7 +90,9 @@ def get_queryset(self):
         """
         distro_content = self.get_distro_content(self.kwargs["path"])
 
-        collections = CollectionVersion.objects.select_related("collection").filter(
+        collections = CollectionVersion.objects.select_related(
+            "content_ptr__contentartifact"
```
bmbouter, line 94: Maybe this change is good to make in general because we were running the query one-by-one on `retrieve` and now we're doing it once in the get_queryset which is likely better. Do we need to make this as part of this PR though? I was wondering because I see get_queryset being redefined once again on UnpaginatedCollectionVersionViewSet which I believe is a descendant of this mixin right?

fao89: you are right, it doesn't necessarily need to be on this PR

bmbouter: Want to break it out to another one? It could be a [noissue] I think.

fao89: oh! wait!
actually, this change is needed, because it affects how we use the serializer

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -182,15 +182,15 @@ def get_artifact(self, obj):
         """
         Get atrifact summary.
         """
-        return ArtifactRefSerializer(self.context["content_artifact"]).data
```
fao89, line 185: this is the reason we need the change on this PR

PR #510, opened 2021-01-28
Remove rate_limit from CollectionRemoteSerializer

fao89:
[noissue]


newswangerd:
@fao89 why was this removed? This is something we'd like to allow customers to configure

fao89:
> @fao89 why was this removed? This is something we'd like to allow customers to configure

it was moved to pulpcore,https://github.com/pulp/pulpcore/pull/1091
pulp-ansible 0.7.0 will get it from pulpcore 3.10,
if you want it in pulp-ansible < 0.7 we can do it

PR #511, opened 2021-01-28
Improve sync performance with no-op when possible

fao89:
https://pulp.plan.io/issues/7942
closes #7942


pulp_ansible/app/viewsets.py
```diff
@@ -245,6 +245,11 @@ class CollectionRemoteViewSet(RemoteViewSet):
     queryset = CollectionRemote.objects.all()
     serializer_class = CollectionRemoteSerializer
 
+    def update(self, request, pk, **kwargs):
+        if "requirements_file" in request.data:
+            request.data["last_synced_metadata_time"] = None
```
fao89, line 250: if we change the `requirements_file` unset `last_synced_metadata_time`.

for avoiding no-op when a new requirements file contains more collections than the previous one

pulp_ansible/app/tasks/collections.py
```diff
@@ -503,6 +512,27 @@ async def run(self):
...
+                pass
+
+            collection_info = self.collection_info or []
+            sources = {r.source for r in collection_info if r.source}
+            sources.add(self.remote.url)
+
+            published = metadata.get("published")
+            if published == self.remote.last_synced_metadata_time and len(sources) < 2:
+                log.info(_("no-op"))
+                return
```
fao89, line 532: we may have `self.remote.url` and `sources` pointing to different repos, should we only consider `self.remote.url` for no-op?

bmbouter:
@fao89 maybe we should break this story into two issues: one for just using the sync endpoint and then approach no-op separately? I'm ok to keep it together also. Wdyt?

fao89:
> @fao89 maybe we should break this story into two issues: one for just using the sync endpoint and then approach no-op separately? I'm ok to keep it together also. Wdyt?

I can rename this PR and it would be the no-op PR

pulp_ansible/app/viewsets.py
bmbouter: I think this is good. I didn't think about this. :+1:

bmbouter: Can a test be added for this part please?

pulp_ansible/app/tasks/collections.py
```diff
@@ -503,6 +512,27 @@ async def run(self):
...
+
+            self.last_synced_metadata_time = published
+
```
bmbouter, line 535: I think this should be moved to its own function that is called prior to `_fetch_collection_version_metadata`. I think we want to keep the function lengths small (and responsibilities separated) to keep things manageable.

pulp_ansible/app/tasks/collections.py
```diff
@@ -304,12 +308,13 @@ def __init__(self, remote):
...
+        self.last_synced_metadata_time = None
 
         # Interpret download policy
         self.deferred_download = self.remote.policy != Remote.IMMEDIATE
 
     @alru_cache(maxsize=128)
-    async def _get_collection_api(self, root):
+    async def _get_collection_api_root(self, root):
         """
         Returns the collection api path and api version.
```
bmbouter, line 319: This docstring is out of date now I think.

pulp_ansible/app/tasks/collections.py
```diff
@@ -340,10 +345,14 @@ async def _get_collection_api(self, root):
         else:
             raise RuntimeError(_("Unsupported API versions at {}").format(root))
 
-        endpoint = f"{root}v{api_version}/collections/"
+        endpoint = f"{root}v{api_version}"
 
         return endpoint, api_version
 
+    async def _get_collection_api(self, root):
```
bmbouter, line 352: This needs a docstring I think.

pulp_ansible/app/tasks/collections.py
```diff
@@ -503,6 +512,27 @@ async def run(self):
...
+        if api_version == 3:
+            downloader = self.remote.get_downloader(
+                url=root, silence_errors_for_response_status_codes={404}
+            )
+            try:
+                metadata = parse_metadata(await downloader.run())
+            except FileNotFoundError:
+                pass
+
+            collection_info = self.collection_info or []
```
bmbouter, line 525: I don't expect an `or []` to be needed here. Is it needed?

pulp_ansible/app/tasks/collections.py
```diff
@@ -503,6 +512,27 @@ async def run(self):
...
+            sources = {r.source for r in collection_info if r.source}
```
bmbouter, line 526: Can this be unrolled to be more readable. I had to read it 3 times and while it's allowed I think we want code to be super readable.

pulp_ansible/app/tasks/collections.py
```diff
@@ -503,6 +512,27 @@ async def run(self):
...
+            sources.add(self.remote.url)
+
+            published = metadata.get("published")
+            if published == self.remote.last_synced_metadata_time and len(sources) < 2:
+                log.info(_("no-op"))
```
bmbouter, line 531: Can this log message be expanded some. It would be good if it also logged if it's being no-opped due to too many sources, versus the published time not changing.

pulp_ansible/app/tasks/collections.py
bmbouter: Yeah maybe we should leave off the inspection of sources from this PR at least to keep it simple from the start... wdyt?

pulp_ansible/app/tasks/collections.py
```diff
@@ -503,6 +512,27 @@ async def run(self):
...
+                return
+
+            self.last_synced_metadata_time = published
```
bmbouter, line 534: Does this need to be saved?

bmbouter:
I think another part should be added to this PR, a new, optional sync parameter called `force`. When `force=True` the no-op is skipped no matter what. When `force=False` a no-op could occur. If unspecified the sync defaults to `force=False`.

I think having the `force` option as a sync param will allow us to not worry so much about the `sources` corner cases. If a user is getting a no-op and they don't want one, they just call with `force=True`.

pulp_ansible/app/tasks/collections.py
fao89: for successful sync, we want to keep the published time to compare with future syncs

bmbouter: I agree, I'm asking where is the call to `self.save()`

pulp_ansible/app/tasks/collections.py
```diff
@@ -84,6 +84,10 @@ def sync(remote_pk, repository_pk, mirror):
     d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
     d_version.create()
 
+    if first_stage.last_synced_metadata_time:
+        remote.last_synced_metadata_time = first_stage.last_synced_metadata_time
+        remote.save()
```
fao89, line 89: @bmbouter I only save published time when the sync is successful

fao89:
> I think another part should be added to this PR, a new, optional sync parameter called `force`. When `force=True` the no-op is skipped no matter what. When `force=False` a no-op could occur. If unspecified the sync defaults to `force=False`.
> 
> I think having the `force` option as a sync param will allow us to not worry so much about the `sources` corner cases. If a user is getting a no-op and they don't want one, they just call with `force=True`.

on pulp_rpm we specify `optimize=True`, I changed to this approach

pulp_ansible/tests/functional/api/collection/v3/test_sync.py
```diff
@@ -24,32 +24,54 @@ class SyncCollectionsFromPulpServerTestCase(TestCaseUsingBindings, SyncHelpersMi
...
+            requirements_file=self.requirements_file,
+        )
+        second_remote = self.remote_collection_api.create(second_body)
+        self.addCleanup(self.remote_collection_api.delete, second_remote.pulp_href)
+
+        second_repo = self._create_repo_and_sync_with_remote(second_remote)
+
+        # Resync
+        self._sync_repo(second_repo, remote=second_remote.pulp_href, optimize=True)
+
```
fao89, line 71: I don't know how to assert the log, the test run on client side and the log happens on server side

pulp_ansible/app/models.py
```diff
@@ -235,6 +235,8 @@ class AnsibleRepository(Repository):
     CONTENT_TYPES = [Role, CollectionVersion]
     REMOTE_TYPES = [RoleRemote, CollectionRemote]
 
+    last_synced_metadata_time = models.DateTimeField(null=True)
```
bmbouter, line 238: I think we should add this to the docstring

pulp_ansible/app/viewsets.py
```diff
@@ -245,6 +246,19 @@ class CollectionRemoteViewSet(RemoteViewSet):
...
+        if "url" in request.data or "requirements_file" in request.data:
+            repos = list(
+                AnsibleRepository.objects.filter(
+                    remote_id=pk, last_synced_metadata_time__isnull=False
+                ).all()
+            )
+            for repo in repos:
+                repo.last_synced_metadata_time = None
+            AnsibleRepository.objects.bulk_update(repos, ["last_synced_metadata_time"])
+        return super().update(request, pk, **kwargs)
```
bmbouter, line 260: I think this should all be wrapped in one db transaction.

CHANGES/7942.feature
```diff
@@ -0,0 +1 @@
+Improve sync performance with no-op when possible.
```
bmbouter, line 1: Please add the fact that the no-op can be disabled with the `optimize=False` sync parameter also.

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +626,60 @@ async def _find_all_collections(self):
...
+        return False
+
     async def run(self):
         """
         Build and emit `DeclarativeContent` from the ansible metadata.
         """
+        if self.repository.remote and self.remote.pk == self.repository.remote.pk:
+            can_be_optimized = await self._can_be_optimized()
+            if self.optimize and can_be_optimized:
+                return
```
bmbouter, line 676: It would be good to add a logging statement at debug level here indicating that the no-op occurred also.

pulp_ansible/app/tasks/collections.py
bmbouter: So where does this self object get saved?

CHANGES/7942.feature
fao89: it is disabled by default, we need to set it to true to enable

bmbouter: Oh I think optimize should be enabled by default.

Regardless though, we need to identify that this new option is available to the user.

pulp_ansible/app/viewsets.py
fao89: super() triggers a task, not sure if it would be effective to add it

CHANGES/7942.feature
bmbouter: I think we definitly want it enabled by default. This is also how other content types work, e.g. rpm.

pulp_ansible/tests/functional/api/collection/v3/test_sync.py
```diff
@@ -75,6 +86,53 @@ def test_sync_collections_from_pulp_using_mirror_second_time(self):
         second_content = self.cv_api.list(repository_version=f"{second_repo.pulp_href}versions/1/")
         self.assertGreaterEqual(len(second_content.results), 1)
 
+    def test_noop_resync_collections_from_pulp(self):
+        """Test sync collections from pulp server."""
```
bmbouter, line 90: Can this be more descriptive. This seems not accurate for this test.

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +626,61 @@ async def _find_all_collections(self):
...
+        return False
+
     async def run(self):
         """
         Build and emit `DeclarativeContent` from the ansible metadata.
         """
+        if self.repository.remote and self.remote.pk == self.repository.remote.pk:
+            can_be_optimized = await self._can_be_optimized()
+            if self.optimize and can_be_optimized:
+                log.debug("no-op: repmote wasn't updated since last sync.")
```
bmbouter, line 676:
```suggestion
                log.debug(_("no-op: remote data has not changed since last sync."))
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +626,61 @@ async def _find_all_collections(self):
...
+            )
+            try:
+                metadata = parse_metadata(await downloader.run())
+            except FileNotFoundError:
+                return False
+
+            sources = set()
+            if self.collection_info:
+                sources = {r.source for r in self.collection_info if r.source}
+            sources.add(self.remote.url)
```
bmbouter, line 644: I think having the > 2 check right here are returning False would be better than leaving the variable to be reused further down in a more complex check.

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +626,61 @@ async def _find_all_collections(self):
...
+
+            published = metadata.get("published")
+            repo_metadata_time = self.repository.last_synced_metadata_time
+            repo_metadata_time = repo_metadata_time.isoformat() if repo_metadata_time else ""
+
+            # From serializers.DateTimeField.to_representation:
+            if repo_metadata_time.endswith("+00:00"):
+                repo_metadata_time = repo_metadata_time[:-6] + "Z"
```
bmbouter, line 652: How can this part be made to have `published and `repo_metadata_time` both be DateTime objects instead of strings? I think we want to use DateTime for comparisons when those strings represent time.

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +626,61 @@ async def _find_all_collections(self):
...
+
+            if published == repo_metadata_time and len(sources) < 2:
+                msg = _(
+                    "no-op: {repo} did not change since last sync - {published}".format(
+                        repo=self.remote.url, published=published
+                    )
+                )
+                noop = ProgressReport(message=msg, code="noop")
+                noop.state = TASK_STATES.COMPLETED
+                noop.save()
```
bmbouter, line 662: It I think would be more reliable to report to the user that the no-op check occured anytime it did occur. To do that I would move this part higher up in the function before any `return` statements (maybe even at the beginning).

pulp_ansible/app/viewsets.py
```diff
@@ -245,6 +247,20 @@ class CollectionRemoteViewSet(RemoteViewSet):
     queryset = CollectionRemote.objects.all()
     serializer_class = CollectionRemoteSerializer
 
+    def update(self, request, pk, **kwargs):
```
bmbouter, line 250: I forgot that this `super()` call runs inside a task. Since that's the case we need to do this differently for correctness reasons. I recommend doing these things:

1) Create a new task called `update_collection_remote` and have it contain this update code.
2) Instead of calling super() which would dispatch another task have your `update_collection_remote` task call the `tasks.base.general_update` [in the saw way that the super does](https://github.com/pulp/pulpcore/blob/5a6d5c5f2ebe06922319235b3c845608d10e6f3a/pulpcore/app/viewsets/base.py#L454).
3) You'll need to ensure the `update_collection_remote` task not only locks on the remote being updated but also the AnsibleRepository (this is the main correctness change also)
4) Since you're not calling super have your code return the `OperationPostponed`.
5) Also have you `update_collection_remote` use a Transaction so all postgresql operations happen or none of them do.

pulp_ansible/app/tasks/collections.py
fao89: this is actually what this code is doing, converting the string it gets from serializer to datetime

pulp_ansible/app/tasks/collections.py
fao89: this is what this code is doing, no-op only happens when return True, this is the only part when it returns True

bmbouter: What I'm saying is the progress reporting should say that it is checking for a no-op regardless of it if occurs or not. In terms of reporting on what it's doing I expect this progress report to be there when optimize=True and not to be there when optimize=False.

pulp_ansible/app/viewsets.py
fao89: it may be zero or more than one repository, so I don't think it is possible to lock on repository

pulp_ansible/app/tasks/collections.py
fao89: https://github.com/pulp/pulp_ansible/pull/511#discussion_r566897461

pulp_ansible/app/tasks/collections.py
fao89: so how can I assert if no-op happened?
I added the ProgressReport just for it

bmbouter: I expected it's not on asserting that the progress reporting shows no-op was checked or not, it's that the rest of the 'syncing, etc, etc' progress reports are not present and the state is `completed`.

pulp_ansible/app/tasks/collections.py
```diff
@@ -294,23 +342,30 @@ class CollectionSyncFirstStage(Stage):
     The first stage of a pulp_ansible sync pipeline.
     """
 
-    def __init__(self, remote):
+    def __init__(self, remote, repository, optimize=False):
```
bmbouter, line 345: I see the release note chnaged, should this also be `optimize=True` for the default?

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +663,56 @@ async def _find_all_collections(self):
...
+            try:
+                metadata = parse_metadata(await downloader.run())
+            except FileNotFoundError:
+                return False
+
+            sources = set()
+            if self.collection_info:
+                sources = {r.source for r in self.collection_info if r.source}
+            sources.add(self.remote.url)
+            has_multiple_sources = len(sources) > 1
```
bmbouter, line 682: Can this just `return False` here?

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +663,56 @@ async def _find_all_collections(self):
...
+
+            published = parse_datetime(metadata.get("published", ""))
+
+            if published == self.repository.last_synced_metadata_time and not has_multiple_sources:
+                msg = _(
+                    "no-op: {repo} did not change since last sync - {published}".format(
```
bmbouter, line 688: I'm suggesting to make this a report that the no-op was checked, not that it occurred. So that would have it be reporting in all cases unless optimize=False at which point I expect Pulp not to evaluate the no-op case at all.

pulp_ansible/app/tasks/collections.py
fao89: it gets True from here: https://github.com/fao89/pulp_ansible/blob/7942/pulp_ansible/app/serializers.py#L101
like pulp_rpm does

pulp_ansible/app/viewsets.py
bmbouter: Why couldn't it just lock on 0 or more also? Pulp supports multi-locking.

pulp_ansible/app/tasks/collections.py
bmbouter: Got it! Then I guess I'm confused why it's a kwarg at all if there will always be a value specified from the serializer. Do you know why? (I don't).

CHANGES/7942.feature
```diff
@@ -0,0 +1,2 @@
...
+For disabling no-op optimization set ``optimize=False``.
```
bmbouter, line 2: Can this be combine moved to a second sentence in the previous line and also:

```suggestion
To disabled the no-op optimization use the ``optimize=False`` option on the ``sync`` call.
```

pulp_ansible/app/tasks/collections.py
fao89: I don't see the value of having no-op check in progress reports

pulp_ansible/app/tasks/collections.py
bmbouter: Oh actually I see it's always specified in `CollectionSyncFirstStage(remote, repository, optimize)` so this should be an args.

pulp_ansible/app/tasks/collections.py
bmbouter: If you want to take it out I'm ok with that. My thought was that if we don't have any progress report and the no-op runs the user has no info about why. Also it tells the user what was performed differently when they call `optimize=False`.

What we're doing right now though doesn't seem right (to me). Usually progress reports report on what occurred in the task code, but this is more like a message to the user about why the task reached its completed state.

pulp_ansible/app/viewsets.py
```diff
@@ -200,30 +198,33 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
-        serializer = RepositorySyncURLSerializer(
+        serializer = AnsibleRepositorySyncURLSerializer(
             data=request.data, context={"request": request, "repository_pk": repository.pk}
         )
         serializer.is_valid(raise_exception=True)
 
         remote = serializer.validated_data.get("remote", repository.remote)
         remote = remote.cast()
 
+        mirror = serializer.validated_data.get("mirror", False)
```
bmbouter, line 215: This is similar, I don't think we should have a default here if the serializer provides one.

pulp_ansible/app/viewsets.py
```diff
@@ -200,30 +198,33 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+        sync_kwargs = {"remote_pk": remote.pk, "repository_pk": repository.pk, "mirror": mirror}
+
         if isinstance(remote, RoleRemote):
             sync_func = role_sync
         elif isinstance(remote, CollectionRemote):
             sync_func = collection_sync
+            sync_kwargs["optimize"] = serializer.validated_data.get("optimize", False)
```
bmbouter, line 222: Similar here, I don't think this should have a default.

pulp_ansible/app/tasks/collections.py
fao89: @daviddavis wdyt?

pulp_ansible/app/viewsets.py
bmbouter: This looks great

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +663,55 @@ async def _find_all_collections(self):
...
+        if api_version == 3:
+            downloader = self.remote.get_downloader(
+                url=root, silence_errors_for_response_status_codes={404}
+            )
+            try:
+                metadata = parse_metadata(await downloader.run())
+            except FileNotFoundError:
+                return False
+
+            self.last_synced_metadata_time = parse_datetime(metadata.get("published", ""))
```
bmbouter, line 678: In cases where there is no published I think the op-op must return False. So we can be more explicit here with:
```
try:
    self.last_synced_metadata_time = parse_datetime(metadata["published"])
except KeyError:
    return False
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +663,55 @@ async def _find_all_collections(self):
 
         await asyncio.gather(*tasks)
 
+    async def _can_be_optimized(self):
```
bmbouter, line 666: I think this name is confusing. I think can be optimized is the user desire to optimize or not. This is really can we skip the sync or not. With that I'm suggesting:

```suggestion
    async def _should_we_sync(self):
```
Then invert the false and true statements to align with the logic called out by the name.

pulp_ansible/app/tasks/collections.py
daviddavis: > So that would have it be reporting in all cases unless optimize=False at which point I expect Pulp not to evaluate the no-op case at all.

This makes sense to me.

> I'm suggesting to make this a report that the no-op was checked, not that it occurred.

Would this also contain the result of the no-op check? If not, is there some other way a user could determine if a sync task was no-op?

bmbouter: With what I imagined (which the test would actually be asserting on) the only way the user would know is they would see the progress report that the no-op was checked and that there are no other progress reports, e.g. metadata downloading, etc. That's not really explicit though.

daviddavis: It's not explicit but it's probably still obvious enough for users. I don't have a strong opinion here other than thinking we should show something to users even if it's just showing that there was a no-op check performed.

pulp_ansible/app/tasks/collections.py
```diff
@@ -64,7 +65,43 @@
 log = logging.getLogger(__name__)
 
 
-def sync(remote_pk, repository_pk, mirror):
+def update_collection_remote(remote_pk, *args, **kwargs):
+    """
+    Update a CollectionRemote.
```
bmbouter, line 70: Can this docstring get small paragraph after that describes the other behaviors it does besides updating a CollectionRemote?

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +663,60 @@ async def _find_all_collections(self):
 
         await asyncio.gather(*tasks)
 
+    async def _should_we_sync(self):
```
bmbouter, line 666: I think the True/False return statements in this function need to be inverted.

pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +663,60 @@ async def _find_all_collections(self):
...
+                return True
+
+        return False
+
     async def run(self):
         """
         Build and emit `DeclarativeContent` from the ansible metadata.
         """
+        if self.repository.remote and self.remote.pk == self.repository.remote.pk:
+            can_be_optimized = await self._should_we_sync()
```
bmbouter, line 700:
```suggestion
            should_we_sync = await self._should_we_sync()
```
This name aligns with the True/False value it's receiving.

fao89:
PR already got unreadable with all the comments,
closing in favor of https://github.com/pulp/pulp_ansible/pull/521

PR #512, opened 2021-01-29
Efficient sync with unpaginated metadata endpoints if they are available

fao89:
https://pulp.plan.io/issues/8177
closes #8177

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -307,14 +307,16 @@ class CollectionSyncFirstStage(Stage):
         self.remote = remote
         self.collection_info = parse_collections_requirements_file(remote.requirements_file)
         self.deprecations = Q()
+        self._unpaginated_collection_metadata = None
+        self._unpaginated_collection_version_metadata = None
 
         # Interpret download policy
         self.deferred_download = self.remote.policy != Remote.IMMEDIATE
 
     @alru_cache(maxsize=128)
-    async def _get_collection_api(self, root):
+    async def _get_root_api(self, root):
         """
-        Returns the collection api path and api version.
+        Returns the root api path and api version.
 
         Based on https://git.io/JTMxE.
         """
@@ -343,14 +345,27 @@ class CollectionSyncFirstStage(Stage):
         else:
             raise RuntimeError(_("Unsupported API versions at {}").format(root))
 
-        endpoint = f"{root}v{api_version}/collections/"
+        endpoint = f"{root}v{api_version}"
 
         return endpoint, api_version
+
+    @alru_cache(maxsize=128)
+    async def _get_paginated_collection_api(self, root):
+        """
+        Returns the collection api path and api version.
+
+        Based on https://git.io/JTMxE.
+        """
+        endpoint, api_version = await self._get_root_api(root)
+        return f"{endpoint}/collections/", api_version
 
     async def _fetch_collection_version_metadata(self, api_version, collection_version_url):
         downloader = self.remote.get_downloader(url=collection_version_url)
         metadata = parse_metadata(await downloader.run())
+        await self._add_collection_version(api_version, collection_version_url, metadata)
 
+    async def _add_collection_version(self, api_version, collection_version_url, metadata):
+        """Add CollectionVersion to the sync pipeline."""
         url = metadata["download_url"]
 
         collection_version = CollectionVersion(
@@ -400,17 +415,9 @@ class CollectionSyncFirstStage(Stage):
             versions_list_url = f"{url_without_get_params}?limit={page_size}&offset={offset}"
         return self.remote.get_downloader(url=versions_list_url)
 
-    async def _fetch_collection_metadata(self, requirements_entry):
-        if requirements_entry.version == "*":
-            requirement_version = Requirement.parse("collection")
-        else:
-            requirement_version = Requirement.parse(f"collection{requirements_entry.version}")
-        if requirements_entry.source:
-            root = requirements_entry.source
-        else:
-            root = self.remote.url
-        collection_endpoint, api_version = await self._get_collection_api(root)
-        namespace, name = requirements_entry.name.split(".")
+    async def _fetch_paginated_collection_metadata(self, name, namespace, requirement, source=None):
+        root = source or self.remote.url
+        collection_endpoint, api_version = await self._get_paginated_collection_api(root)
         collection_url = f"{collection_endpoint}{namespace}/{name}"
         collection_metadata_downloader = self.remote.get_downloader(url=collection_url)
         collection_metadata = parse_metadata(await collection_metadata_downloader.run())
@@ -428,7 +435,7 @@ class CollectionSyncFirstStage(Stage):
             else:
                 collection_versions = collection_versions_list["data"]
             for collection_version in collection_versions:
-                if collection_version["version"] in requirement_version:
+                if collection_version["version"] in requirement:
                     version_num = collection_version["version"]
                     collection_version_detail_url = f"{collection_url}/versions/{version_num}/"
                     if collection_metadata["deprecated"]:
@@ -448,6 +455,43 @@ class CollectionSyncFirstStage(Stage):
 
         await asyncio.gather(*tasks)
 
+    async def _read_from_downloaded_metadata(self, name, namespace, requirement):
+        tasks = []
+        loop = asyncio.get_event_loop()
+
+        collection_metadata = self._unpaginated_collection_metadata[namespace][name]
+        for collection_data in collection_metadata:
+            if collection_data["deprecated"]:
+                self.deprecations |= Q(namespace=namespace, name=name)
+
+        collection_version_metadata = self._unpaginated_collection_version_metadata[namespace][name]
+
+        for col_version_metadata in collection_version_metadata:
+            if col_version_metadata["version"] in requirement:
+                collection_version_url = urljoin(self.remote.url, f"{col_version_metadata['href']}")
+                tasks.append(
+                    loop.create_task(
+                        self._add_collection_version(
+                            self._api_version, collection_version_url, col_version_metadata
+                        )
+                    )
+                )
+
+    async def _fetch_collection_metadata(self, requirements_entry):
+        if requirements_entry.version == "*":
+            requirement_version = Requirement.parse("collection")
+        else:
+            requirement_version = Requirement.parse(f"collection{requirements_entry.version}")
+
+        namespace, name = requirements_entry.name.split(".")
+
+        if self._unpaginated_collection_version_metadata and requirements_entry.source is None:
+            await self._read_from_downloaded_metadata(name, namespace, requirement_version)
+        else:
+            await self._fetch_paginated_collection_metadata(
+                name, namespace, requirement_version, requirements_entry.source
+            )
+
     @staticmethod
     def _get_response_next_value(api_version, response):
         if api_version == 2:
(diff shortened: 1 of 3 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -460,8 +473,52 @@ def _collection_list_downloader(self, api_version, collection_endpoint, page_num
...
+                    )
+
+                await asyncio.gather(*tasks)
+                return True
+
+        return False
+
     async def _find_all_collections(self):
-        collection_endpoint, api_version = await self._get_collection_api(self.remote.url)
+        found = await self._find_all_unpaginated_collections()
```
bmbouter, line 517: The issue with calling this new workflow from here is that it doesn't trigger when using a requirements file. I think the new metadata format needs to be used when using a requirements file too whenever its available.

I was testing this with `pytest -v -r sx --color=yes --pyargs pulp_ansible.tests.functional.api.collection.v3.test_sync::SyncCollectionsFromPulpServerTestCase::test_sync_collections_from_pulp`

fao89: I intentionally added it here, why would I want a huge blob of collections when I have a little requirements file?

bmbouter: I think of a few reasons.
* We need to cut down on the number of requests becasue it triggers rate limiting
* If we configure `rate_limit` making the 3*N calls for the content in the requirements file will likely take much longer than requesting 1 larger response once
* Consider requirements files with many collections in them, like hundreds.

fao89: I agree with all your reasons, but take our tests,
we generally do:
```
"collections:\n  - testing.k8s_demo_collection"
```
Let's say we have a repo with 30000 collections and this one is in the middle, I don't think it would be good to loop many collections to get just a few.

Maybe we can adopt different strategies depending on `len(self.collection_info)`

bmbouter: I can see your point, but consider the actual execution times here. This isn't exactly an apples for apples because there is json processing time and more comparisons, etc, but here's a little script I wrote that loops over 30,000 tuples checking each one:

```
def test():
    """Test function"""
    for item in data:
        if item[1] == 'yes':
            var = 1

data = [('a', 'yes') * 30000]

if __name__ == '__main__':
    import timeit
    print(timeit.timeit("test()", setup="from __main__ import test, data"))
```

When I run this it takes `0.11673803627490997` seconds.

I was then pinging galaxy to show it's network round trip time by comparison:

```
[bmbouter@localhost tmp]$ ping galaxy.ansible.com
PING galaxy.ansible.com (104.26.0.234) 56(84) bytes of data.
64 bytes from 104.26.0.234 (104.26.0.234): icmp_seq=1 ttl=55 time=25.0 ms
64 bytes from 104.26.0.234 (104.26.0.234): icmp_seq=2 ttl=55 time=31.6 ms
64 bytes from 104.26.0.234 (104.26.0.234): icmp_seq=3 ttl=55 time=36.9 ms
64 bytes from 104.26.0.234 (104.26.0.234): icmp_seq=4 ttl=55 time=30.0 ms
64 bytes from 104.26.0.234 (104.26.0.234): icmp_seq=5 ttl=55 time=27.7 ms
64 bytes from 104.26.0.234 (104.26.0.234): icmp_seq=6 ttl=55 time=27.6 ms
```

Say we say it takes 27ms which is .027. seconds. So if we even make 4 requests looping over 30K items would still be faster.

This isn't a great experiement for a lot of reasons but my point is that fetching a single blob of data that is large, but only having to go to the server once is very efficient.

bmbouter:
I was thinking about how to organize this code and here's what I came up with:

Regardless of if it's doing a full mirror or using a requirements file, have it perform the fetching and parsing of the metadata early on and store the data. Then have the "collection fetching" code (either full mirror or requirements file based) try to read that data first, and if its present use it, and if not fall back to the original implementation.

fao89:
> I was thinking about how to organize this code and here's what I came up with:
> 
> Regardless of if it's doing a full mirror or using a requirements file, have it perform the fetching and parsing of the metadata early on and store the data. Then have the "collection fetching" code (either full mirror or requirements file based) try to read that data first, and if its present use it, and if not fall back to the original implementation.

do you mean storing in memory? Or start something at db level?

```python
class NewModel
    published = DateTime()
    url = CharField()
    blob = JsonField()
```

bmbouter:
> > I was thinking about how to organize this code and here's what I came up with:
> > Regardless of if it's doing a full mirror or using a requirements file, have it perform the fetching and parsing of the metadata early on and store the data. Then have the "collection fetching" code (either full mirror or requirements file based) try to read that data first, and if its present use it, and if not fall back to the original implementation.
> 
> do you mean storing in memory? Or start something at db level?
> 
> ```python
> class NewModel
>     published = DateTime()
>     url = CharField()
>     blob = JsonField()
> ```

I mean in memory. Maybe something like:

namespaces = ['namespace']['name']
collection_versions = ['namespace']['name']['version']

pulp_ansible/app/tasks/collections.py
```diff
@@ -460,8 +473,59 @@ def _collection_list_downloader(self, api_version, collection_endpoint, page_num
             collection_list_url = f"{collection_endpoint}?limit={page_size}&offset={offset}"
         return self.remote.get_downloader(url=collection_list_url)
 
+    async def _download_unpaginated_repo_metadata(self):
+        self._unpaginated_collection_metadata = None
+        self._unpaginated_collection_version_metadata = None
+
+        root_endpoint, api_version = await self._get_root_api(self.remote.url)
+        self._api_version = api_version
+        collection_endpoint = f"{root_endpoint}/metadata/collections/?deprecated=true"
```
alikins, line 482: Is there any reason the /metadata/ is needed?  This seems like info that would be fitting for Collection and CollectionVersion endpoints?

alikins (changes requested):
Feel free to override of course, but I would really like to avoid adding a new 'metadata' if at all possible.

pulp_ansible/app/tasks/collections.py
fao89: the metadata endpoints are unpaginated, the main difference is getting all metadata with just one request instead of doing a lot of requests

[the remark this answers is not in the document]
fao89: we can, do you have any suggestion? Naming is always hard for me

bmbouter: @alikins yeah the issue is that the existing endpoints are paginated and also for the CollectionVersion detail view it doesn't contain all of the info needed when the responses come in one large response. Also these endpoints are already merged; this code actually just uses them for syncing. So if we're changing the design that's ok but that wouldn't be in this PR.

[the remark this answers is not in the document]
bmbouter: @fao89 for this I think we don't want server-side filtering, we want all the data. For right now it may not make a difference because really we just need this endpoint to get the `deprecated=true` but over time we probably want the fetching of this endpoint to respond with all data, not filtered data.

pulp_ansible/app/tasks/collections.py
```diff
@@ -463,8 +511,59 @@ def _collection_list_downloader(self, api_version, collection_endpoint, page_num
...
+                pass
+            else:
+                self._unpaginated_collection_metadata = metadata
+                collection_version_endpoint = f"{root_endpoint}/metadata/collection_versions/"
+                downloader = self.remote.get_downloader(url=collection_version_endpoint)
+                self._unpaginated_collection_version_metadata = parse_metadata(
+                    await downloader.run()
+                )
+
+    async def _find_all_unpaginated_collections(self):
```
bmbouter, line 537:
```suggestion
    async def _find_all_collections_from_unpaginated_data(self):
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -463,8 +511,59 @@ def _collection_list_downloader(self, api_version, collection_endpoint, page_num
             collection_list_url = f"{collection_endpoint}?limit={page_size}&offset={offset}"
         return self.remote.get_downloader(url=collection_list_url)
 
+    async def _download_unpaginated_repo_metadata(self):
+        self._unpaginated_collection_metadata = None
+        self._unpaginated_collection_version_metadata = None
```
bmbouter, line 516: These should probably move to the init since code in other functions later checks them.

pulp_ansible/app/tasks/collections.py
```diff
@@ -463,8 +511,59 @@ def _collection_list_downloader(self, api_version, collection_endpoint, page_num
...
+
+        root_endpoint, api_version = await self._get_root_api(self.remote.url)
+        self._api_version = api_version
+        collection_endpoint = f"{root_endpoint}/metadata/collections/"
```
bmbouter, line 520: This line can move inside the `if` statement.

pulp_ansible/app/tasks/collections.py
```diff
@@ -463,8 +511,59 @@ def _collection_list_downloader(self, api_version, collection_endpoint, page_num
             collection_list_url = f"{collection_endpoint}?limit={page_size}&offset={offset}"
         return self.remote.get_downloader(url=collection_list_url)
 
+    async def _download_unpaginated_repo_metadata(self):
```
bmbouter, line 514:
```suggestion
    async def _download_unpaginated_metadata(self):
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -448,6 +453,49 @@ async def _fetch_collection_metadata(self, requirements_entry):
 
         await asyncio.gather(*tasks)
 
+    async def _fetch_unpaginated_collection_metadata(self, name, namespace, requirement):
```
bmbouter, line 456:
```suggestion
    async def _read_from_downloaded_metadata(self, name, namespace, requirement):
```

bmbouter: I'm proposing this because no actual fetching (aka downloading) is being done here.

pulp_ansible/app/tasks/collections.py
```diff
@@ -448,6 +453,49 @@ async def _fetch_collection_metadata(self, requirements_entry):
...
+        tasks = []
+        loop = asyncio.get_event_loop()
+
+        collection_metadata = filter(
+            lambda data: data["name"] == name and data["namespace"] == namespace,
+            self._unpaginated_collection_metadata,
```
bmbouter, line 462: Algorithmically I think we should consider doing this differently. For a requirements file with N entries when there are M items to search through this results in O(N*M) performance.

Instead I think the downloading function should process the data before assigning it to `_unpaginated_collection_metadata` and `_unpaginated_collection_version_metadata`. It should organize the dictionaries with the top level keys being namespace and the second level key being collection_name.

Then these lookups can become just key/value lookups which would require just O(N + c) processing which is equivalent to O(N).

pulp_ansible/app/tasks/collections.py
```diff
@@ -448,6 +455,43 @@ async def _fetch_collection_metadata(self, requirements_entry):
 
         await asyncio.gather(*tasks)
 
+    async def _read_from_downloaded_metadata(self, name, namespace, requirement):
+        tasks = []
+        loop = asyncio.get_event_loop()
+
+        collection_metadata = self._unpaginated_collection_metadata[namespace][name]
```
bmbouter, line 462: Won't this key error if the requirements file specifies a collection that doesn't exist?

pulp_ansible/app/tasks/collections.py
```diff
@@ -448,6 +455,43 @@ async def _fetch_collection_metadata(self, requirements_entry):
...
+        for collection_data in collection_metadata:
+            if collection_data["deprecated"]:
+                self.deprecations |= Q(namespace=namespace, name=name)
+
+        collection_version_metadata = self._unpaginated_collection_version_metadata[namespace][name]
```
bmbouter, line 467: Similar here.

pulp_ansible/app/tasks/collections.py
```diff
@@ -463,8 +507,76 @@ def _collection_list_downloader(self, api_version, collection_endpoint, page_num
...
+                        self._unpaginated_collection_metadata["deprecated"].append(
+                            {"namespace": namespace, "name": name}
+                        )
+                    self._unpaginated_collection_metadata[namespace] = {name: [data]}
+                collection_version_endpoint = f"{root_endpoint}/metadata/collection_versions/"
+                downloader = self.remote.get_downloader(url=collection_version_endpoint)
+                self._unpaginated_collection_version_metadata = {
+                    "raw": parse_metadata(await downloader.run())
+                }
+                for data in self._unpaginated_collection_version_metadata["raw"]:
```
bmbouter, line 537: I didn't imagine us storing two copies of the metadata in memory.

bmbouter (approved):
This gets us very close to where I think we need to be with this feature. I'd be willing to merge this and I can open up some of the last more subjective changes in a followup PR. @fao89 wdyt?

fao89:
> This gets us very close to where I think we need to be with this feature. I'd be willing to merge this and I can open up some of the last more subjective changes in a followup PR. @fao89 wdyt?

I agree

PR #516, opened 2021-02-08
Adds test for metadata sync with no reqs file

bmbouter:
This reworks the metadata storage some to use defaultdict which also
deduplicates the storage of the data.

It adds a test which will sync data into Pulp and then mirror data from
that repo. The mirror is meaningful because a different codepath is used
to "discover" all collections and collection_versions using the sync'd
metadata.

[noissue]


fao89 (approved):
thank you!

bmbouter:
Thanks @fao89 !

PR #518, opened 2021-02-09
Adds the ``execution_environment`` attribute to the Galaxy V3 CollectionVersion APIs

bmclaughlin:
https://pulp.plan.io/issues/8233
closes #8233


pulp_ansible/app/tasks/collections.py
```diff
@@ -717,6 +717,9 @@ async def _post_save(self, batch):
             docs_blob = d_content.extra_data.get("docs_blob", {})
             if docs_blob:
                 collection_version.docs_blob = docs_blob
+            execution_environment = d_content.extra_data.get("execution_environment", {})
```
fao89, line 720: `execution_environment` needs to be set on `d_content.extra_data`
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/tasks/collections.py#L397-L405

fao89 (commented):
should this be on pulp-ansible 0.7? Or can it go on 0.8?

bmclaughlin:
This should be on pulp-ansible 0.7 if possible.

pulp_ansible/app/tasks/collections.py
bmclaughlin: Thanks, thought I had removed this.  Seems ctrl-z wasn't my friend earlier today.

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
```diff
@@ -224,6 +224,7 @@ def test_collection_version(artifact, pulp_client, collection_detail):
     assert "files" in version
     assert "manifest" in version
     assert "requires_ansible" in version
+    assert "execution_environment" in version
```
bmbouter, line 227: This is good, but we also need two more tests I think:

1) Upload a collection and assert that the API renders the data correctly
2) Sync in a collection and assert that the API renders the data correctly

bmbouter: For example here's a similar test added for uploading and then asserting on the `requires_ansible` field https://github.com/pulp/pulp_ansible/pull/466/files#diff-0ac20bc70d45017ec41ec2a189688b5814a44a91b0c55df22e03a3c9fd5c7b74R77

bmbouter: Here is one that does syncing and then asserting, also for the `requires_ansible` field: https://github.com/pulp/pulp_ansible/pull/466/files#diff-33689e25ac3c04bf435e438f2d0c86bbf6e21b4cc1524eeaf1e65cd33e8939f3R72-R91

bmbouter: Actually before making these tests, let me ask. How does data get into this field? In pulp_ansible it comes either from the artifact uploaded through the CLI or from sync. It can't come from an API call though.

bmclaughlin: I looked at adding both of those tests, but the rendered data would be `None` in either case (as the data is populated in Galaxy Importer by combining multiple files into said data).  I'm not sure how helpful that data point would be in tests?

I was leaning towards adding a Todo for myself to add additional tests (probably syncing in a collection) when execution environment data would be more widely available to test against.  Do you have a preference?

bmbouter: The galaxy importer's data is returned here https://github.com/pulp/pulp_ansible/blob/8dd22a1ba89dfe2a11164d0272fc3e9fea67d3eb/pulp_ansible/app/tasks/collections.py#L153-L155 so that needs to be extended to populate the data. That would bring the data in during import. The unit test could then assert on that.

Then since the data is available on the server side (via the galaxy API V3) the Pulp sync code needs to bring it in to sync's clients. So the test we use there is upload data into repo a, make a distribution serving repo a, make repo b, make a distribution serving repo b, sync repo a-> b, and assert on the data now available in repo b. We have pulpToPulpV3 sync tests already so it's basically just doing that only with an upload at the beginning.

pulp_ansible/app/migrations/0031_collectionversion_execution_environment.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.18 on 2021-02-08 20:36
+
+import django.contrib.postgres.fields.jsonb
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ansible', '0030_collectionversion_requires_ansible'),
```
fao89, line 10:
```suggestion
        ('ansible', '0031_ansiblerepository_last_synced_metadata_time'),
```

fao89: and please rename the migration to: 
```
0032_collectionversion_execution_environment.py 
```

pulp_ansible/app/migrations/0032_collectionversion_execution_environment.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.18 on 2021-02-08 20:36
+
+import django.contrib.postgres.fields.jsonb
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ansible', '0030_collectionversion_execution_environment'),
```
fao89, line 10: it depends on the 31
```suggestion
        ('ansible', '0031_ansiblerepository_last_synced_metadata_time'),
```

bmclaughlin: :facepalm:

bmclaughlin:
Closing due to feature no longer being needed.

PR #519, opened 2021-02-09
Fix AttributeError when CONTENT_ORIGIN is undefined

daviddavis:
fixes #8235

### DIFF
--- a/pulp_ansible/app/settings.py
+++ b/pulp_ansible/app/settings.py
@@ -21,4 +21,4 @@ LOGGING = {
 }
 
 ANSIBLE_API_HOSTNAME = "http://" + socket.getfqdn()
-ANSIBLE_CONTENT_HOSTNAME = settings.CONTENT_ORIGIN + "/pulp/content"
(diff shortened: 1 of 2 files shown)


daviddavis:
@bmbouter does this fix makes sense or is there a better way to solve this?

pulp_ansible/app/settings.py
```diff
@@ -21,4 +21,4 @@
 }
 
 ANSIBLE_API_HOSTNAME = "http://" + socket.getfqdn()
-ANSIBLE_CONTENT_HOSTNAME = settings.CONTENT_ORIGIN + "/pulp/content"
+ANSIBLE_CONTENT_HOSTNAME = getattr(settings, "CONTENT_ORIGIN", "") + "/pulp/content"
```
fao89, line 24:
```suggestion
ANSIBLE_CONTENT_HOSTNAME = getattr(settings, "CONTENT_ORIGIN", "http://" + socket.getfqdn()) + "/pulp/content"
```

bmbouter: I don't think we can do this because we can't assume the http protocol.

daviddavis:
Closing in favor of https://github.com/pulp/pulpcore/pull/1119.

PR #521, opened 2021-02-10
Improve sync performance with no-op when possible

fao89:
https://pulp.plan.io/issues/7942
closes #7942


pulp_ansible/app/tasks/collections.py
```diff
@@ -608,13 +661,60 @@ async def _find_all_collections(self):
...
+            if self.optimize and not should_we_sync:
+                noop.message = _(
+                    "no-op: {remote} did not change since last sync - {published}".format(
+                        remote=self.remote.url, published=self.last_synced_metadata_time
+                    )
+                )
+                noop.save()
+                log.debug(_("no-op: remote wasn't updated since last sync."))
+                return
+            noop.save()
```
bmbouter, line 711: What do you think about moving this into the `_should_we_sync` and then have this be:

```
if self.optimize:
    should_we_sync = await self._should_we_sync():
    if should_we_sync is False:
        return
    
```

The reasoning is to keep the concerns of the no-op in the area where the no-op happens.

CHANGES/7942.feature
```diff
@@ -0,0 +1,2 @@
+Improve sync performance with no-op when possible.
+To disable the no-op optimization use the ``optimize=False`` option on the ``sync`` call.
```
bmbouter, line 2: I try to have the line wraps in the release notes occur at the bullet points. To do that, can these two lines be grouped like:

```
Improve sync performance with no-op when possible. To disable the no-op optimization use the
``optimize=False`` option on the ``sync`` call.
```

bmbouter (approved):
Thank you so much @fao89. I've asked a lot, lot of changes on this PR and I really appreciate you going through this with me. You've done a great job. :100:

PR #525, opened 2021-02-18
Update plugin template

mdellweg:
This accommodates for a change in pulpcore that will effect the CI.

https://github.com/pulp/pulpcore/pull/1102

[noissue]


bmbouter:
This looks good to me. We need this to unblock an upcoming change in pulpcore pulp/pulpcore#1102

bmbouter (approved):
This looks good to me. We need this to unblock an upcoming change in pulpcore pulp/pulpcore#1102

PR #526, opened 2021-02-23
Fix broken bindings filters

fao89:
[noissue

### DIFF
--- a/CHANGES/8300.bugfix
+++ b/CHANGES/8300.bugfix
@@ -1,0 +1,1 @@
+Fixed bug where rate limit wasn't being honored.
--- a/pulp_ansible/app/downloaders.py
+++ b/pulp_ansible/app/downloaders.py
@@ -212,4 +212,6 @@ class AnsibleDownloaderFactory(DownloaderFactory):
         if not self._remote.token and self._remote.username and self._remote.password:
             options["auth"] = BasicAuth(login=self._remote.username, password=self._remote.password)
 
+        kwargs["throttler"] = self._remote.download_throttler if self._remote.rate_limit else None
+
         return download_class(url, self._remote.auth_url, self._remote.token, **options, **kwargs)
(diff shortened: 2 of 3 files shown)


pulp_ansible/app/downloaders.py
```diff
@@ -212,4 +212,6 @@ class to be instantiated.
         if not self._remote.token and self._remote.username and self._remote.password:
             options["auth"] = BasicAuth(login=self._remote.username, password=self._remote.password)
 
+        kwargs["throttler"] = self._remote.download_throttler if self._remote.rate_limit else None
```
fao89, line 215: we were not setting it

CHANGES/8300.bugfix
```diff
@@ -0,0 +1 @@
+Fix rate limit feature
```
daviddavis, line 1: Can this be more specific? Maybe something like "Fixed bug where rate limiting wasn't being used."

daviddavis:
Does the DRF issue manifest to users who are using 0.7.0? Or will it only be an issue when pulpcore 3.11 is released?

fao89:
> Does the DRF issue manifest to users who are using 0.7.0? Or will it only be an issue when pulpcore 3.11 is released?

second option, only when 3.11 is released

pulp_ansible/app/downloaders.py
bmbouter: This is good. :+1:

pulp_ansible/tests/functional/utils.py
```diff
@@ -63,7 +63,7 @@ def gen_ansible_remote(url=ANSIBLE_FIXTURE_URL, **kwargs):
     :param url: The URL of an external content source.
     """
     if "rate_limit" not in kwargs:
-        kwargs["rate_limit"] = 10
+        kwargs["rate_limit"] = 5
```
bmbouter, line 66: Did we have problems with it at 10? I'm asking because if we can make the tests go twice as fast that would be good.

fao89: that was my first attempt before I realize the rate_limit was not being set,
I'll keep it 10

daviddavis (approved):

bmbouter (approved):
Looks great, thanks for fixing this!

PR #533, opened 2021-03-02
Use DRF token when no auth_url is provided

fao89:
https://pulp.plan.io/issues/8290
closes #8290


bmbouter (approved):
This looks great. :+1:

PR #545, opened 2021-03-12
Added full syncing of collection dependencies

gerrod3:
fixes: #7751
https://pulp.plan.io/issues/7751

Still working on this a bit, I have some questions about the behavior and usage we want for this feature.
First, should this be specified as an option on the remote or as another parameter to the sync operation? I went with adding it to the remote since I felt that made sense, but either way I think I still need to add logic to update the last sync parameter when this field is updated. 

Second, do you know of collections that have multiple dependencies and collections with dependencies with dependencies? I just used the pulp_installer collection as a simple test case, but I would like to add others to ensure full functionality.  Also, what should we do about versioning mismatches? I think I still need to add better logic for checking if the dependency collection is already in the list of collections to sync(ed), but if you list a dependency with a certain version and that version is too low for the actual dependency should the sync go ahead and sync the right one or just let it error later when using the galaxy cli.

Also I've never tried recursion with async code, so if it's really bad, then I guess I could try another way.


fao89 (commented):
We have a test that was breaking due to a dependency issue:
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/cli/test_collection_install.py#L62
As workaround, we added `--no-deps`, with this PR we can remove it

pulp_ansible/app/serializers.py
```diff
@@ -134,6 +134,12 @@ class CollectionRemoteSerializer(RemoteSerializer):
         max_length=2000,
     )
 
+    sync_dependencies = serializers.BooleanField(
+        help_text=_("Sync dependencies for collections specified in requirements file"),
+        allow_null=True,
+        default=False,
+    )
```
fao89, line 141: we support mirroring and sync from requirements file, I believe when syncing from requirements it would always sync dependencies

pulp_ansible/app/tasks/collections.py
```diff
@@ -707,6 +725,24 @@ async def _should_we_sync(self):
...
+        while True:
+            try:
+                requirement_entry = self.collection_queue.get_nowait()
+                tasks.append(loop.create_task(self._fetch_collection_metadata(requirement_entry)))
+            except asyncio.QueueEmpty:
+                break
+
+        await asyncio.gather(*tasks)
+        if not self.collection_queue.empty():
+            await self.full_dependency_sync()
```
fao89, line 744: cool!

pulp_ansible/app/tasks/collections.py
```diff
@@ -431,6 +437,18 @@ async def _add_collection_version(self, api_version, collection_version_url, met
...
+            dependencies = info["dependencies"]
+            for name, version in dependencies.items():
+                requirements_file = RequirementsFileEntry(
+                    name=name,
+                    version=version,
+                    source=None,
+                )
+                if requirements_file not in self.collection_info:
+                    self.collection_queue.put_nowait(requirements_file)
+                    self.collection_info.append(requirements_file)
```
fao89, line 450: I think you also need to check if it is on the queue,
pulp-installer 3.10.0 requirement: ansible.posix
.
.
.
pulp-installer 3.6.0 requirement: ansible.posix

gerrod3: I don't think so. As far as I understand async code is still serial and .put_nowait() is not a coroutine, so these two lines should always get executed before another async task is ran.

fao89: I thought `put_nowait` was a coroutine

gerrod3: According to this, https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.put `put` is, but `put_nowait` is not.

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -59,7 +59,7 @@ def create_install_scenario(self, body, collection_name):
 
         with tempfile.TemporaryDirectory() as temp_dir:
             # FIXME Use --no-deps while https://pulp.plan.io/issues/7751 is not addressed
```
bmbouter, line 61: This FIXME can also be removed

pulp_ansible/tests/functional/utils.py
```diff
@@ -63,7 +63,7 @@ def gen_ansible_remote(url=ANSIBLE_FIXTURE_URL, **kwargs):
     :param url: The URL of an external content source.
     """
     if "rate_limit" not in kwargs:
-        kwargs["rate_limit"] = 10
+        kwargs["rate_limit"] = 5
```
bmbouter, line 66: Is this needed? I thought the rate_limit of 10 was enough for galaxy.ansible.com and c.rh.c syncs?

pulp_ansible/tests/functional/api/collection/test_sync.py
```diff
@@ -136,3 +150,42 @@ def test_sync_collection_with_stranger_version_numbers_to_check_comparisons(self
...
+        super().setUpClass()
+        cls.col_api = AnsibleCollectionsApi(cls.client)
+
+    def _clean_database(self):
+        """Cleans the database."""
+        cli_client = cli.Client(cfg)
+        cmd = get_psql_smash_cmd(TRUNCATE_TABLES_QUERY_BASH)
+        cli_client.run(cmd, sudo=True)
+
+    def test_simple_one_level_dependency(self):
```
bmbouter, line 170: For this I recommend manually, locally creating 3 test collections, A -> B -> C and trying to have the test sync A and asserting A, B, and C are present. I think it'll give more control to use manufactured test data. The test would upload A, B, and C to Pulp, expose via distribution, create a second repo, and then sync A to the second repo (so the sync occurs from pulp). There are other tests that do this so that can provide most of the example code.

pulp_ansible/app/tasks/collections.py
```diff
@@ -360,6 +360,12 @@ def __init__(self, remote, repository, optimize):
         self.optimize = optimize
         self.collection_info = parse_collections_requirements_file(remote.requirements_file)
         self.deprecations = Q()
+        self.add_dependents = False
+        if self.collection_info and self.remote.sync_dependencies:
+            self.add_dependents = True
+            self.collection_queue = asyncio.Queue()
```
bmbouter, line 366: I may be missing something here, but I imagined we didn't need a dedicated codepath to handle this case differently from our existing sync code. Can't we take new requirements and spawn new coroutines to handle them as if they came from the original requirements file?

pulp_ansible/tests/functional/utils.py
fao89: this PR was constantly failing before this change

pulp_ansible/tests/functional/api/collection/test_sync.py
fao89: maybe we can use: https://github.com/peaqe/orion-utils
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/v3/test_collection.py#L25

pulp_ansible/tests/functional/api/collection/test_sync.py
```diff
@@ -136,3 +150,42 @@ def test_sync_collection_with_stranger_version_numbers_to_check_comparisons(self
...
+class FullDependenciesSync(TestCaseUsingBindings, SyncHelpersMixin):
+    """Collection sync tests for syncing collections and their dependencies."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Set up class variables."""
+        super().setUpClass()
+        cls.col_api = AnsibleCollectionsApi(cls.client)
+
+    def _clean_database(self):
```
bmbouter, line 164: At some point in the future we're going to try to have tests not require the db to be in a specific state. When that happens this type of reset will be problematic. I suspect we could remove this and have the sync into the new repo demonstrate the sync is pulling in dependencies when syncing from Pulp itself. @gerrod3 wdyt? Does this idea make sense?

gerrod3: Yeah, I'm in the middle of rewriting the test to use custom uploaded collections and I have removed the database cleanup code.

pulp_ansible/app/tasks/collections.py
```diff
@@ -529,6 +544,7 @@ async def _read_from_downloaded_metadata(self, name, namespace, requirement):
                         )
                     )
                 )
+        await asyncio.gather(*tasks)
```
gerrod3, line 547: The dependencies wouldn't be synced for the v3 codepath without this gather here. Not sure what other potential bugs were present before hand.

pulp_ansible/app/tasks/utils.py
```diff
@@ -43,7 +43,15 @@ def parse_metadata(download_result):
...
-RequirementsFileEntry = namedtuple("RequirementsFileEntry", ["name", "version", "source"])
+class RequirementsFileEntry(namedtuple("RequirementsFileEntry", ["name", "version", "source"])):
+    """Object to store collections from requirements file."""
+
+    def included(self, other):
+        """Simple inclusivity test, check if this RFE is part of another."""
+        if self.name == other.name:
+            if other.version == "*":
+                return True
+        return False
```
gerrod3, line 54: This should help reduce api requests for collections with shared dependencies when using paginated metadata endpoints.

pulp_ansible/app/tasks/collections.py
```diff
@@ -431,6 +430,22 @@ async def _add_collection_version(self, api_version, collection_version_url, met
...
+            loop = asyncio.get_event_loop()
+            for name in dependencies.keys():
+                # get everything to avoid adding potential duplicates to pipeline
+                new_req = RequirementsFileEntry(
+                    name=name,
+                    version="*",
+                    source=None,
+                )
+                if not any(map(lambda x: new_req.included(x), self.collection_info)):
+                    self.collection_info.append(new_req)
```
gerrod3, line 445: Based on our discussion @bmbouter @daviddavis, I decided to get all the versions of a dependency so that the galaxy CLI will have everything it needs to resolve dependencies, a "safe rather than sorry" approach.

CHANGES/7751.feature
```diff
@@ -0,0 +1 @@
+Pulp Ansible can now sync collection dependencies by setting remote option sync_dependencies true
```
bmbouter, line 1:
```suggestion
Pulp Ansible can now sync collection dependencies by setting the ``sync_dependencies`` option for ``CollectionRemote`` objects.
```

bmbouter: Also mentioning the default is `True` would be good.

pulp_ansible/app/serializers.py
```diff
@@ -134,6 +134,12 @@ class CollectionRemoteSerializer(RemoteSerializer):
         max_length=2000,
     )
 
+    sync_dependencies = serializers.BooleanField(
+        help_text=_("Sync dependencies for collections specified in requirements file"),
```
bmbouter, line 138:
```suggestion
        help_text=_("Sync dependencies for collections specified via requirements file"),
```

pulp_ansible/app/tasks/collections.py
bmbouter: If this is an unrelated bugfix can it receive it's own issue and separate commit please. That's important in cases where we say need to cherry pick back the bugfix by itself to an earlier branch.

pulp_ansible/tests/functional/api/collection/test_sync.py
```diff
@@ -136,3 +148,110 @@ def test_sync_collection_with_stranger_version_numbers_to_check_comparisons(self
...
+    Collection sync tests for syncing collections and their dependencies.
+
+    Dependency Trees:
+    A               E
+    |             /   \
+    B            F     G
+    |            |     |
+    C           H(1)   D
+    |
+    D
```
fao89, line 164: sweet!

pulp_ansible/tests/functional/constants.py
```diff
@@ -119,3 +119,18 @@
...
+    {"name": "d", "namespace": "test"},
+    {"name": "e", "namespace": "test", "dependencies": {"test.f": "*", "test.g": "*"}},
+    {"name": "f", "namespace": "test", "dependencies": {"test.h": "1.0.0"}},
+    {"name": "g", "namespace": "test", "dependencies": {"test.d": "*"}},
+    {"name": "h", "namespace": "test", "version": "1.0.0"},
+    {"name": "h", "namespace": "test", "version": "2.0.0"},
+    {"name": "h", "namespace": "test", "version": "3.0.0"},
+    {"name": "h", "namespace": "test", "version": "4.0.0"},
+    {"name": "h", "namespace": "test", "version": "5.0.0"},
+]
```
fao89, line 136: cool!

pulp_ansible/app/tasks/utils.py
```diff
@@ -43,7 +43,15 @@ def parse_metadata(download_result):
         return json.load(fd)
 
 
-RequirementsFileEntry = namedtuple("RequirementsFileEntry", ["name", "version", "source"])
+class RequirementsFileEntry(namedtuple("RequirementsFileEntry", ["name", "version", "source"])):
+    """Object to store collections from requirements file."""
+
+    def included(self, other):
```
fao89, line 49: never seen this before, I liked it!

fao89 (approved):
Looks good to me!
But I missed the meeting, @bmbouter and @daviddavis have more context than me

pulp_ansible/app/tasks/utils.py
```diff
@@ -43,7 +43,15 @@ def parse_metadata(download_result):
...
+        """Simple inclusivity test, check if this RFE is part of another."""
+        if self.name == other.name:
+            if other.version == "*":
```
bmbouter, line 52: It maybe doesn't matter but I figured an inequality test too? If the two versions were parsed with the semver library it could defer to their comparison for all cases other than the one written here. Just a though.

pulp_ansible/app/tasks/collections.py
```diff
@@ -431,6 +430,22 @@ async def _add_collection_version(self, api_version, collection_version_url, met
...
 
+        if self.add_dependents:
+            dependencies = info["dependencies"]
+            tasks = []
+            loop = asyncio.get_event_loop()
+            for name in dependencies.keys():
+                # get everything to avoid adding potential duplicates to pipeline
+                new_req = RequirementsFileEntry(
+                    name=name,
+                    version="*",
```
bmbouter, line 441: I think we want to use the version specifier declared by the dependency itself. If none is specified we should use `*` in that case.

bmbouter (changes requested):
I requested some changes. Most of them I think are straightforward. Let me know if any of them don't make sense or aren't right.

pulp_ansible/app/tasks/collections.py
gerrod3: Did you see my comment above? https://github.com/pulp/pulp_ansible/pull/545#discussion_r597069389 

There is an interesting scenario where say collection C has 10 versions and collection A has a dependency on collection C version <= 9.0 and collection B has a dependency on collection C any version.  If collection A is first in the requirements file it'll see the dependency and add the 9 versions of C to the sync pipeline.  Then B will be synced and it'll see the dependency and since it's any version it'll add all 10 versions of C to the pipeline even though 9 versions are already in it.  Pulp will sort this out later on, but the sync has gotten slower and more api requests where made then needed.  Now if there weren't shared dependencies across collections your suggestion would reduce the amount of CVs added to the pipeline, but I think it's hard to tell what case is more common.  I'm fine doing it either way, but I wanted to share my reasoning for having it like this.

pulp_ansible/app/tasks/utils.py
gerrod3: Not sure what you mean here.  We don't compare RFE's anywhere and I'm only using this function when determining to add a dependency to the sync pipeline. It has an expectation that it is being called with a new dependency, `self.version == "*"`, to check against other RFEs already in `collection_info`.

 So say the requirement file had collection A: "*", B: "*", C: "1.0" and A depended on C: "2.0" and B depended on C: "3.0".  `collection_info` would have `[(A, "*") , (B, "*"), (C, "1.0")]` and when it synced A it would see the dependency and call this method with `(C, "*")` The method would return false and all the other versions of C would be added to the sync.  When it syncs B `collection_info` will have `[(A, "*") , (B, "*"), (C, "1.0"), (C, "*")]` and the method will return true and thus the dependency won't be added to sync again.

pulp_ansible/app/tasks/collections.py
bmbouter: If that's how it's working in that example, I think we may do better having it deduplicate just before actually sending specific collection versions down the pipeline. I think this would happen in `_add_collection_version` metadata itself. Having a dictionary that stors the name, namespace, version triplet that was sync'd and ensuring that isn't already present I think would do it. What do you think about this alternative?

pulp_ansible/app/tasks/collections.py
```diff
@@ -422,15 +424,33 @@ async def _fetch_collection_version_metadata(self, api_version, collection_versi
...
-
         collection_version = CollectionVersion(
             namespace=metadata["namespace"]["name"],
             name=metadata["collection"]["name"],
             version=metadata["version"],
         )
+        cv_unique = attrgetter("namespace", "name", "version")(collection_version)
+        if cv_unique in self.already_synced:
+            return
+        self.already_synced.add(cv_unique)
```
gerrod3, line 435: @bmbouter Here is the dictionary (set) check for adding the CV to the pipeline. Tell me if I need to have this before the CollectionVersion object is created. I did it this way since I didn't like how long the tuple `(metadata["namespace"]["name"], metadata["collection"]["name"], metadata["version"])` was, but I can change it to that and remove the extra import I added if you want.

bmbouter: This is good how you've written it. +1 `attrgetter`

bmbouter: It also more DRY in case how the `namespace`, `name`, and `version` is read in from the metadata in the future.

pulp_ansible/tests/functional/constants.py
bmbouter: Yes!

bmbouter (approved):
Thank you @gerrod3. I'm really pleased with this implementation. Great job on the tests too. :+1:

daviddavis (approved):

PR #546, opened 2021-03-23
Ensure V3 async tasks complete

gerrod3:
fixes: #8442
https://pulp.plan.io/issues/8442


fao89 (approved):
LGTM!
From the issue, I understood it doesn't need to be backported, right?

gerrod3:
> LGTM!
> From the issue, I understood it doesn't need to be backported, right?

Yes, only the new dependency sync feature needs this as far as I know.

PR #550, opened 2021-03-31
Provide backend storage url to galaxy-importer on collection import

awcrosby:
Moves the setting of `ResponseContentDisposition` out of `galaxy-importer`, and allows for support of Azure backend as well as AWS, see [pulp issue 8486](https://pulp.plan.io/issues/8486).

This implementation replicates the logic found in [pulpcore _serve_content_artifact()](https://github.com/pulp/pulpcore/blob/8a77316d2d73dd31d6f47b964bf148a602da6a21/pulpcore/content/handler.py#L581), unsure if there is a good way to share this logic between pulpcore and a plugin.

fixes: #8486
https://pulp.plan.io/issues/8486

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -10,6 +10,7 @@ from urllib.parse import urljoin
 
 from aiohttp.client_exceptions import ClientResponseError
 from async_lru import alru_cache
+from django.conf import settings
 from django.db import transaction
 from django.db.models import Q
 from django.urls import reverse
@@ -199,8 +200,9 @@ def import_collection(
 
     try:
         with temp_file.file.open() as artifact_file:
+            url = _get_backend_storage_url(artifact_file)
             importer_result = process_collection(
-                artifact_file, filename=filename, logger=user_facing_logger
+                artifact_file, filename=filename, file_url=url, logger=user_facing_logger
             )
             artifact = Artifact.from_pulp_temporary_file(temp_file)
             importer_result["artifact_url"] = reverse("artifacts-detail", args=[artifact.pk])
@@ -278,6 +280,23 @@ def create_collection_from_importer(importer_result):
 
         collection_version.save()  # Save the FK updates
     return collection_version
+
+
+def _get_backend_storage_url(artifact_file):
+    """Get artifact url from pulp backend storage."""
+    if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
+        url = None
+    elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
+        parameters = {"ResponseContentDisposition": "attachment;filename=archive.tar.gz"}
+        url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
+    elif settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage":
+        url = artifact_file.storage.url(artifact_file.name)
+    else:
+        raise NotImplementedError(
+            f"The value settings.DEFAULT_FILE_STORAGE={settings.DEFAULT_FILE_STORAGE} "
+            "was not expected"
+        )
+    return url
 
 
 def _update_highest_version(collection_version):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 async_lru
-galaxy_importer
+galaxy_importer>=0.3.1
 jsonschema>=3.0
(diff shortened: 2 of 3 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -280,6 +282,20 @@ def create_collection_from_importer(importer_result):
     return collection_version
 
 
+def _get_backend_storage_url(artifact_file):
+    """Get artifact url from pulp backend storage."""
+    if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
+        url = None
+    elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
+        parameters = {"ResponseContentDisposition": "attachment;filename=archive.tar.gz"}
```
awcrosby, line 290: uses `"archive.tar.gz"` rather than `content_artifact.relative_path` since this only has artifact file from PulpTemporaryFile

pulp_ansible/app/tasks/collections.py
```diff
@@ -280,6 +282,20 @@ def create_collection_from_importer(importer_result):
     return collection_version
 
 
+def _get_backend_storage_url(artifact_file):
```
fao89, line 285: I feel pulpcore would be the right place for this, but we can have it on pulp_ansible by now

pulp_ansible/app/tasks/collections.py
```diff
@@ -280,6 +282,20 @@ def create_collection_from_importer(importer_result):
...
+    """Get artifact url from pulp backend storage."""
+    if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
+        url = None
+    elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
+        parameters = {"ResponseContentDisposition": "attachment;filename=archive.tar.gz"}
+        url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
+    elif settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage":
+        url = artifact_file.storage.url(artifact_file.name)
+    else:
+        raise NotImplementedError()
```
fao89, line 295: we need a more expressive exception

pulp_ansible/app/tasks/collections.py
fao89: generic name `"archive.tar.gz"` made me think about debugging/logging, maybe `f"{collection_name}.tar.gz"`

awcrosby:
> generic name "archive.tar.gz" made me think about debugging/logging, maybe f"{collection_name}.tar.gz"

not sure if that has been an issue to-date, but could be useful. if we did that, i assume we'd implement that separately than this pr

awcrosby:
@fao89 this can be tested with https://github.com/awcrosby/galaxy-importer/tree/feature/use_caller_file_url

pulp_ansible/app/tasks/collections.py
bmbouter: I agree. If we can also open it against pulpcore and keep it here so pulp_ansible doesn't require a pulpcore upgrade that would be ideal. @awcrosby can you also contribute this to pulpcore? @daviddavis I think can guide where/how to contribute it if he agrees.

fao89:
> @fao89 this can be tested with https://github.com/awcrosby/galaxy-importer/tree/feature/use_caller_file_url

Please rebase your PR and add the following to your commit message:
```
Required PR: https://github.com/ansible/galaxy-importer/pull/119
```

pulp_ansible/app/tasks/collections.py
awcrosby: Sounds good, I added a pulpcore issue to track this separately: https://pulp.plan.io/issues/8521

fao89 (commented):
Looks good to me, but I think we should pin galaxy-importer (>0.3), otherwise, it will break when passing the URL

https://github.com/pulp/pulp_ansible/blob/master/requirements.txt#L2

requirements.txt
```diff
@@ -1,5 +1,5 @@
 async_lru
-galaxy_importer
+galaxy_importer>=3.0
```
fao89, line 2:
```suggestion
galaxy_importer>=0.3
```

awcrosby: I went ahead and set it to the full version with the `file_url` parameter: `galaxy-importer==0.3.1`

fao89 (approved):
Thank you @awcrosby

PR #553, opened 2021-04-07
Added new AnsiblePublication to precompute metadata endpoints

gerrod3:
This PR is still very early WIP. I have only gotten the `/collection_versions/all/` endpoint to serve published content, but I've run into so many problems that I want to post this to start getting feedback.  

**1. How should I handle the published metadata requiring `base_path` to be created?**
For plugins with publications the workflow usually goes: create repository -> sync w/ remote -> create publication -> create distribution serving publication. The metadata we want to publish contains many hrefs that are dependent on the distribution's `base_path`.  The two solutions I could think of are 1: reverse the order / couple the creation of distributions and publications so that we know the `base_path` (what I did), or 2: save the published metadata in a templated form and have `base_path` be dynamically replaced depending on the distribution serving it (I tried this first and couldn't get it to work easily, so I would need some help).  Number 1 limits publications to being only viable to be served by one distribution and probably needs a lot of code to keep the coupling correct, but the data is fully pre-computed and servable by the content app. Number 2 allows publications to be served by multiple distributions and would be in-line with the behavior of other plugins' publications, but it does require some extra template computations at the endpoint, but hopefully not more than what is being done now.

**2. When and who should be creating this published metadata?**
This question is tightly linked to my first and third questions, but it boils down to how do we expect the user to use this feature?  Is this just an extra step they will now need to do or is it something that should just work immediately and have all the work be done in the background. Currently I have the publish work being done in the background when you create a distribution and when the repository gets updated, but no matter which way we go I think we'll need autopublish/distribute to make this easier, since what I'm doing right now is a poor man's copy of Daniel's work.  

**3. Do we want distributions to only be able to serve published content, aka remove the live update ability?**
If we switch to having distributions only serve publications we should also think of removing the galaxy collection upload endpoint or at least restricting it to only work if autopublishing is enabled.  The new distribution model allows us to do some pretty crazy things like having publications and repository-versions at the same time, but we probably shouldn't do this as it's not how it was intended.


rochacbruno:
@gerrod3 does it require pulpcore 3.12? or is it possible to have it working on 3.11 and create publication manually?

For Galaxy we could create the publication on the sync view.

gerrod3:
> @gerrod3 does it require pulpcore 3.12? or is it possible to have it working on 3.11 and create publication manually?
> 
> For Galaxy we could create the publication on the sync view.

@rochacbruno Currently this PR is using 3.12. The first commit was using 3.11 and it involved potentially adding a bunch of temporary code that would be replaced in 3.12.  It's still possible to have it on 3.11, but more migration code will need to be written for when you upgrade to 3.12.

newswangerd:
Before this gets merged, we have to verify a couple of things to make sure this scales well. For context, automation hub creates a repository for each customer, and each time a new collection version is uploaded, it gets copied into each customer's repo. At the moment we have 2000+ collections. This means that:

- The task to generate publication information has to be _fast_, because it needs to be able to compute 2000 publications each time a collection version is uploaded
- Publication artifacts have to be garbage collected and de duplicated. If we assume that the metadata for all collection versions is a modest 10mb, computing 2000 of them will consume 20gb of disk space.

This will also have to scale to meet galaxy.ansible.com. While we will only have a handful of repos there, galaxy has to be capable of handling tens of thousands of collection versions. There are some users who have their CI configured to push a new collection version with each commit, and galaxy needs to be able to quickly process lots of simultaneous uploads as well as all of the data that accompanies those uploads.

gerrod3:
>  The task to generate publication information has to be _fast_, because it needs to be able to compute 2000 publications each time a collection version is uploaded

This is not what Bruno told us, he said there was going to be a need for 5 publications.  Each distribution is going to need a publication. So if there is 5 distributions then 5 publications will be needed.  If each customer gets a distribution then there will be n publications created and updated each time a new collection version is uploaded to their repository.

>   Publication artifacts have to be garbage collected and de duplicated. If we assume that the metadata for all collection versions is a modest 10mb, computing 2000 of them will consume 20gb of disk space.

Currently we don't do garbage collection but since publications are tied to repository versions and you'll be using the upcoming keep N repository versions I would assume that would create an automatic garbage collection.  Published artifacts like all artifacts in Pulp will be de duplicated upon creation. So say there was 2000 customers we had to create publications for, but 1900 of them had the exact same contents in their repository then only 101 published artifacts would be created.

> This will also have to scale to meet galaxy.ansible.com. While we will only have a handful of repos there, galaxy has to be capable of handling tens of thousands of collection versions. There are some users who have their CI configured to push a new collection version with each commit, and galaxy needs to be able to quickly process lots of simultaneous uploads as well as all of the data that accompanies those uploads.

This is something that probably wouldn't be handled well.  Publication creation task locks on all publications and this operation currently calculates the metadata for every collection version present each time.  I'll have to think about how to handle this case.

rochacbruno:
@gerrod3 we have 5 distributions/remotes, those are fixed, user doesn't create new distributions but can create namespaces to upload collections to the staging/rejected/published distros and then we will have lots of repos under it. is that right @newswangerd ?

newswangerd:
> @gerrod3 we have 5 distributions/remotes, those are fixed, user doesn't create new distributions but can create namespaces to upload collections to the staging/rejected/published distros and then we will have lots of repos under it. is that right @newswangerd ?
> 
> 

We also have synclist repos. Those are the ones that are created on a per customer basis, but only on cloud.redhat.com.

There are repos created for each namespace, but those don't really have much content in them.

gerrod3:
Closing since no longer working on this.

PR #557, opened 2021-04-09
Fix bindings issue

fao89:
[noissue]


daviddavis:
Was this caused by the drf-spectacular upgrade to 0.15?

fao89:
> Was this caused by the drf-spectacular upgrade to 0.15?

yes, it is creating a fake request now

daviddavis (approved):

fao89:
> Can you add that info to the commit message?

not sure how to communicate that,
make pulp_ansible compatible with drf-spectacular 0.15 ?

daviddavis:
> make pulp_ansible compatible with drf-spectacular 0.15 ?

:+1:

PR #558, opened 2021-04-13
Replace Unpaginated endpoints with unified metadata for sync

rochacbruno:
- Remove Unpaginated /all/ endpoints
- Adds a Paginator with bigger limit
- Adds /collection/metadata/ endpoint with unified metadata for
  collection and its versions
- Changed sync to use this `optimized` downloader using 1000 limit per
  pagination

output example: https://gist.github.com/rochacbruno/df9c1c3e2653d1e9592cd8f650de2239

this is a temporary solution to be replaced by #553 which will pack the collections/metadata/ in a publication.
related to: https://github.com/ansible/galaxy_ng/pull/726

fixes: #8439

PR #561, opened 2021-04-15
Attempt to increase page_size limit for sync

rochacbruno:
[noissue]

### DIFF
--- a/pulp_ansible/app/galaxy/v3/pagination.py
+++ b/pulp_ansible/app/galaxy/v3/pagination.py
@@ -10,7 +10,7 @@ class LimitOffsetPagination(pagination.LimitOffsetPagination):
     """
 
     default_limit = 10
-    max_limit = 100
+    max_limit = 1000
 
     def get_first_link(self):
         """First link."""
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -45,7 +45,6 @@ from pulpcore.plugin.stages import (
 )
 from semantic_version import Version
 
-from pulp_ansible.app.constants import PAGE_SIZE
 from pulp_ansible.app.models import (
     AnsibleCollectionDeprecated,
     AnsibleRepository,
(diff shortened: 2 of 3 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -385,6 +384,7 @@ def __init__(self, remote, repository, optimize):
         self._unpaginated_collection_metadata = None
         self._unpaginated_collection_version_metadata = None
         self.last_synced_metadata_time = None
+        self.page_size = settings.get("SYNC_PAGE_SIZE", 1000)
```
gerrod3, line 387: Do you set this on galaxy's side?

gerrod3 (approved):
I think this PR should be used over #558. It's less invasive and is less likely to timeout. While it's not as performant I think we need more time to find a long term solution then trying to ship a last-minute temporary solution.

pulp_ansible/app/galaxy/v3/pagination.py
```diff
@@ -10,7 +10,7 @@ class LimitOffsetPagination(pagination.LimitOffsetPagination):
     """
 
     default_limit = 10
-    max_limit = 100
+    max_limit = 1000
```
alikins, line 13: Does this imply that OptimizedLimitOffsetPagination won't be used now? If so, presumably https://github.com/ansible/galaxy_ng/pull/726/files#diff-e43be673099bbec9ffef097ed5503589ff4597e387c89ed3ea2fdf9365e3d05dR81 needs updating as well

pulp_ansible/app/tasks/collections.py
rochacbruno: it defaults to 1000 but can be set to other value via PULP_SYNC_PAGE_SIZE variable

pulp_ansible/app/galaxy/v3/pagination.py
rochacbruno: if this PR is accepted then 726 can be closed without merging.

rochacbruno:
@gerrod3 lets close it and go with #562 ?

PR #562, opened 2021-04-16
/collection_versions/all/ endpoint is now streamed

gerrod3:
This is super fast I think. Potentially fast enough to rethink when or if we will need publications.

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -6,10 +6,12 @@ from django.contrib.postgres.aggregates import ArrayAgg
 from django.db.models import F
 from django.db.models.expressions import Window
 from django.db.models.functions.window import FirstValue
+from django.http import StreamingHttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils.dateparse import parse_datetime
 from django_filters import filters
 from drf_spectacular.utils import OpenApiParameter, extend_schema
+from jinja2 import Template
 from rest_framework import mixins
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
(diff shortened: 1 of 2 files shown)


rochacbruno (approved):

daviddavis (approved):

daviddavis:
Is it correct that the /collection_versions/all/ no longer accepts a page option? If so, we probably need a removal changelog entry.

gerrod3:
> Is it correct that the /collection_versions/all/ no longer accepts a page option? If so, we probably need a removal changelog entry.

Pretty sure it never accepted a page option. This endpoint is suppose to be unpaginated, but the code still had remnants of the paginated code, probably from copying and pasting.

gerrod3:
@rochacbruno Still doing some testing to see if I can get it even faster.  With this change I no longer get timed-out when accessing /collection_versions/all/ with 3000 collection_versions. Going to test it with 6000 collection versions and check that it still works, but I'm feeling that this might be the best current solution we can go with. Here are some results using django-silk and a distribution with 94 collection versions.

**Current**
Screenshot from 2021-04-15 17-46-17
**Streamed**
Screenshot from 2021-04-16 09-00-18

One other thing is that the streamed response doesn't use DRF's fancy styling, it looks pretty bad in the browser. Maybe I could fix this, but I don't think it's such a big deal.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -6,10 +6,12 @@
 from django.db.models import F
 from django.db.models.expressions import Window
 from django.db.models.functions.window import FirstValue
+from django.http import StreamingHttpResponse
```
fao89, line 9: cool!

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -446,15 +448,12 @@ def list(self, request, *args, **kwargs):
...
-            serializer = self.get_serializer(page, many=True, context=context)
-            return self.get_paginated_response(serializer.data)
-
-        serializer = self.get_serializer(queryset, many=True, context=context)
-        return Response(serializer.data)
+        cvs_template_string = "[{{ versions|map('tojson')|join(',') }}]"
+        cvs_template = Template(cvs_template_string)
+        serialized_map = map(lambda x: self.get_serializer(x, context=context).data, queryset)
+        streamed = map(lambda x: x.encode("utf-8"), cvs_template.stream(versions=serialized_map))
+        return StreamingHttpResponse(streamed)
```
fao89, line 456: should we take these to `UnpaginatedCollectionViewSet` also? Or the timeout only happens here?

fao89 (commented):
This is looking pretty good, why does it still as draft?

gerrod3:
@fao89 Should I tag this PR with the original issue for the unpaginated endpoint timeout https://pulp.plan.io/issues/8439?

fao89:
> @fao89 Should I tag this PR with the original issue for the unpaginated endpoint timeout https://pulp.plan.io/issues/8439?

I think so!

fao89 (approved):
Thank you!

bmbouter (approved):
This is really nice! Great job!

I believe we still need publications for 2 reasons:
* to resolve "the deprecated field issues"
* pre-computing something this large once and saving it is more efficient and will speed up the sync too. This is more of a nice-to-have really.

PR #563, opened 2021-04-19
Remove features deprecated in 3.12 for 3.13 compatibility

dralley:
closes: #8589
https://pulp.plan.io/issues/8589


dralley:
The test failures are because of a loosened restriction - "repository" and "repository_version" are now allowed to be used together.  @fao89 Does the ansible plugin want to enforce this? Depending on the use case I think it could be reasonable for some plugins to want to do that.

fao89:
> The test failures are because of a loosened restriction - "repository" and "repository_version" are now allowed to be used together. @fao89 Does the ansible plugin want to enforce this? Depending on the use case I think it could be reasonable for some plugins to want to do that.

I can't think of a reason to enforce it, currently, we already try both https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/v3/views.py#L57-L71

@daviddavis wdyt?

pulp_ansible/app/viewsets.py
```diff
@@ -258,10 +262,10 @@ def update(self, request, pk, **kwargs):
             remote_id=pk, last_synced_metadata_time__isnull=False
         ).all()
         lock.extend(repos)
-        async_result = enqueue_with_reservation(
+        async_result = dispatch(
             update_collection_remote,
             lock,
-            args=(pk,),
+            args=(str(pk),),
```
bmbouter, line 268: This str() can be removed I think.

bmbouter:
Using master (without this PR) I created a distribution, and here is what the database master/detail shows:

```
pulp=> select * from core_basedistribution;
               pulp_id                |         pulp_created          |       pulp_last_updated       |    pulp_type    |       name        | base_path | content_guard_id | remote_id 
--------------------------------------+-------------------------------+-------------------------------+-----------------+-------------------+-----------+------------------+-----------
 4084b54c-82ab-4826-8d6c-f4e772192cab | 2021-05-10 21:22:30.251446+00 | 2021-05-10 21:22:30.251458+00 | ansible.ansible | distribution26149 | 17985     |                  | 
(1 row)

pulp=> select * from ansible_ansibledistribution;
       basedistribution_ptr_id        |            repository_id             | repository_version_id 
--------------------------------------+--------------------------------------+-----------------------
 4084b54c-82ab-4826-8d6c-f4e772192cab | c1858c7b-1715-429a-a0fd-c7b0ccdad0e1 | 
(1 row)
```

Then I upgrade to this PR, run the migrations and the db now looks like:

```
pulp=> select * from core_basedistribution;
 pulp_id | pulp_created | pulp_last_updated | pulp_type | name | base_path | content_guard_id | remote_id 
---------+--------------+-------------------+-----------+------+-----------+------------------+-----------
(0 rows)

pulp=> select * from core_distribution;
               pulp_id                |         pulp_created          |       pulp_last_updated       |    pulp_type    |       name        | base_path | content_guard_id | publication_id | remote_id |            repository_id             | repository_version_id 
--------------------------------------+-------------------------------+-------------------------------+-----------------+-------------------+-----------+------------------+----------------+-----------+--------------------------------------+-----------------------
 4084b54c-82ab-4826-8d6c-f4e772192cab | 2021-05-10 21:24:50.448434+00 | 2021-05-10 21:24:50.448456+00 | ansible.ansible | distribution26149 | 17985     |                  |                |           | c1858c7b-1715-429a-a0fd-c7b0ccdad0e1 | 
(1 row)

pulp=> select * from ansible_ansibledistribution;
         distribution_ptr_id          
--------------------------------------
 4084b54c-82ab-4826-8d6c-f4e772192cab
(1 row)
```

So that all looks successful!

Also when starting the upgraded ansible, the CLI continues to list the distribution correctly:

```
(pulp) [vagrant@pulp3-source-fedora33 ~]$ pulp ansible distribution list
[
  {
    "pulp_href": "/pulp/api/v3/distributions/ansible/ansible/4084b54c-82ab-4826-8d6c-f4e772192cab/",
    "pulp_created": "2021-05-10T21:24:50.448434Z",
    "base_path": "17985",
    "content_guard": null,
    "name": "distribution26149",
    "repository": "/pulp/api/v3/repositories/ansible/ansible/c1858c7b-1715-429a-a0fd-c7b0ccdad0e1/",
    "repository_version": null,
    "client_url": "http://pulp3-source-fedora33.localhost.example.com/pulp_ansible/galaxy/17985/",
    "pulp_labels": {}
  }
]

```

So functionally speaking, this is all looking great.

bmbouter:
> > I can't think of a reason to enforce it, currently, we already try both https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/v3/views.py#L57-L71
> 
> Apologies for not seeing this sooner. My understanding (which is perhaps outdated) is that you can either distribute the repository (which distributes the latest version) or a particular repo version. So I'm not sure I understand the use case where a user would set both?

This is also my understanding.

dralley:
The semantics on publication-based plugins are:

* Set `repository` if you want the distribution to be automatically updated with the latest publications for that repo (updates are pushed to `publication`)
* Set `publication` if you want to temporarily override that (or initially set it) with a particular publication, or if you aren't setting `repository`, this value never changes

The semantics on repository-version based plugins are:

* Set `repository` if you want the distribution to automatically fetch the latest repository version for that repo (the update is `pull`ed), OR
* Set `repository_version`, which never changes.

What I'm asking is, in effect, if we want to keep these separate semantics, or if we want to make plugins like ansible use the same semantics as publication-based plugins.

bmbouter:
@dralley I'm not fully understanding the suggestion. I can tell I'm not because I don't see an opportunity to change the semantics. Right now I think since the detail serializer doesn't have a publication field so I'm thinking all we can do is to keep the semantics of repository-version based plugins. At some point in the future we will probably make `pulp_ansible` a publication-based plugin, but at that time we can adjust the serializers, introduce the right fields and switch the semantics. Is this helpful to give you the answer, or did I miss it?

dralley:
@bmbouter I'm not suggesting using publications, I'm suggesting using the same behavior.

i.e. New version gets created, if `repository` is set, we update `repository_version` on the distributions.

bmbouter:
I think this is waiting on pulpcore PR, is that right? If so, which one?

bmbouter:
I was told on IRC this was waiting on https://github.com/pulp/pulpcore/pull/1335

pulp_ansible/app/tasks/collections.py
```diff
@@ -711,8 +711,8 @@ async def _find_all_collections(self):
 
     async def _should_we_sync(self):
         """Check last synced metadata time."""
-        msg = _("no-op: Checking if remote changed since last sync.")
-        noop = ProgressReport(message=msg, code="noop")
+        msg = _("no_change: Checking if remote changed since last sync.")
+        noop = ProgressReport(message=msg, code="sync.no_change")
```
bmbouter, line 715: I didn't expect this, but I'm ok with it.

pulp_ansible/app/tasks/roles.py
```diff
@@ -50,7 +50,7 @@ def synchronize(remote_pk, repository_pk, mirror=False):
     )
     first_stage = RoleFirstStage(remote)
     d_version = DeclarativeVersion(first_stage, repository, mirror=mirror)
-    d_version.create()
+    return d_version.create()
```
bmbouter, line 53: I didn't expect this, but I'm ok with it.

pulp_ansible/app/tasks/roles.py
```diff
@@ -76,7 +76,7 @@ async def run(self):
         """
         Build and emit `DeclarativeContent` from the ansible metadata.
         """
-        with ProgressReport(message="Parsing Role Metadata", code="parsing.metadata") as pb:
+        with ProgressReport(message="Parsing Role Metadata", code="sync.parsing.metadata") as pb:
```
bmbouter, line 79: also this, but I'm ok with it.

bmbouter (approved):
Thank you @dralley !

fao89:
This change is compatible with 3.12? If so, I can do the release tomorrow!

pulp_ansible/app/tasks/collections.py
dralley: As an explanation, since more than one logical "task" could be going on within a literal "task", we've been updating progress reports to be specific about what progress they are reporting.

A publish can now happen in the same task as a sync, so therefore we want to make the progress reporting conventions more specific.

But you're right, that isn't clear from just the PR description and the issue.  Apologies, this is wrapping up a few different things at once.

bmbouter:
> This change is compatible with 3.12? If so, I can do the release tomorrow!

It is because the new `Distribution` object was introduced in 3.12

pulp_ansible/app/tasks/roles.py
dralley: Same here.  It doesn't change anything since we're not using this result, but the convention is that we *could* be using this result.

pulp_ansible/app/tasks/collections.py
bmbouter: It's no problem. Thank you @dralley !

dralley:
@fao89 Not 3.12.0, but yes 3.12.1.  There was an oversight where we forgot to put `RepositoryVersionRelatedField` into the plugin API in 3.12.0.

daviddavis (approved):

PR #565, opened 2021-04-23
Update CI from plugin-template

fao89:
[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -1,16 +1,16 @@
 # This config represents the latest values used when running the plugin-template. Any settings that
 # were not present before running plugin-template have been added with their default values.
 
-additional_plugins: []
 additional_repos:
-- branch: master
-  bindings: false
+- bindings: false
+  branch: master
   name: galaxy-importer
   org: ansible
 black: true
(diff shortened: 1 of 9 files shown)


template_config.yml
```diff
@@ -48,9 +48,8 @@ redmine_project: ansible_plugin
 stable_branch: null
 test_bindings: false
 test_cli: true
-test_fips_nightly: false
 test_performance: false
-test_released_plugin_with_next_pulpcore_release: true
+test_released_plugin_with_next_pulpcore_release: false
```
daviddavis, line 52: Why is this set to false?

fao89: it almost never passes
https://github.com/pulp/pulp_ansible/actions.html?page=1&workflow_file_name=nightly.yml

daviddavis: Oh ok. I'm not familiar with the job but I guess we can safely ignore it?

fao89: we've been ignoring all this time, I'm also not very familiar with this job

daviddavis (approved):

PR #568, opened 2021-04-28
Cherrypick bugfixes to 0.5

fao89:


daviddavis:
I see these issues are being backported to two releases (0.5.9 and 0.7.3)? I didn't think you could do that. Which milestone will they be associated with in redmine?

fao89:
> I see these issues are being backported to two releases (0.5.9 and 0.7.3)? I didn't think you could do that. Which milestone will they be associated with in redmine?

I need to file backport issues, what I did in the past was changing issue numbers to backport issue numbers right before running the release script

daviddavis:
Ah ok. That works.

daviddavis (approved):

PR #573, opened 2021-05-04
Optimize unpaginated collection_versions endpoint query

gerrod3:
[noissue]

This should reduce the amount of database queries 5x. So now the amount of queries for this endpoint is n + 4 where n is the number of collectionversions.  I'm not sure if I can remove the final n queries to make this constant since they are for the tags which are a many-to-many relation.

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -434,24 +434,15 @@ class UnpaginatedCollectionVersionViewSet(CollectionVersionViewSet):
         """
         distro_content = self._distro_content
 
-        return CollectionVersion.objects.select_related("content_ptr__contentartifact").filter(
-            pk__in=distro_content
-        )
+        return CollectionVersion.objects.select_related().filter(pk__in=distro_content)
 
     def list(self, request, *args, **kwargs):
         """
         Returns paginated CollectionVersions list.
         """
-        queryset = self.filter_queryset(self.get_queryset())
-        queryset = sorted(
-            queryset, key=lambda obj: semantic_version.Version(obj.version), reverse=True
-        )
+        queryset = self.get_queryset().iterator()
 
         context = self.get_serializer_context()
-        page = self.paginate_queryset(queryset)
-        if page is not None:
-            serializer = self.get_serializer(page, many=True, context=context)
-            return self.get_paginated_response(serializer.data)
 
         serializer = self.get_serializer(queryset, many=True, context=context)
         return Response(serializer.data)
--- a/CHANGES/8746.bugfix
+++ b/CHANGES/8746.bugfix
@@ -1,0 +1,1 @@
+Optimized unpaginated collection_versions endpoint
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -6,6 +6,7 @@ from rest_framework.reverse import reverse
 from rest_framework import serializers, relations
 
 from pulp_ansible.app import models
+from pulpcore.plugin.models import ContentArtifact
 
 
 class CollectionSerializer(serializers.ModelSerializer):
@@ -198,7 +199,8 @@ class UnpaginatedCollectionVersionSerializer(CollectionVersionListSerializer):
         """
         Get atrifact summary.
         """
-        return ArtifactRefSerializer(obj.contentartifact_set.get()).data
+        content_artifact = ContentArtifact.objects.select_related("artifact").filter(content=obj)
+        return ArtifactRefSerializer(content_artifact.get()).data
 
     def get_download_url(self, obj) -> str:
         """
@@ -206,7 +208,7 @@ class UnpaginatedCollectionVersionSerializer(CollectionVersionListSerializer):
         """
         host = settings.ANSIBLE_CONTENT_HOSTNAME.strip("/")
         distro_base_path = self.context["path"]
-        filename_path = obj.contentartifact_set.get().relative_path.lstrip("/")
+        filename_path = obj.relative_path.lstrip("/")
         download_url = f"{host}/{distro_base_path}/{filename_path}"
         return download_url
 


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -434,24 +439,55 @@ def get_queryset(self):
...
+        setattr(collection_version, "contentartifact", content_artifact)
+        return collection_version
 
     def list(self, request, *args, **kwargs):
         """
         Returns paginated CollectionVersions list.
         """
-        queryset = self.filter_queryset(self.get_queryset())
-        queryset = sorted(
-            queryset, key=lambda obj: semantic_version.Version(obj.version), reverse=True
```
fao89, line 447: we need to sort by version

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -434,24 +439,55 @@ def get_queryset(self):
...
+    def construct_full_collection_version(obj):
+        """Constructs a new CollectionVersion object."""
+        collection_attrs = {
+            "pk": obj["collection_id"],
+            "pulp_created": obj.pop("collection__pulp_created"),
+            "pulp_last_updated": obj.pop("collection__pulp_last_updated"),
+            "name": obj["name"],
+            "namespace": obj["namespace"],
+        }
+        collection = Collection(**collection_attrs)
```
fao89, line 474: I don't know much about django ORM, but this method seems to be rebuilding some django method

pulp_ansible/app/galaxy/v3/views.py
gerrod3: Why though? It's just a json dict that we are returning, does galaxy expect the dictionary to be ordered?

fao89: I don't know exactly the `ansible-galaxy` CLI code, but I think it expects the versions to be ordered:
https://galaxy.ansible.com/api/v2/collections/pulp/pulp_installer/versions/

gerrod3: But this endpoint is only for syncing, correct? They would still use the other endpoints that sort by version for installing collections.

fao89: Actually I checked the code and they just loop through versions 
https://github.com/ansible/ansible/blob/devel/lib/ansible/galaxy/api.py#L767-L847

fao89: you are right! `UnpaginatedCollectionVersionViewSet` is only used on sync

bmbouter: Dropping the ordering requirement probably would make it go faster.

bmbouter: and I agree it's only used on sync

pulp_ansible/app/galaxy/v3/views.py
bmbouter: I typically use cprofile to get python code timings and visualize them. (lmk if you want any help).

Also getting the SQL EXPLAIN statements for the SQL produced is then more easily analyzed to show 'oh look these queries are better/faster'.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -434,24 +439,55 @@ def get_queryset(self):
...
 
     def list(self, request, *args, **kwargs):
         """
         Returns paginated CollectionVersions list.
         """
-        queryset = self.filter_queryset(self.get_queryset())
-        queryset = sorted(
-            queryset, key=lambda obj: semantic_version.Version(obj.version), reverse=True
-        )
+        queryset = map(self.construct_full_collection_version, self.get_queryset().iterator())
```
dralley, line 488: A list comprehension would probably be clearer in this case.

gerrod3:
@bmbouter Here is some comparison data. I synced pulp_installer which has 33 collection versions.

**Without optimization:**
Screenshot from 2021-05-12 17-35-36

**With optimization:**
Screenshot from 2021-05-12 17-36-27

The old way does 4 + 5 * n database queries (4 + 5 * 33 = 169) while the new optimized way does 4 + n (4 + 33 = 37).  The new way alone is not enough to fix the timeout issues with large collection versions count (2000+), but this optimization with streaming endpoint #562 does fix the timeout issue for repositories up to 7000 collection versions. (as far as I tested)

These two changes only move the goal post down further and can't be a permanent solution. Do you think it's still worth to try them?

bmbouter:
Thank you for posting this and doing the analysis. I think I'm reading that the streaming option and this query optimization PR together resolved the issue as high as you've tested 7K collections. That's great. I'm suspecting of the two, the streaming resolution will resolve it on its own because even if the queries take longer, we can just make the batches smaller until it also works up to 7K.

If that's the case, that's what I'd recommend (keep the streaming resolution and not merge this query change). My thinking is that this mostly makes the ORM faster and it's not worth carrying that if we don't need to.

@gerrod3 would you be willing to test the "largest collection count you tested" (iirc 7K?) against just the streaming PR? And if it doesn't work maybe adjust the batch sizes to be smaller? Or is this not a great idea? wdyt?

gerrod3:
> @gerrod3 would you be willing to test the "largest collection count you tested" (iirc 7K?) against just the streaming PR? And if it doesn't work maybe adjust the batch sizes to be smaller? Or is this not a great idea? wdyt?

I'm not sure adjusting the batch_sizes will make the streaming PR pass on its own, but maybe it will.  I can adjust this PR to be less crazy, but still have some optimizations.  Like I said in our meeting, the 5 queries per collection version include 2 duplicates and 1 that could be removed with select_related (the current select_related is actually doing nothing on the query). Less queries will make the streaming endpoint faster.

bmbouter:
Also we could accept as-is and move on to other work? Overall I don't think this code is going to be here that long because I believe we need to switch pulp_ansible to use publications and have this be pre-computed instead. If we are going to keep this code around then I think we do need to be careful, but if not, not really.

@gerrod3 @daviddavis @fao89 wdyt should be done? Also are there concerns about converting pulp_ansible to a publication-based plugin soon? Any feedback is welcome.

gerrod3:
@bmbouter @fao89 @daviddavis I just simplified the optimization so now the amount of requests is 4 + 2n.  I think this is much more manageable and still provides much better performance. I think we should merge this variation.

bmbouter:
> @bmbouter @fao89 @daviddavis I just simplified the optimization so now the amount of requests is 4 + 2n. I think this is much more manageable and still provides much better performance. I think we should merge this variation.

This looks good to me. Is the stremaing response one already merged?

fao89:
This PR looked good to me before, I didn't approve so far because I was hoping someone would jump in and say: "here, use this django_orm_magic_method instead"

fao89 (commented):
But we need a changelog entry for this!
Other than that, consider this PR approved!

gerrod3:
> This looks good to me. Is the stremaing response one already merged?

The streaming response #562 is dependent on this PR.

@fao89 Added changed log.

fao89 (approved):
Thank you!

bmbouter:
Agreed. Thank you!

bmbouter:
@gerrod3 ready for merging?

PR #575, opened 2021-05-12
Use dispatch instead of enqueue_with_reservation

fao89:
[noissue]

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -365,12 +365,12 @@ class CollectionUploadViewSet(
         async_result = self._dispatch_import_collection_task(
             temp_file.pk, distro.repository, **kwargs
         )
-        CollectionImport.objects.create(task_id=async_result.id)
+        CollectionImport.objects.create(task_id=async_result.pk)
 
         data = {
             "task": reverse(
                 "collection-imports-detail",
-                kwargs={"path": path, "pk": async_result.id},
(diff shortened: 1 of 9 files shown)


daviddavis:
`dispatch` was introduced in pulpcore 3.12 so you'll need to update the pulpcore requirement

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -365,12 +365,12 @@ def create(self, request, path):
         async_result = self._dispatch_import_collection_task(
             temp_file.pk, distro.repository, **kwargs
         )
-        CollectionImport.objects.create(task_id=async_result.id)
+        CollectionImport.objects.create(task_id=async_result.pulp_id)
```
daviddavis, line 368:
```suggestion
        CollectionImport.objects.create(task_id=async_result.pk)
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -365,12 +365,12 @@ def create(self, request, path):
...
 
         data = {
             "task": reverse(
                 "collection-imports-detail",
-                kwargs={"path": path, "pk": async_result.id},
+                kwargs={"path": path, "pk": async_result.pulp_id},
```
daviddavis, line 373:
```suggestion
                kwargs={"path": path, "pk": async_result.pk},
```

daviddavis (approved):
Suggestion to use pk instead of pulp_id. Other than that, LGTM.

PR #576, opened 2021-05-20
Update CI + Upgrade tests

fao89:
[noissue]


.github/workflows/scripts/script.sh
```diff
@@ -95,6 +97,51 @@ cmd_prefix bash -c "PULP_DATABASES__default__USER=postgres django-admin test --n
...
+  cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-resource-manager"
+  cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-worker@1"
+  cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-worker@2"
+
+  echo "Restarting in 60 seconds"
+  sleep 60
+
+  # CLI commands to display plugin versions and content data
+  pulp status
+  pulp ansible content list
```
fao89, line 129: @gerrod3 @daviddavis I use pulp-cli here to show we still have content, I thought this was a generic command for all plugins. How can I template the right command here to show we still have data after migration?

.github/workflows/scripts/before_install.sh
```diff
@@ -97,6 +118,9 @@ fi
 cd pulp-cli
 pip install -e .
 pulp config create --base-url http://pulp --location tests/settings.toml
+sed -i "s/true/false/g" tests/settings.toml
```
fao89, line 121: another CLI question, do we have a better way of handling this?

.github/workflows/scripts/script.sh
fao89: https://github.com/pulp/plugin_template/blob/master/templates/github/.github/workflows/scripts/script.sh.j2#L147-L153

gerrod3: `pulp show --href /pulp/api/v3/content/` There is no generic content list command, we can add it. You could also use `pulp artifact list`, although technically you could have artifacts with no content, but that would be a weird scenario, and if you can have content with no artifacts, which is more likely since on-demand sync doesn't generate any artifacts.

.github/workflows/scripts/before_install.sh
fao89: I tried `--verify-ssl false`, I knew I was doing something wrong, thank you!

fao89:
@pulp/ansible can I have a review?

template_config.yml
```diff
@@ -52,4 +52,7 @@ test_performance: false
 test_released_plugin_with_next_pulpcore_release: false
 test_s3: true
 update_redmine: true
+upgrade_range:
+  - pulp_ansible_branch: 0.7
+    pulpcore_branch: 3.11
```
gerrod3, line 57: In the future when we are releasing pulp_ansible 0.9.0, will the upgrade tests check 0.7->0.8->0.9 and 0.7->0.9? 

Is this setting something we will change continuously from release to release or just keep constant?

pulp_ansible/tests/upgrade/pre/test_collection_install.py
```diff
@@ -0,0 +1,77 @@
...
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.client = gen_ansible_client()
+        cls.repo_api = RepositoriesAnsibleApi(cls.client)
+        cls.remote_collection_api = RemotesCollectionApi(cls.client)
+        cls.distributions_api = DistributionsAnsibleApi(cls.client)
+
+    def create_install_scenario(self, body, collection_name):
+        """Create Install scenario."""
+        repo = self.repo_api.create(gen_repo(name="pre_upgrade"))
```
gerrod3, line 38: This repo isn't cleaned up in the post tests. Probably not a problem if this is only ran in the CI.

gerrod3 (approved):
Looks good, left some questions.

template_config.yml
fao89: At this first moment, I see we keep it constant, but in the future, we may create some rule: "only test upgrade for latest 2 releases"

we can add entries, and test pulpcore x pulp_ansible combinations:
```yaml
upgrade_range:  
  - pulp_ansible_branch: 0.7    
    pulpcore_branch: 3.11
 - pulp_ansible_branch: 0.8    
    pulpcore_branch: 3.13
- pulp_ansible_branch: 0.8    
    pulpcore_branch: 3.14
```

pulp_ansible/tests/upgrade/pre/test_collection_install.py
fao89: Nice catch!

PR #577, opened 2021-05-25
WIP Add excludes filter to collections remote

gerrod3:
[noissue]

### DIFF
--- a/CHANGES/8857.feature
+++ b/CHANGES/8857.feature
@@ -1,0 +1,1 @@
+Added ``excludes`` parameter to sync call to perform synclist filtering
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -105,7 +105,7 @@ def update_collection_remote(remote_pk, *args, **kwargs):
         serializer.save()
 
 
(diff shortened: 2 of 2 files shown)


pulp_ansible/app/serializers.py
```diff
@@ -102,6 +102,15 @@ class AnsibleRepositorySyncURLSerializer(RepositorySyncURLSerializer):
     optimize = serializers.BooleanField(
         help_text=_("Whether to optimize sync or not."), default=True
     )
+    excludes_file = serializers.CharField(
```
rochacbruno, line 105: do we still need this on the serializer now that `excludes` are being passed to sync function as a parameter?

gerrod3: Yes, either way it needs to be there.

bmbouter: I believe we should have this data stored on the Remote. The high-level design is that all meaningful details for sync live on the Remote. We started introducing these sync-call options because remotes could be reused between multiple repositories and it became clear passing them at runtime was a better option, mirror. Or for things that are truly the functionality of the sync, e.g. `optimize`.

gerrod3:
Closing in favor of #654

PR #579, opened 2021-06-01
Release 0.8.0

dkliban:


requirements.txt
```diff
@@ -2,6 +2,6 @@ async_lru
 galaxy_importer>=0.3.1
 jsonschema>=3.0
 packaging
-pulpcore>=3.12.1
+pulpcore>=3.12.1,<3.14
```
fao89: I think we can also support 3.14, otherwise pulp-ansible may be a blocker for pulpcore 3.14 soon
wdyt @bmbouter @daviddavis @gerrod3 ?

gerrod3: I believe so, but I could be wrong, we do have a lot planned for 3.14 potentially.

bmbouter: It's hard to say without running and looking for deprecation warnings. If you get no deprecation warnings with 3.13 you can declare 3.14 safely (I think). I think @dkliban did some work on the CI to look for these deprecation warnings and point them out, but I forget.

bmbouter: Generally if it's a compatibility release against 3.13 it should declare 3.14 compat also to allow users to upgrade more easily, e.g. the single contianer etc etc.

dkliban: I updated the upper pulpcore bound to be 3.15.

fao89 (approved):
Thank you!

PR #592, opened 2021-07-07
Adding after failure to nightly + update CI

fao89:
[noissue]


.github/workflows/scripts/install.sh
```diff
@@ -80,6 +80,8 @@ fi
 
 cat >> vars/main.yaml << VARSYAML
 pulp_settings: {"allowed_export_paths": "/tmp", "allowed_import_paths": "/tmp", "ansible_api_hostname": "http://pulp:80", "ansible_content_hostname": "http://pulp:80/pulp/content"}
+pulp_scheme: http
```
bmbouter, line 83: is this preparing us for the switch to https?

fao89: yep, just switch to `https` and reapply the template

bmbouter: niiiiice

bmbouter (approved):
Thanks!

PR #593, opened 2021-07-07
AnsibleCollectionDeprecated as Content

fao89:
https://pulp.plan.io/issues/8205
closes #8205


pulp_ansible/app/migrations/0034_deprecation_content.py
```diff
@@ -0,0 +1,79 @@
...
+                pulp_id=deprecated_collection.pulp_id,
+            ).save()
+            repo_content.append(RepositoryContent(
+                repository=deprecated_collection.repository_version.repository,
+                content_id=deprecated_collection.pk,
+                version_added=deprecated_collection.repository_version,
+            ))
+
+        RepositoryContent.objects.bulk_create(repo_content)
+        AnsibleCollectionDeprecated.objects.all().delete()
```
fao89, line 28: @bmbouter @daviddavis @gerrod3 the PR is almost ready, the only thing I couldn't make work is the data migration.
Have you seen something like this before?
```
django.db.utils.OperationalError: cannot ALTER TABLE "core_repositorycontent" because it has pending trigger events
```

fao89: I believe I need to divide into 2 migrations:
https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/docs/ref/migration-operations.txt#L390-L395

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -295,14 +299,18 @@ def update(self, request, *args, **kwargs):
...
-            AnsibleCollectionDeprecated.objects.filter(
-                repository_version=repo_version, collection=collection
-            ).delete()
+        AnsibleCollectionDeprecated.objects.get_or_create(
+            namespace=collection.namespace, name=collection.name
+        )
+        deprecation = AnsibleCollectionDeprecated.objects.filter(
+            namespace=collection.namespace, name=collection.name
+        )
+        with repo_version.repository.new_version() as new_version:
```
bmbouter, line 308: This has to use a task because it's making a new repository version. I think it should use a brand new task.

pulp_ansible/app/models.py
```diff
@@ -227,6 +226,21 @@ class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
 
 
+class AnsibleCollectionDeprecated(Content):
```
newswangerd, line 229: Can we generalize this? I'd rather not have to create a new content type for each piece of metadata that gets added down the line.

bmbouter: Like we talked at our meetup about 1.5 weeks ago, we can do this but overall we think now isn't the right time. We could switch to a single model to store all the metadata (the rpm plugin has a few "bag" style models) at any point even if we continue with this PR as is (which is our proposal).

The reasoning is that while yes we will need additional models for additional metadata, in practice even if you put all the data into a generic, single model you'll still need additional code in various places anyway. So the simplicity it presents isn't really that beneficial in terms of avoiding shipping code to handle new metadata.

* it's unavoidable to have to parse the metadata for copying purposes. Copying a collection from one repo to another needs to bring some subset of the metadata. At that point having it unparsed to start with makes it harder.
* the data likely needs to be parsed to be included in the Galaxy APIs anyway. For example, with `deprecated` we'd have to parse this data anyway to be included in the Collection Serializer
* Merging remote metadata from a new sync, on-top of existing content which also has metadata requires joining. We can't just use data brought in at sync-time over metadata for collections already in the repo version.

So for now we're planning to keep this implementation pretty much as it is now, and we'll gladly generalize it later if that is helpful and when there are more examples for us to generalize with.

newswangerd: @bmbouter I hope this doesn't lead us to a nasty migration later.

bmbouter: @newswangerd I think it'll be straightforward given that the former objects are Content and the latter objects are Content.

Did you have any thoughts on the reasoning and conclusion we came to?

newswangerd: @bmbouter I'm not terribly happy about it, but do what you must.

bmbouter: I'm trying to explain that it's not a subjective design decision. There isn't an opportunity to make a "simple catch all solution" due to the bullets above that @daviddavis  and @fao89  and I could see. I can understand disappointment with the reality of that.

I'm unpacking the reasoning in bullets because I want this to be a shared decision. Can we reframe how this decision is being thought of to a "we must do what we must"? Basically I want your input to help us understand what we could do better, and if we can't find something, we accept the path we take together.

newswangerd: @bmbouter I understand that this is our best option and I see why it is. I just wish there was a simpler solution for supporting more metadata in the future. I don't have any suggestions to improve it. You can go ahead and implement the proposal as you've outlined here.

pulp_ansible/app/tasks/collections.py
```diff
@@ -105,6 +106,43 @@ def update_collection_remote(remote_pk, *args, **kwargs):
...
+    )
+
+    if deprecated:
+        add_content_units.append(deprecation.pk)
+    else:
+        remove_content_units.append(deprecation.pk)
+
+    repository = AnsibleRepository.objects.get(pk=repository_pk)
+    dispatch(
+        add_and_remove,
```
bmbouter, line 135: Is there a way to have us call `add_and_remove` directly instead?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -295,15 +304,29 @@ def update(self, request, *args, **kwargs):
         serializer = self.get_serializer(collection, data=request.data, partial=True)
         serializer.is_valid(raise_exception=True)
         deprecated_value = request.data.get("deprecated")
+        add_content_units = []
+        remove_content_units = []
+
+        deprecation, created = AnsibleCollectionDeprecated.objects.get_or_create(
```
bmbouter, line 310: @daviddavis with the new orphan cleanup behavior does anything additional need to happen here like an additional touch() call?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -295,15 +304,29 @@ def update(self, request, *args, **kwargs):
...
+            add_content_units.append(deprecation.pk)
         else:
-            AnsibleCollectionDeprecated.objects.filter(
-                repository_version=repo_version, collection=collection
-            ).delete()
-        return Response(serializer.data)
+            remove_content_units.append(deprecation.pk)
+
+        task = dispatch(
+            add_and_remove,
```
bmbouter, line 320: this looks great!

pulp_ansible/app/galaxy/v3/views.py
daviddavis: Yes, it's ugly but I think we'll need something like this:

```python
deprecation, created = AnsibleCollectionDeprecated.objects.get_or_create(
    namespace=collection.namespace, name=collection.name
)
if not created:
    try:
        deprecation.touch()
    except DatabaseError:
        # deprecation has since been removed. try to create it
        deprecation = AnsibleCollectionDeprecated.objects(namespace=collection.namespace, name=collection.name)

**Review by bmbouter — APPROVED** [review]
This looks great to me! Thanks so much @fao89 and @daviddavis ! @newswangerd and @rochacbruno FYI ^

File: CHANGES/8205.bugfix [hunk 7]
```diff
@@ -0,0 +1 @@
+Making deprecation exportable/importable
```
daviddavis, line 1:
```suggestion
Made deprecation exportable/importable
```

Also, I think this is more of a feature than a bugfix.

pulp_ansible/app/migrations/0036_update_repository_content.py
```diff
@@ -0,0 +1,76 @@
...
+            content=deprecated_collection,
+            repository_id=deprecated_collection.repository_id,
+            version_added_id=deprecated_collection.version_added_id,
+        ))
+
+    for repo_version_id, total in counter.items():
+        count_obj = RepositoryVersionContentDetails(
+            content_type=deprecated_collection.pulp_type,
+            repository_version_id=repo_version_id,
+            count=total,
```
daviddavis, line 28: I don't know that this is correct. Say I have 2 deprecated collections in repo version 2 but version 1 has 1 of them too. Then this number should be 1 and not 2. 

Also, I think you'll need to calculate the deprecations removed as well.

bmbouter: Can't we just call repo_version._compute_counts() ?

fao89: ORM methods don't work here

daviddavis: Actually this is even more of a huge pain because I think we'll need to delete any existing RepositoryVersionContentDetails first and then totally recompute them:

https://github.com/pulp/pulpcore/blob/1d76e8a316a5fc92cc3397f01329be10fe68a7a5/pulpcore/app/models/repository.py#L926

Maybe we should just not compute `RepositoryVersionContentDetails` in our migration and add a note how users can recompute these themselves.

bmbouter: Oh I see. :( I agree

bmbouter: I'm ok to leave these counts as is, and if it's a problem folks can file a new issue about them being incorrect.

daviddavis: To be clear, you're suggesting we don't update RepositoryVersionContentDetails (and thus remove this part of the migration)? If so, +1 from me.

fao89: Actually, RepositoryVersionContentDetails uncovered a big issue,

> I don't know that this is correct. Say I have 2 deprecated collections in repo version 2 but version 1 has 1 of them too. Then this number should be 1 and not 2.

> Also, I think you'll need to calculate the deprecations removed as well.

I have to check if deprecations were removed at some point, right now I only do `version_added`, I need to cover `version_removed` as well

```python
repo_content.append(RepositoryContent(
            content=deprecated_collection,
            repository_id=deprecated_collection.repository_id,
            version_added_id=deprecated_collection.version_added_id,
        ))
```

daviddavis: Yes, I think you may need to change how you calculate version_added as well. If a deprecation is in versions 1, 2, 3 then there should only be one RepositoryContent with version_added=1.

fao89: can I have another round of review?
I've improved the migration, but I'm not sure if it is covering version_added/version_removed well

pulp_ansible/app/migrations/0035_deprecation_content.py
```diff
@@ -0,0 +1,80 @@
...
+
+    dependencies = [
+        ('core', '0068_add_timestamp_of_interest'),
+        ('ansible', '0034_handle_jsonfield_warnings'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            code=migrations.RunPython.noop,
+            reverse_code=delete_deprecation_type,
```
newswangerd, line 51: I don't fully understand what's happening in this step, but shouldn't all delete actions happen after `migrate_data_from_old_model_to_new_model_up`?

newswangerd:
@ironfroggy this has some pretty hairy looking migrations. We'll need to make sure to test this carefully when we upgrade to the next version of pulp_ansible.

fao89:
> @ironfroggy this has some pretty hairy looking migrations. We'll need to make sure to test this carefully when we upgrade to the next version of pulp_ansible.

We were trying to rebuild history, adding Deprecation to the correspondent RepositoryVersion, but it was getting too complicated.
We can simplify this migration if we consider only the latest RepositoryVersion.

Given that, I have a question, do you need all the deprecation history? Or do you only need the current deprecation status (latest RepositoryVersion of each Repository)?

newswangerd:
Only using the latest repo version is fine by me

daviddavis:
I'm fine with only updating the latest repo version but two thoughts:

* we should make it very clear in the changelog that we're only updating the latest repo version.
* we should probably double check with Katello that it's ok.

fao89:
> I'm fine with only updating the latest repo version but two thoughts:
> 
> * we should make it very clear in the changelog that we're only updating the latest repo version.
> * we should probably double check with Katello that it's ok.

I've improved the changelog message and double-checked with them, @daviddavis could you please re-review?

daviddavis:
The migration looks good. A test for deprecation in the test_import or test_export test would be nice but optional.

Also, can we get an approval from someone on galaxy?

fao89:
> The migration looks good. A test for deprecation in the test_import or test_export test would be nice but optional.
> 
> Also, can we get an approval from someone on galaxy?

I've added deprecation to import/export tests, 
@newswangerd could take another look at this PR?

newswangerd (approved):
I think this looks good to me.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -168,12 +170,16 @@ class CollectionViewSet(
...
-        self.deprecated_collections_context = deprecated  # needed by get__serializer_context
-        return deprecated
+        deprecations = Q()
+        for namespace, name in AnsibleCollectionDeprecated.objects.filter(
+            pk__in=self._distro_content
+        ).values_list("namespace", "name"):
+            deprecations |= Q(namespace=namespace, name=name)
+        collection_pks = []
+        if len(deprecations):
+            collection_pks = Collection.objects.filter(deprecations).values_list("pk", flat=True)
```
fao89, line 180: @rochacbruno I queried all deprecations at latest_repo_version.content,
got all namespaces and names and filtered the collections with that

PR #598, opened 2021-07-21
Fixed ``JSONField`` deprecation warnings

fao89:
closes #9119


bmbouter (approved):
Thanks!

bmbouter:
@daviddavis did you think this needs anything more?

daviddavis (approved):

newswangerd (approved):

pulp_ansible/app/modelresource.py
```diff
@@ -30,6 +30,9 @@ class CollectionVersionContentResource(BaseContentResource):
     Resource for import/export of ansible_collectionversion-content entities.
     """
 
+    contents = fields.Field(
+        column_name="contents", attribute="contents", widget=JSONWidget(), default=list()
```
daviddavis, line 34: Nice find :+1:

daviddavis (approved):

pulp_ansible/app/modelresource.py
daviddavis: Bonus points if you want to add a comment here too:

```
# required for jsonb fields in PostgreSQL database
```

fao89: Actually, their introspection handles it well, the problem happens when the default is list.
For the other JSONFields that had dict as default, I didn't need to do anything

PR #604, opened 2021-07-22
Reduce progress report increment

fao89:
[noissue]


fao89:
it is just a PoC, solution should live in pulpcore:
https://github.com/pulp/pulpcore/pull/1512

fao89:
https://github.com/pulp/pulp_ansible/pull/605

PR #605, opened 2021-07-22
Improve efficiency on progress reports

fao89:
[noissue]

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -371,10 +371,6 @@ class CollectionSyncFirstStage(Stage):
 
         """
         super().__init__()
-        msg = _("Parsing CollectionVersion Metadata")
-        self.parsing_metadata_progress_bar = ProgressReport(
-            message=msg, code="sync.parsing.metadata"
-        )
         self.remote = remote
         self.repository = repository
         self.optimize = optimize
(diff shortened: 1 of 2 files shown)


bmbouter:
I think this PR will both change the counts, and also cause almost no progress updates until the FirstStage is totally parsed. That's kind of the beauty of having the progress report update where it is now.

What if instead, we just switch the ProgressBar to use the context manager form. That way when save's occur the ProgressBar's throttling feature will just reduce the number of actual save's efficiently.

To do that I think instead of instantiating the ProgressBar in the init, just instantiate it in the `FirstStage.run()` and have it saved on `self` and have `FirstStage.run()` have all of it's code roughly inside the context manager of the ProgressBar.

pulp_ansible/app/tasks/collections.py
```diff
@@ -493,7 +493,6 @@ async def _add_collection_version(self, api_version, collection_version_url, met
             d_artifacts=[d_artifact],
             extra_data=extra_data,
         )
-        self.parsing_metadata_progress_bar.increment()
         await self.put(d_content)
```
gerrod3, line 496: Here I think we need to return 1 + number of dependencies synced (from line 471). Basically we need to sum up the return values from all gathers() to accurately update the amount of collection versions synced. 
```
num_synced = sum(await asyncio.gather(*tasks))
self.parsing_metadata_progress_bar.increase_by(num_synced)
```

bmbouter (approved):
This looks like exactly what I was hoping for!

PR #607, opened 2021-07-29
Use proxy auth when refreshing token

fao89:
[noissue]


rochacbruno (approved):
It did not fixed the bug, however is worth to merge, lgtm.

PR #630, opened 2021-08-27
Remove lines to hardcode contents to use JSONWidget

daviddavis:
[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ async_lru~=1.0
 galaxy_importer>=0.3.1
 jsonschema>=3.0
 packaging
-pulpcore>=3.15.0.dev
+pulpcore>=3.15.1
 PyYAML
 semantic_version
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,6 +2,6 @@ async_lru~=1.0
 galaxy_importer>=0.3.1
 jsonschema>=3.0
 packaging
-pulpcore>=3.15.0.dev
+pulpcore>=3.15.1
```
fao89, line 5: I think it will break the CI even when you add "Required PR"

daviddavis: I think it runs against 3.16.0.dev (which is >= 3.15.1) and thus works?

fao89: yeah, you are right, I thought pulpcore master was 3.15.1.dev, but it is 3.16.0.dev

fao89 (approved):
thank you!

PR #632, opened 2021-08-31
Update CI files

pulpbot:
[noissue]


fao89 (approved):
it is breaking because it needs https://github.com/pulp/pulpcore/pull/1585

PR #641, opened 2021-09-09
Make noop async

fao89:
[noissue]

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -19,7 +19,6 @@ from galaxy_importer.collection import CollectionFilename
 from galaxy_importer.exceptions import ImporterError
 from pkg_resources import Requirement
 
-from pulpcore.plugin.constants import TASK_STATES
 from pulpcore.plugin.models import (
     Artifact,
     ContentArtifact,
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -705,46 +704,42 @@ async def _find_all_collections(self):
...
-        root, api_version = await self._get_root_api(self.remote.url)
-        if api_version == 3:
-            downloader = self.remote.get_downloader(
-                url=root, silence_errors_for_response_status_codes={404}
-            )
-            try:
-                metadata = parse_metadata(await downloader.run())
-            except FileNotFoundError:
+        async with ProgressReport(message=msg, code="sync.no_change") as noop:
+            if not self.repository.remote:
```
dralley, line 708: This and `.cast()` are likely to result in queries

pulp_ansible/app/tasks/collections.py
```diff
@@ -705,46 +710,39 @@ async def _find_all_collections(self):
...
         """Check last synced metadata time."""
         msg = _("no_change: Checking if remote changed since last sync.")
-        noop = ProgressReport(message=msg, code="sync.no_change")
-        noop.state = TASK_STATES.COMPLETED
-        noop.save()
-
-        if not self.repository.remote:
-            return True
-
-        if self.remote != self.repository.remote.cast():
```
dralley, line 715: https://github.com/pulp/pulpcore/pull/1614/files#diff-03bf88cd6f4c5051a7b1478ba05ea63287e7bb4c29feb00413b80ba35d22ccf2R75

pulp_ansible/app/tasks/collections.py
```diff
@@ -705,46 +710,39 @@ async def _find_all_collections(self):
     async def _should_we_sync(self):
         """Check last synced metadata time."""
         msg = _("no_change: Checking if remote changed since last sync.")
-        noop = ProgressReport(message=msg, code="sync.no_change")
-        noop.state = TASK_STATES.COMPLETED
-        noop.save()
-
-        if not self.repository.remote:
-            return True
```
dralley, line 713: Same here, I don't know if the refactor was for other reasons but if it was just to resolve this issue you could do something like `await sync_to_async(self.repository.remote)` (might need to be wrapped in a function?)

fao89: I decided to move it out of the async code

bmbouter (approved):
This would probably benefit from a refactor issue attached to this PR in case it needs a backport.

I probably would have kept it structured as-is and converted the calls to be async safe, but I'm ok with doing it this way too.

PR #654, opened 2021-09-29
Sync excludes collections found at /excludes/ endpoint

gerrod3:
fixes: #8194

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -19,6 +19,7 @@ from galaxy_importer.collection import import_collection as process_collection
 from galaxy_importer.collection import CollectionFilename
 from galaxy_importer.exceptions import ImporterError
 from pkg_resources import Requirement
+from rest_framework.serializers import ValidationError
 
 from pulpcore.plugin.models import (
     Artifact,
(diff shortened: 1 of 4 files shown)


rochacbruno (approved):

pulp_ansible/app/tasks/collections.py
```diff
@@ -455,6 +476,9 @@ async def _add_collection_version(self, api_version, collection_version_url, met
             version=metadata["version"],
         )
         cv_unique = attrgetter("namespace", "name", "version")(collection_version)
+        fullname = ".".join(cv_unique[:2])
+        if fullname in self.exclude_info and cv_unique[2] in self.exclude_info[fullname]:
```
fao89, line 480: could you add a test for checking the scenario when we want to exclude only a specific version?

pulp_ansible/app/tasks/collections.py
```diff
@@ -634,15 +645,32 @@ async def _download_unpaginated_metadata(self):
...
-            try:
-                collection_metadata_list = parse_metadata(await downloader.run())
-            except FileNotFoundError:
-                pass
-            else:
+            exc_downloader = self.remote.get_downloader(
+                url=excludes_endpoint, silence_errors_for_response_status_codes={404}
+            )
+            tasks = [loop.create_task(col_downloader.run()), loop.create_task(exc_downloader.run())]
+            results = await asyncio.gather(*tasks, return_exceptions=True)
```
fao89, line 659:
```suggestion
            results = await asyncio.gather(*tasks, return_exceptions=True)
            col_result, exc_result = results
```
and then replace results[0]...

pulp_ansible/app/tasks/collections.py
fao89:
```suggestion
        fullname = ".".join(cv_unique[:2])
        if fullname in self.exclude_info and collection_version.version in self.exclude_info[fullname]:
```
`cv_unique[2]` made me confuse, I had to read multiple times to understand

mdellweg:
Should we backport to all versions starting from 0.7?

PR #659, opened 2021-10-04
Adjust dispatch calls to core>=3.15 interface

mdellweg:
fixes # 9483

Maybe we can declare some of the resources shared. LMK if you want to see it in this PR.

### DIFF
--- a/pulp_ansible/app/viewsets.py
+++ b/pulp_ansible/app/viewsets.py
@@ -267,7 +267,7 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
 
         result = dispatch(
             sync_func,
-            [repository, remote],
+            exclusive_resources=[repository, remote],
             kwargs=sync_kwargs,
         )
         return OperationPostponedResponse(result, request)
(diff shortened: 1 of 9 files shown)


fao89 (approved):
LGTM!

pulp_ansible/app/viewsets.py
```diff
@@ -267,7 +267,7 @@ def sync(self, request, pk):
 
         result = dispatch(
             sync_func,
-            [repository, remote],
+            exclusive_resources=[repository, remote],
```
ipanova, line 270: remote can be among shared_resources.

pulp_ansible/app/viewsets.py
```diff
@@ -401,7 +401,7 @@ def create(self, request):
 
         config, repos = self._process_config(config)
 
-        async_result = dispatch(copy_content, repos, args=[config], kwargs={})
+        async_result = dispatch(copy_content, exclusive_resources=repos, args=[config], kwargs={})
```
ipanova, line 404: since this is a content copy operation, the source repo can be among shared_resources, `_process_config` should be updated so it returns 2 lists of repos.

pulp_ansible/app/viewsets.py
mdellweg: I drafted this to be the minimal change to unbreak master, but i can add shared_resources if desired. Not sure if it were a feature though.

PR #664, opened 2021-10-08
Adds GitRemote

dkliban:
This remote has no 'policy', but instead has a 'metadata_only' boolean. It defaults to False. When set to True,
only metadata is saved in Pulp. Clients will install the content directly from the remote's URL.

closes: #9345
https://pulp.plan.io/issues/9345


fao89:
> I'm curious about what we should do if a user requests a piece of content that was synced with metadata_only?

I think this is the equivalent of on demand, which we don't support on pulp_ansible yet

pulp_ansible/app/serializers.py
```diff
@@ -83,6 +84,26 @@ class Meta:
...
+        required=False,
+    )
+
+    class Meta:
+        fields_list = []
+        for field in RemoteSerializer.Meta.fields:
+            if field != "policy":
+                fields_list.append(field)
+        fields_list.append("metadata_only")
+        fields = tuple(fields_list)
```
gerrod3, line 103: Instead of this just use the `excludes` variable to remove `policy`: https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include

daviddavis: How about:

```python
tuple(set(RemoteSerializer.Meta.fields) - {'policy'}) + ('metadata_only',)
```

PR #673, opened 2021-11-02
Raises the galaxy-importer version requirement

dkliban:
re: #9346

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 async_lru~=1.0
-galaxy_importer>=0.3.1
+galaxy_importer>=0.4.1
 GitPython>=3.1.24
 jsonschema>=3.0
 packaging


requirements.txt
```diff
@@ -1,5 +1,5 @@
 async_lru~=1.0
-galaxy_importer>=0.3.1
+galaxy_importer>=0.4.1
```
fao89, line 2: 👍🏽

fao89 (approved):

mdellweg (changes requested):
Can you please also remove installing of galaxy-importer from the before-install script? I believe we want to treat that as an external dependency, and so we should only test against released versions. I think, this is what broke the oci nightly image for several days.

mdellweg (approved):
I see we are installing it via the plugin-template now, which will install from pypi in a release workflow. That will prevent broken releases of this type at least.

fao89:
merging to see if it fixes pulp_installer CI:
https://github.com/pulp/pulp_installer/actions/runs/1451438364

PR #724, opened 2021-11-17
Use Github Issues

fao89:
[noissue]


mdellweg:
Do you think, this should include a preliminary run of towncrier to make sure the merged changelogs point to the right tickets?

fao89:
> Do you think, this should include a preliminary run of towncrier to make sure the merged changelogs point to the right tickets?

I'm not sure, I don't know how towncrier would behave if we run it 2 times for the same version

gerrod3 (approved):

mdellweg:
> > Do you think, this should include a preliminary run of towncrier to make sure the merged changelogs point to the right tickets?
> 
> I'm not sure, I don't know how towncrier would behave if we run it 2 times for the same version

I would run it for x.y.z-unreleased and adjust the docs by hand in the actual release. The other option is to wait and adjust all the old links at release time. Neither is a nice solution.

fao89:
I think I will manually create these issues on github and change the number of the changelog entries

mdellweg (approved):

PR #754, opened 2021-11-18
CollectionVersion Signature Support

gerrod3:
Haven't tested this yet, but this PR has two commits: first for the Signature model declaration and generic viewing API, second for the signing task to actually create signed content.  This needs a corresponding pulpcore PR with an async version of the SigningService's sign method.

TODO:

- [x] Test, Test, Test
- [x] Add signature field to galaxy collection version api
- [x] ?Add signature syncing?


pulp_ansible/app/serializers.py
```diff
@@ -477,6 +479,54 @@ class Meta:
...
+    A serializer for signature models.
+    """
+
+    content = DetailRelatedField(
+        help_text=_("The content this signature is pointing to."),
+        view_name_pattern=r"content(-.*/.*)-detail",
+        queryset=CollectionVersion.objects.all(),
+    )
+    public_key = serializers.CharField(help_text=_("The public key used to verify the signature"))
+    pubkey_fingerprint = serializers.CharField(help_text=_("The fingerprint of the public key."))
```
mdellweg, line 493: I think, these fields could benefit from a keyring table. Specifically if you sign all your stuff with the same key.
If we don't want to do that now, the fingerprint alone should be sufficient. Think about how to get this information in a sync workflow.

gerrod3: Would we store the keyring table? Not really sure what you mean by how the fields would benefit.

I was thinking when I add the signature to the collection-version metadata, it would be added as a link to view all the signatures on that collection-version. That view would be a json list of `{'pubkey_fingerprint': ..., 'signature': 'signature_filename_link'}` for each signature. This would be what I think the CLI would use (it might still be too many requests for the CLI though). For syncing I would create a special endpoint that would aggregate all of the signatures for each collection-version into one giant json list, similar to how the `collections/all` endpoint works now.

mdellweg: I was thinking, that you will end up with many collectionversions singed by the same key, and while the fingerprint is something small, the actual public key can be rather big. So duplicating the key data in the CollectionVersionSignature table seems resource hungry.
I would imagine, there would be a public-key-table (fields: fingerprint, public_key, uid, _whatever-else-interesting_, ...) and the collection would reference this table either via uuid/pk, or simply by the fingerprint. The CollectionSerializer can then annotate the signature lookup query with e.g. the `pulic_key__uid` field to provide a human readable answer to "Who signed this?".

pulp_ansible/app/models.py
```diff
@@ -178,6 +179,22 @@ class Meta:
...
+    """
+    A content type representing a signature that is attached to a content unit
+    # This model might live in pulpcore if it can be abstracted.
+    """
+    TYPE = "collection_signature"
+
+    content = models.ForeignKey(CollectionVersion, on_delete=models.CASCADE, related_name="signatures")
+    public_key = models.TextField()  # How long can public keys be?
+    pubkey_fingerprint = models.TextField()
+    signing_service = models.ForeignKey(SigningService, on_delete=models.SET_NULL, related_name="signatures", null=True)
```
bmbouter, line 192: If we have a FK to the signing_service do we need the `public_key` and `pubkey_fingerprint` ?

pulp_ansible/app/models.py
```diff
@@ -178,6 +179,22 @@ class Meta:
...
 
 
+class CollectionVersionSignature(Content):
+    """
+    A content type representing a signature that is attached to a content unit
+    # This model might live in pulpcore if it can be abstracted.
+    """
+    TYPE = "collection_signature"
+
+    content = models.ForeignKey(CollectionVersion, on_delete=models.CASCADE, related_name="signatures")
```
bmbouter, line 189: Is this the signature itself? If so maybe `s/content/signature/`?

pulp_ansible/app/serializers.py
bmbouter: Re my comments in the model section, maybe we just pull the fingerprint only into this serializer via the through looking on `signing_service` ?

pulp_ansible/app/models.py
gerrod3: No this is a reference to the content this signature is for.

pulp_ansible/app/models.py
gerrod3: This can be null, since I was assuming in the future when we sync signatures from a remote they wouldn't have an associated signing service, so you would at least need the `pubkey_fingerprint` for easy filtering.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,91 @@
...
+            self.first_stage,  # CollectionSigningStage
+            QueryExistingArtifacts(),
+            ArtifactSaver(),
+            # QueryExistingContents(),
+            ContentSaver(),
+        ]
+        return pipeline
+
+
+class CollectionSigningStage(Stage):
```
bmbouter, line 54: Maybe call this `CollectionSigningFirstStage` ?

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # Signatures typically have their date created embedded within the signature which means multiple
+            # signatures on the same content from the same signing-service will produce multiple valid and
+            # *unique* signatures which won't be de-duplicated. Maybe another setting should be added to act
+            # as a policy for these potential "duplicate" signatures.
+            result = await self.signing_service.asign(t.name)
+            # Is it always a guarantee that the result will have a "signature" where the file is?
+            artifact = Artifact.init_and_validate(result["signature"])
+            # Should this relative path be the name of the "{collection-version}-signature"?
+            da = DeclarativeArtifact(artifact, url="NA", relative_path="MANIFEST.json.asc")
```
bmbouter, line 83: I mentioned on the call, but wanted to leave it here that I think we should be storing this signature entirely in memory and there would be no `Artifact` or `DeclarativeArtifact` here. Especially given that we don't want the API to have to read the Artifact data later from S3 or Azure object store backends.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,91 @@
...
+            d = tempfile.mkdtemp(dir=self.base_dir)
+            os.chdir(d)  # GPG doesn't take filename into account for signatures, but not sure about other methods
+            with open(f"MANIFEST.json", 'w') as t:
+                json.dump(collection_version.manifest, t)
+
+            # Signatures typically have their date created embedded within the signature which means multiple
+            # signatures on the same content from the same signing-service will produce multiple valid and
+            # *unique* signatures which won't be de-duplicated. Maybe another setting should be added to act
+            # as a policy for these potential "duplicate" signatures.
+            result = await self.signing_service.asign(t.name)
```
bmbouter, line 79: Perhaps this could dispatch all the signatures and then wait on the "first finished" send that one down the pipeline, and then wait for the next "first finished" etc. This is how we run the downloaders stage [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/plugin/stages/artifact_stages.py#L173).

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, content, signing_service):
+        super().__init__()
+        self.content = content
+        self.signing_service = signing_service
+        self.base_dir = os.getcwd()
+
+    async def run(self):
+        """does the magic."""
+        async for collection_version in sync_to_async_iterable(self.content.iterator()):
+            # Can add extra parallelization here with tasks
```
bmbouter, line 68: If you mean task in the tasking system, we avoid keeping "this task running and dispatching another that we need to wait on" because in cases where the worker count is 1 it would deadlock. For parallelization I think we can get coroutine based parallelization with the design in my comment below (maybe).

pulp_ansible/app/viewsets.py
```diff
@@ -288,6 +319,40 @@ def sync(self, request, pk):
         )
         return OperationPostponedResponse(result, request)
 
+    @extend_schema(
+        description="Trigger an asynchronous task to sync Ansible content.",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=True, methods=["post"], serializer_class=AnsibleRepositorySignatureSerializer)
+    def sign(self, request, pk):
```
bmbouter, line 327: I like this design, to sign everything and then make a new repo version out of it.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __init__(self, content, signing_service):
+        super().__init__()
+        self.content = content
+        self.signing_service = signing_service
+        self.base_dir = os.getcwd()
+
+    async def run(self):
+        """does the magic."""
+        async for collection_version in sync_to_async_iterable(self.content.iterator()):
```
bmbouter, line 67: What if we filtered this some here to exclude those "already signed CollectionVersions" ? Perhaps my suggestion is not a good one because we like to let the stages do the work.

This gets me thinking about a problematic case though if all we can do is sign everything in a repo, then repo version has collectionversion A-1.1.1, and it gets signed. Then B-2.2.2 gets added and signed it will sign A-1.1.1 again. And then C-3.3.3 gets added it will sign A-1.1.1 and B-2.2.2 again and so forth.

After thinking this over more, I think we should have it generate signatures just for CollectionVersion items that do not have signatures.

pulp_ansible/app/models.py
```diff
@@ -178,6 +179,22 @@ class Meta:
...
+    public_key = models.TextField()  # How long can public keys be?
+    pubkey_fingerprint = models.TextField()
+    signing_service = models.ForeignKey(SigningService, on_delete=models.SET_NULL, related_name="signatures", null=True)
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
```
bmbouter, line 195: If we want stages to recognize CollectionVersionSignature's as the same so it can deduplicate them, we probably should make the `content` field the uniqueness constraint.

pulp_ansible/app/tasks/signature.py
gerrod3: Sure, I'll change the model to have the signature be apart of it.

pulp_ansible/app/tasks/signature.py
gerrod3: This comment is referring to async tasks.

pulp_ansible/app/tasks/signature.py
gerrod3: The API also takes an optional list of content to sign so you don't have to sign everything in the repository. I was thinking of making this optional list also be added to the repository if not already present, but maybe that feature would be too much functionality.

As for an exclude "already signed content", I think that would be a good idea for use cases of syncing new content and then signing them with your own keys. I think users might want to sign content with two different services, so this should be an option and not the only default.

pulp_ansible/app/models.py
rochacbruno: Yes, there is that case where we sync from remote and also the case for signing service being deleted or keys rotated but content not changed.

Signing Service must be detached from the signature, we could store as a soft reference.

pulp_ansible/app/models.py
rochacbruno: IIRC on container modeling `content` is the signature base64 (c/c @ipanova) probably best to have specific naming here such as:

```py
signature_content =
collection_version =
```

pulp_ansible/app/models.py
```diff
@@ -178,6 +179,38 @@ class Meta:
...
+
+    PROTECTED_FROM_RECLAIM = False
+    TYPE = "collection_signature"
+
+    signed_collection = models.ForeignKey(
+        CollectionVersion, on_delete=models.CASCADE, related_name="signatures"
+    )
+    data = models.BinaryField()  # This is what the pulp_container folks used
+    digest = models.CharField(max_length=64)  # does adding a min length improve db efficiency?
+    pubkey_fingerprint = models.CharField(max_length=64)
```
mdellweg, line 204: I kept thinking about this, and i think it is ok, if we can deliver the name of the signer (key user_id) for the keys we know and say unknown for keys we don't have/care about.
So this is kind of a voluntary foreign key into a possible future key database.

pulp_ansible/app/models.py
ipanova: we went with a more explicit naming and the field that is a reference to the content is named `signed_manifest` https://github.com/pulp/pulp_container/blob/signing/pulp_container/app/models.py#L236
the payload of the signature itself is stored in `data` field, we will serialize it to `content` in the live api for the  client compatibility

pulp_ansible/app/models.py
gerrod3: How are you thinking of getting the key user id into the signature model? Would a new field have to be added to the SigningService, or maybe a new method to extract it from the public key (although I don't think key user id is a requirement for the public key, atleast in OpenPGP)?

gerrod3:
> Hello @gerrod3! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
>     * In the file [`pulp_ansible/app/tasks/collections.py`](https://github.com/pulp/pulp_ansible/blob/20b62391146989bdd4db2ce19ac45556d7fa358d/pulp_ansible/app/tasks/collections.py):
> 
> 
> > [Line 538:23](https://github.com/pulp/pulp_ansible/blob/20b62391146989bdd4db2ce19ac45556d7fa358d/pulp_ansible/app/tasks/collections.py#L538): [E231](https://duckduckgo.com/?q=pep8%20E231) missing whitespace after ':'
> > [Line 538:23](https://github.com/pulp/pulp_ansible/blob/20b62391146989bdd4db2ce19ac45556d7fa358d/pulp_ansible/app/tasks/collections.py#L538): [E701](https://duckduckgo.com/?q=pep8%20E701) multiple statements on one line (colon)
> > [Line 538:24](https://github.com/pulp/pulp_ansible/blob/20b62391146989bdd4db2ce19ac45556d7fa358d/pulp_ansible/app/tasks/collections.py#L538): [E225](https://duckduckgo.com/?q=pep8%20E225) missing whitespace around operator
> 
> ##### Comment last updated at 2021-12-01 00:15:06 UTC

PEP 8 bot doesn't seem to understand the walrus operator.

pulp_ansible/app/models.py
mdellweg: I would not put it in the signature model at all. We have the key in the signing service so we can either add a user id field there or extract from the key field on the go. We would report unknown for keys we don't have.

bmbouter: I thought the value of this was in cases where the signature wasn't made by a Pulp SigningSerice, but instead sync'd from another Pulp where the signature was produced.

mdellweg: > I thought the value of this was in cases where the signature wasn't made by a Pulp SigningSerice, but instead sync'd from another Pulp where the signature was produced.

That is why i think we need a pubkey model eventually.

pulp_ansible/app/tasks/collections.py
```diff
@@ -534,6 +535,16 @@ async def _add_collection_version(self, api_version, collection_version_url, met
         await self.parsing_metadata_progress_bar.aincrement()
         await self.put(d_content)
 
+        if signatures := metadata.get("signatures"):
```
ipanova, line 538: in pulp-container we've added a boolean option which can be specified during sync `signed_only`. In case it's enabled it would skip any content that has no signature, does it make sense to add it here as well?

rochacbruno: makes sense as we are going to have sort of `SYNC_ONLY_SIGNED_CONTENT=True` for Private Hub config, we were thinking about doing that per remote and then pass it to the sync task.

pulp_ansible/app/viewsets.py
```diff
@@ -288,6 +318,44 @@ def sync(self, request, pk):
         )
         return OperationPostponedResponse(result, request)
 
+    @extend_schema(
+        description="Trigger an asynchronous task to sync Ansible content.",
```
rochacbruno, line 322:
```suggestion
        description="Trigger an asynchronous task to sign Ansible content.",
```

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,113 @@
...
+
+    async def sign_collection_version(self, collection_version):
+        """Signs the collection version."""
+        # Limits the number of subprocesses spawned/running at one time
+        # Should everything be under the semaphore or just the signing part?
+        async with self.semaphore:
+            # We use the manifest to create the signature
+            with tempfile.TemporaryDirectory() as d:
+                # OpenPGP doesn't take filename into account for signatures, not sure about others
+                with open(f"{d}/MANIFEST.json", 'w') as t:
```
rochacbruno, line 73: does it work when running on s3 storage?

rochacbruno (approved):
LGTM, I will do more tests and report if I find any issue

https://github.com/ansible/galaxy_ng/pull/1076

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -171,6 +171,22 @@ class CollectionNamespaceSerializer(serializers.Serializer):
...
+
+    signature = serializers.SerializerMethodField()
+
+    def get_signature(self, obj) -> str:
+        # If I check for signing service I would incur a N+1 db issue, assuming ascii for now
+        return bytes(obj.data).decode("utf-8")
+
+    class Meta:
+        model = models.CollectionVersionSignature
+        fields = ("signature", "pubkey_fingerprint")
```
rochacbruno, line 187: can we add signing-service.name here? it will be useful to see via API if a signature is one created within the system or synced from a remote.

rochacbruno: It would be also useful to have the pulp-created time surfaced here for debugging purposes.

rochacbruno (approved):

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,114 @@
...
+class SigningDeclarativeVersion(DeclarativeVersion):
+    """Custom signature pipeline."""
+
+    def pipeline_stages(self, new_version):
+        """The stages for the signing process."""
+        pipeline = [
+            self.first_stage,  # CollectionSigningFirstStage
+            ContentSaver(),
+        ]
+        return pipeline
```
mdellweg, line 50: Nice pattern!

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -235,6 +261,13 @@ def get_git_commit_sha(self, obj) -> str:
         if not content_artifact.get().artifact:
             return content_artifact.get().remoteartifact_set.all()[0].url[-40:]
 
+    def get_signatures(self, obj):
+        """
+        Get the signatures.
+        """
+        filtered_signatures = obj.signatures.filter(pk__in=self.context["sigs"])
```
mdellweg, line 268: This is a bit fuzzy to me. If i get it correctly, this only works when the collection version is listed as part of a distribution.
What if we did (assuming that passing `None` to `pk__in` is like not passing at all [with python you never _know_]):
```suggestion
        filtered_signatures = obj.signatures.filter(pk__in=self.context.get("sigs"))
```

Then again, this is used in the galaxy api, and there may always be that `_distro_content`.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    This stage signs the content and creates CollectionVersionSignatures.
+    """
+
+    def __init__(self, content, signing_service):
+        """Initialize Signing first stage."""
+        super().__init__()
+        self.content = content
+        self.signing_service = signing_service
+        # Maybe make this value a pulp_ansible constant setting?
```
mdellweg, line 63: Yes, i would make this configurable. Think about having an over the network signing service with harsh rate limits. You may be better off setting this to 1.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,114 @@
...
+        self.semaphore = asyncio.Semaphore(10)
+
+    async def sign_collection_version(self, collection_version):
+        """Signs the collection version."""
+        # Limits the number of subprocesses spawned/running at one time
+        # Should everything be under the semaphore or just the signing part?
```
mdellweg, line 69: Good question. The nice part about this approach is that we also protect the working directory with the semaphore.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,114 @@
...
+        async with self.semaphore:
+            # We use the manifest to create the signature
+            with tempfile.TemporaryDirectory() as d:
```
mdellweg, line 72:
```suggestion
            with tempfile.TemporaryDirectory(dir=".") as d:
```
BTW: aiofiles.tempfile may be interesting.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,114 @@
...
+                # OpenPGP doesn't take filename into account for signatures, not sure about others
+                with open(f"{d}/MANIFEST.json", "w") as t:
+                    json.dump(collection_version.manifest, t)
+                result = await self.signing_service.asign(t.name)
+                # Is it always a guarantee that the result will have a signature where the file is?
```
mdellweg, line 77: That's the contract the signing service has with the signing script. It is verified with the `verify` called on `save`.

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,114 @@
...
+        ntotal = await sync_to_async(new_content.count)()
+        nmsg = _("Signing new CollectionVersions")
+        async with ProgressReport(message=nmsg, code="sign.new.signature", total=ntotal) as p:
+            self.progress_report = p
+            async for collection_version in sync_to_async_iterable(new_content.iterator()):
+                tasks.append(asyncio.create_task(self.sign_collection_version(collection_version)))
+            await asyncio.gather(*tasks)
+
+        # Add any signatures already present in Pulp if part of content list
+        present_content = current_signatures.filter(signed_collection__in=self.content)
```
mdellweg, line 108: This is lying a little bit. If the present signature is already in the repo version, it will still be counted here, but no actual content will be added to the new repo version.

pulp_ansible/app/viewsets.py
```diff
@@ -290,6 +320,45 @@ def sync(self, request, pk):
...
+
+        repository = self.get_object()
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        signing_service = serializer.validated_data["signing_service"]
+        content = serializer.validated_data["content_units"]
+
+        if "*" in content:
+            content_units = ["*"]
```
mdellweg, line 342: This is unfortunate. `content_units` used to be a `dict`.

mdellweg (commented):
Overall LGTM but still some questions/remarks

pulp_ansible/app/galaxy/v3/serializers.py
gerrod3: I tried passing in None, django gives an error. I'll make the default an empty list, that should return an empty queryset.

pulp_ansible/app/tasks/signature.py
gerrod3: You're right. I'll try a different query which excludes signatures already in the repo_version.

pulp_ansible/app/viewsets.py
gerrod3: I don't know what you mean. I thought it was always a `list` of strings and if at any point there is an asterisk in it the list would switched to only be an asterisk because mixed behavior wouldn't make sense.

mdellweg: Oh, now i see. The confusion is contained to this function. I read the line 332 `content_units = {}` and thought the argument to the task was a `dict`. But you are converting it back to a list in 349 `content_units = list(content_units.keys())`. Why are we not passing the already extracted `content_unit_pks`?

pulp_ansible/app/viewsets.py
```diff
@@ -290,6 +320,45 @@ def sync(self, request, pk):
...
+        else:
+            for url in content:
+                content_units[NamedModelViewSet.extract_pk(url)] = url
+            content_units_pks = set(content_units.keys())
```
mdellweg, line 346: dict keys are sufficiently unique already...
```suggestion
            content_units_pks = list(content_units.keys())
```

pulp_ansible/app/viewsets.py
mdellweg: To be fair, i see the same pattern of redefining the from dict to list in the modify action in pulpcore.

gerrod3: Yes this was confusing, I changed the variable names so `content_units` is always a `dict` and `content_units_pks` is always a `list`.

rochacbruno:
LGMT 💯

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -171,6 +171,30 @@ class CollectionNamespaceSerializer(serializers.Serializer):
...
+    """
+
+    signature = serializers.SerializerMethodField()
+    signing_service = serializers.SlugRelatedField(
+        slug_field="name",
+        allow_null=True,
+        read_only=True,
+    )
+
+    def get_signature(self, obj) -> str:
```
bmbouter, line 186: nitpick: with pulp_ansible not using the type hinting I'd leave it out to stay consistent with the rest of the codebase.

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -171,6 +171,30 @@ class CollectionNamespaceSerializer(serializers.Serializer):
...
+        """
+        Get the signature data.
+        """
+        # Maybe I should decode into ascii? utf8 should be superset of ascii
```
bmbouter, line 190: I think utf-8 is right and also can this comment be removed?

pulp_ansible/app/tasks/signature.py
```diff
@@ -0,0 +1,119 @@
...
+    AnsibleRepository,
+    CollectionVersion,
+    CollectionVersionSignature,
+)
+
+from django.conf import settings
+from pulpcore.plugin.models import SigningService, ProgressReport
+from pulpcore.plugin.sync import sync_to_async_iterable, sync_to_async
+
+_ = gettext.gettext
```
bmbouter, line 24: Would you mind doing this in the import as `from gettext import gettext as _` ?

pulp_ansible/app/tasks/signature.py
bmbouter: I agree!

bmbouter (approved):
I left some small nitpick comments. If you wouldn't mind addressing those, then I'm LGTM. Thank you so much!

pulp_ansible/tests/functional/api/collection/test_signatures.py
```diff
@@ -0,0 +1,199 @@
...
+    TestCaseUsingBindings,
+    skip_if,
+)
+from pulp_ansible.tests.functional.constants import TEST_COLLECTION_CONFIGS
+from orionutils.generator import build_collection
+from pulpcore.client.pulp_ansible import AnsibleCollectionsApi, ContentCollectionSignaturesApi
+from pulp_ansible.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+
+
+class CRUDCollectionVersionSignatures(TestCaseUsingBindings, SyncHelpersMixin):
```
bmbouter, line 20: I'm hoping these can be switched to the pytest tests (some other time). We can't afford to take the time to change this prior to merging.

PR #761, opened 2021-11-30
Fix functional tests with roles

mdellweg:
Functional tests truncated too many tables. Especially roles and role
permissions must not be wiped for pulp to function properly.

[noissue]


mdellweg:
What was the reason to wipe tables that way in the first place?
Can we do this more carefully with e.g. `AnsibleRepository.objects.delete()` like statements?

bmbouter (approved):
Thanks!

PR #760, opened 2021-11-28
Update CI files

pulpbot:
[noissue]

PR #762, opened 2021-11-30
Update GHA config with latest plugin_template

dkliban:
This change also reflects the change in pulpcore branch from master to main.

[noissue]


bmbouter (approved):
Thanks! Also we should close https://github.com/pulp/pulp_ansible/pull/760 right?

PR #767, opened 2021-12-14
Make it clear that the Git remote is a tech preview feature.

dkliban:
[noissue]


pulp_ansible/tests/functional/api/collection/test_sync.py
```diff
@@ -7,7 +7,10 @@
 from pulp_ansible.tests.functional.constants import TEST_COLLECTION_CONFIGS
 from orionutils.generator import build_collection
 from pulpcore.client.pulp_ansible import PulpAnsibleArtifactsCollectionsV3Api
-from pulp_ansible.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+from pulp_ansible.tests.functional.utils import (
+    monitor_task,
+    set_up_module as setUpModule,
+)  # noqa:F401
```
gerrod3, line 13: I think this needs to be next to the setUpModule, even though I'm not sure what we use it for.

gerrod3: After looking at noqas in pulpcore, I think it is supposed to go after the top parenthesis.

PR #793, opened 2022-01-07
Fix deprecation migrations 0035 and 0036

### LINKED ISSUE #791: Deprecation content migration error.
When upgrading to pulp ansible 0.10, I get this error:

```
django.db.utils.IntegrityError: could not create unique index "ansible_ansiblecollectio_namespace_name_e1aa1c6d_uniq" DETAIL: Key (namespace, name)=(splunk, enterprise_security) is duplicated. 
```

Steps to reproduce:
- upload two versions of a collection to pulp ansible
- move one of the versions to a new repository
- attempt to perform the upgrade

After performing the failed upgrade, it looks like there is a duplicate collection deprecated content for each repository that the deprecated content was in:

```
galaxy_ng=# select * from ansible_ansiblecollectiondeprecated;
            content_ptr_id            |  namespace  |      name       |            collection_id             |            repository_id             |           version_added_id
--------------------------------------+-------------+-----------------+--------------------------------------+--------------------------------------+--------------------------------------
 ee99298a-86b3-48a1-8aba-6e105b5075eb | newswangerd | main_collection | 41ad8982-f094-4e28-8520-d2e0a119f2ab | fe424277-a0c2-4df0-b4e5-3324b2e2de90 | fe4a53d5-afa7-40c0-ba52-c89dace5a147
 5fca8a31-5562-46cb-bac2-d57cfbcc875c | newswangerd | main_collection | 41ad8982-f094-4e28-8520-d2e0a119f2ab | eb3573b9-1281-4dd7-9ea8-3a40b356892c | bfcee45c-1327-4b18-baf6-f02d483c5be7
 ```

My theory is that these duplicates are getting created here when the data is migrated from the old deprecation table to the new one: https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/migrations/0035_deprecation_content.py#L16

### PR DESCRIPTION
**Pull request by mdellweg**
The transition of deprecation to being repository content failed on a
uniquenes contstraint. The migration has been rewritten to use
deduplicated objects before adding the constraint to the database
schema.

fixes #791

### CODE REVIEW

File: pulp_ansible/app/migrations/0036_update_repository_content.py [hunk 1]
```diff
@@ -9,52 +9,69 @@ def create_repository_content_for_deprecation(apps, schema_editor):
...
+        for deprecation in AnsibleCollectionDeprecated.objects.filter(name=name, namespace=namespace, repository_id__isnull=False):
+            repo_content.append(RepositoryContent(
+                    content=new_deprecation,
+                    repository_id=deprecation.repository_id,
+                    version_added_id=deprecation.version_added_id,
+                ))
+            count[deprecation.version_added_id] += 1
+            if len(repo_content) >= 1024:
+                RepositoryContent.objects.bulk_create(repo_content)
+                repo_content.clear()
```
fao89, line 29: cool!

pulp_ansible/app/migrations/0036_update_repository_content.py
```diff
@@ -9,52 +9,69 @@ def create_repository_content_for_deprecation(apps, schema_editor):
...
-                content=deprecation,
-                repository_id=deprecation.repository_id,
-                version_added_id=deprecation.version_added_id,
-            ))
-        count[deprecation.version_added_id] += 1
+    # Create RepositoryContent with deduplicated deprecation items
+    for name, namespace in AnsibleCollectionDeprecated.objects.values_list("name", "namespace").distinct():
+        new_deprecation = AnsibleCollectionDeprecated(name=name, namespace=namespace)
+        new_deprecation.save()
+        for deprecation in AnsibleCollectionDeprecated.objects.filter(name=name, namespace=namespace, repository_id__isnull=False):
```
fao89, line 20: I see what was the issue now

fao89 (commented):
so far LGTM

mdellweg:
Now this fails for pending triggers. I think i know, i can "fix" this by marking the migration steps as not atomic, but i'd like to avoid that.

mdellweg:
Anyone knows, what operation creates those triggers?

pulp_ansible/app/migrations/0036_update_repository_content.py
```diff
@@ -9,56 +9,77 @@ def create_repository_content_for_deprecation(apps, schema_editor):
...
+
+    AnsibleCollectionDeprecated.objects.filter(repository_id__isnull=True).delete()
 
 
 class Migration(migrations.Migration):
 
+    # You cannot change data and modify the a table in the same migration.
+    # Since this is an already released migration, we cannot split this one in two.
+    # So the only resort is running the steps of this migration without transactions.
+    atomic = False
```
mdellweg, line 81: @fao89 @bmbouter I'd rather avoid this, but i see no option.

mdellweg:
The migration applies cleanly now, but the failure i don't understand.

mdellweg:
> > Anyone knows, what operation creates those triggers?
> 
> which triggers?

Postgres triggers. It seems they prevent changing the tables data and schema in the same transaction.

pulp_ansible/app/migrations/0036_update_repository_content.py
```diff
@@ -9,56 +9,77 @@ def create_repository_content_for_deprecation(apps, schema_editor):
...
-    for deprecation in AnsibleCollectionDeprecated.objects.using(db_alias).all():
-        repo_content.append(RepositoryContent(
-                content=deprecation,
-                repository_id=deprecation.repository_id,
-                version_added_id=deprecation.version_added_id,
-            ))
-        count[deprecation.version_added_id] += 1
+    # Create RepositoryContent with deduplicated deprecation items
+    for name, namespace in AnsibleCollectionDeprecated.objects.values_list("name", "namespace").distinct():
+        new_deprecation = AnsibleCollectionDeprecated.objects.create(pulp_type="collection_deprecation", name=name, namespace=namespace)
```
fao89, line 18:
```suggestion
        new_deprecation = AnsibleCollectionDeprecated.objects.create(pulp_type="ansible.collection_deprecation", name=name, namespace=namespace)
```


```
In [1]: AnsibleCollectionDeprecated().save()

In [2]: AnsibleCollectionDeprecated.objects.first().__dict__
Out[2]: 
{'_state': <django.db.models.base.ModelState at 0x7f92beb6cf40>,
 'pulp_id': UUID('ba9da7e1-c237-4e16-bb0f-508345f4c954'),
 'pulp_created': datetime.datetime(2022, 1, 10, 13, 42, 40, 101129, tzinfo=<UTC>),
 'pulp_last_updated': datetime.datetime(2022, 1, 10, 13, 42, 40, 101157, tzinfo=<UTC>),
 'pulp_type': 'ansible.collection_deprecation',
 'upstream_id': None,
 'timestamp_of_interest': datetime.datetime(2022, 1, 10, 13, 42, 40, 101176, tzinfo=<UTC>),
 'content_ptr_id': UUID('ba9da7e1-c237-4e16-bb0f-508345f4c954'),
 'namespace': '',
 'name': '',
 '_initial_state': {'pulp_id': UUID('ba9da7e1-c237-4e16-bb0f-508345f4c954'),
  'pulp_created': datetime.datetime(2022, 1, 10, 13, 42, 40, 101129, tzinfo=<UTC>),
  'pulp_last_updated': datetime.datetime(2022, 1, 10, 13, 42, 40, 101157, tzinfo=<UTC>),
  'pulp_type': 'ansible.collection_deprecation',
  'upstream_id': None,
  'timestamp_of_interest': datetime.datetime(2022, 1, 10, 13, 42, 40, 101176, tzinfo=<UTC>),
  'content_ptr_id': UUID('ba9da7e1-c237-4e16-bb0f-508345f4c954'),
  'namespace': '',
  'name': ''}}

```

pulp_ansible/app/migrations/0035_deprecation_content.py
```diff
@@ -35,7 +35,7 @@ def migrate_data_from_old_model_to_new_model_down(apps, schema_editor):
             repository_version_id=deprecation.version_added_id,
         ).save()
         deprecation.version_memberships.all().delete()
-    RepositoryVersionContentDetails.objects.filter(content_type=deprecation.pulp_type).delete()
+    RepositoryVersionContentDetails.objects.filter(content_type="collection_deprecation").delete()
```
fao89, line 38:
```suggestion
    RepositoryVersionContentDetails.objects.filter(content_type="ansible.collection_deprecation").delete()
```

pulp_ansible/app/migrations/0036_update_repository_content.py
```diff
@@ -9,56 +9,77 @@ def create_repository_content_for_deprecation(apps, schema_editor):
...
+                repo_content.clear()
+    RepositoryContent.objects.bulk_create(repo_content)
+    repo_content.clear()
+    AnsibleCollectionDeprecated.objects.filter(repository_id__isnull=False).delete()
 
+    # Update Repository content counts
     for repo_version_id, total in count.items():
         count_obj = RepositoryVersionContentDetails(
-            content_type=deprecation.pulp_type,
+            content_type="collection_deprecation",
```
fao89, line 36:
```suggestion
            content_type="ansible.collection_deprecation",
```

pulp_ansible/app/migrations/0036_update_repository_content.py
```diff
@@ -9,56 +9,77 @@ def create_repository_content_for_deprecation(apps, schema_editor):
...
             repository_version_id=repo_version_id,
             count=total,
             count_type="A", # Added
         )
         counts_list.append(count_obj)
 
         count_obj = RepositoryVersionContentDetails(
-            content_type=deprecation.pulp_type,
+            content_type="collection_deprecation",
```
fao89, line 44:
```suggestion
            content_type="ansible.collection_deprecation",
```

pulp_ansible/app/migrations/0036_update_repository_content.py
mdellweg: Thank you so much i would not have found this in a week.

pulp_ansible/app/migrations/0036_update_repository_content.py
bmbouter: I'm not sure of the answer here so feel free to educate me. It seems the issue of one transaction with data changes and SQL alter changes is the crux of the problem. If that's the case and they want to be separated how abut we move all data changes to one migration and leave all SQL alter changes in the other? I think I'm still clinging to the (maybe incorrect) belief that more of this diff should be happening in migration 35. You would know better than I, wdyt?

mdellweg: If we were to introduce these migrations first, i'd agree. But as the matter of things is, a user may have migrated beyond 35 already and failing to migrate 36. At that point we need to fix 36 for him. And i believe we cannot introduce new migrations in the chain. That is why i think, we are stuck fixing it _in_ 36.

bmbouter: I think I only now understand what you were trying to tell me before. What you are saying makes sense, but I've not heard of any users actually ending up in that state. I do worry about the risk to all users with this being atomic to serve a perhaps theoretical user base. We're picking from the lesser of two evils here so I won't claim fixing migration 35 is a big win, but I do think it's better. @mdellweg what do you think about this perspective?

mdellweg: I am the theoretician here. And my biggest concern is that we might redefine the scope of already published migrations.
Also when stuck in that position between migrations 35 and 36 the only way to repair (apart from a complete rollback from backup) would be to checkout the old codebase, revert 35, then install the new codebase again and migrate forward. So the rule that i think applies here (for exactly that crazy dance) is "we don't change existing migrations".
OTOH, i see that migrations 35 and 36 always appear in pairs, and when applied successfully, the result is good enough before _and_ after the change. So the only way to get in the bad situation will be a single failed upgrade. And there you just need to be extra careful to complete all shipped migrations with the same codebase.
Having expressed my concerns loudly here, I yield and will have a go at the combined fix.

mdellweg:
ping @bmbouter

bmbouter (approved):
This looks good to me. Can @fao89 also review and if he is lgtm then let's merge it plz.

fao89 (approved):
LGTM

Issue #728, opened 2021-11-17
Move existing API endpoints under /plugin/ansible.


pulpbot:
Author: @newswangerd (newswangerd)

Redmine Issue: 9040, https://pulp.plan.io/issues/9040

---

Add the new `v3/plugin/ansible` endpoints. Old `v3/` endpoints will be redirected as following:

- `{base_path}/v3/collections/*` -> `v3/plugin/ansible/collections/{base_path}/index/*`
- `{base_path}/v3/artifacts/*` -> `v3/plugin/ansible/collections/{base_path}/artifacts/*`
- `{base_path}/v3/imports/*` -> `v3/plugin/ansible/imports/*`
- `{base_path}/v3/collections/all` -> `v3/plugin/ansible/collections/{base_path}/all-collections`
- `{base_path}/v3/collection_versions/` -> `v3/plugin/ansible/collections/{base_path}/all-versions`

Example:

`pulp_ansible/galaxy/published/api/v3/collections/newswangerd/collection_demo` will redirect to `pulp_ansible/galaxy/published/api/v3/plugin/ansible/collections/published/index/newswangerd/collection_demo`

The old endpoints should all be marked as deprecated.

PR #798, opened 2022-01-20
Galaxy API Refactor stage 1

newswangerd:
This PR does the following:
- moves the existing collection views under `/plugin/ansible/`
- redirects the legacy v3 endpoints to their counterparts in `/plugin/ansible/`
- adds a new configuration option `ANSIBLE_DEFAULT_DISTRIBUTION_PATH` that allows users to configure a default distribution base path for the API
- adds a new configuration option `ANSIBLE_URL_NAMESPACE` that allows django URL namespace to be set on `reverse` so that urls can be configured to point correctly to the galaxy APIs when pulp ansible is deployed as part of automation hub
- adds the get `v3/artifacts/path/file` API endpoint from galaxy_ng
- adds the `ContentRedirectContentGuard` content guard from galaxy_ng

#728


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -161,12 +168,13 @@ class CollectionViewSet(
...
     """
 
-    authentication_classes = []
-    permission_classes = []
     serializer_class = CollectionSerializer
     filterset_class = CollectionFilter
     pagination_class = LimitOffsetPagination
 
+    def urlpattern(*args, **kwargs):
+        return "ansible-collection-viewset"
```
mdellweg, line 176: I can believe this is some kind of hack, but it looks weird to see the term "viewset" in a url pattern.
(And yes, we use the field `viewset_name` for something resembling a path in the access policy for historical reasons.)

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -351,11 +359,12 @@ class CollectionUploadViewSet(
     ViewSet for Collection Uploads.
     """
 
-    authentication_classes = []
-    permission_classes = []
```
mdellweg, line 355: This enables access control by virtue of the default permission class(es), right?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -404,14 +413,41 @@ def create(self, request, path):
 
         data = {
             "task": reverse(
-                "collection-imports-detail",
-                kwargs={"path": path, "pk": async_result.pk},
+                settings.ANSIBLE_URL_NAMESPACE + "collection-imports-detail",
```
mdellweg, line 416: Injecting the `URL_NAMESPACE` here, makes me think, we need to do the same when we connect the views to paths in `urls.py`.

pulp_ansible/app/models.py
```diff
@@ -285,3 +293,57 @@ class AnsibleDistribution(Distribution):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+
+
+def _gen_secret():
+    return os.urandom(32)
+
+
+class ContentRedirectContentGuard(ContentGuard):
```
mdellweg, line 302: There is a redirecting content guard with preauthenticated URL in pulp_container. Can we reuse that one (after moving it to pulpcore)?
https://github.com/pulp/pulp_container/blob/main/pulp_container/app/models.py#L588

newswangerd: We can. IIRC, we just copied the redirect content guard from pulp container pretty as is.

pulp_ansible/app/galaxy/v3/views.py
newswangerd: That's correct. These viewsets inherit from the default drf viewsets, which include the default auth and permission classes.

pulp_ansible/app/galaxy/v3/views.py
newswangerd: I wasn't sure what to name these, so I just came up with something quick and dirty. Please let me know how to make these more pulpy

bmbouter: I haven't looked myself, but maybe look at `show-urls` to get a feel for the patterns in place before and after this PR.

pulp_ansible/app/models.py
bmbouter: +1 to moving this to pulpcore.

pulp_ansible/app/settings.py
```diff
@@ -22,3 +22,5 @@
 
 ANSIBLE_API_HOSTNAME = "https://" + socket.getfqdn()
 ANSIBLE_CONTENT_HOSTNAME = settings.CONTENT_ORIGIN + "/pulp/content"
+ANSIBLE_DEFAULT_DISTRIBUTION_PATH = None
+ANSIBLE_URL_NAMESPACE = ""
```
bmbouter, line 26: Adding docs with some examples would help me understand this better (and we need it to merge anyway). I think put them [here](https://github.com/pulp/pulp_ansible/blob/main/docs/settings.rst).

pulp_ansible/app/galaxy/v3/views.py
mdellweg: I mean, the last argument is used to look for the corresponding urlpattern in urls.py. Yes.
And i was thinking, what might happen, if that setting stops being `""`. I bet, the reverse lookup would fail to find the url_pattern if they are not prefixed in the same way. (Matter of experiment...)

pulp_ansible/app/settings.py
newswangerd: @mdellweg @bmbouter there’s one thing I wanted to get your feedback on. One of the problems I had to solve was to figure out how to get pulp ansible to route correctly to the galaxy_ng apis any time reverse is called. The solution I came up with is the ANSIBLE_URL_NAMESPACE setting which basically allows you to inject a router namespace into the reverse lookup which tells pulp ansible to use the URL routes configured under the galaxy_ng app instead (that’s what’s happening here: https://github.com/pulp/pulp_ansible/pull/798/files#diff-d74e2c9333d6150adbd3c142482e4ba75697f468d983b35e696ab7e4fcfe8ceaR64)

There’s another potential solution to this problem and that’s to just set the `GALAXY_API_ROOT` in pulp ansible to the same as the API root in galaxy ng. I tested this and it seems to work, but the problem I ran into is that it will also include v1 and v2 of the API and I couldn’t figure out how to solve that without making breaking changes to the `GALAXY_API_ROOT` setting. I sort of like this approach better because then there’s no separation between the pulp and galaxy APIs, but it does make it a little bit harder to declare when we don’t want a route from pulp ansible like the v1 and v2 apis and eventually the configuration endpoint, which will have to be different in galaxy ng.

I was wondering what you think about these two approaches?

mdellweg: I'm not sure to see all aspects of the challenge here. First we need to make sure, that pulp_ansible has a consistent api on it's own. So with setting `GALAXY_API_ROOT` you can basically move the location of pulp_ansible's galaxy-api around? What if we introduce settings to move the versions of the api independently (`GALAXY_APIV1_ROOT`, ...)? Or even provide a setting that describes which versions of the api we want to expose at all?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -408,14 +418,58 @@ def create(self, request, path):
...
+
+    def get(self, request, *args, **kwargs):
+        distro_base_path = self.kwargs["distro_base_path"]
+        filename = self.kwargs["filename"]
+        prefix = settings.CONTENT_PATH_PREFIX.strip("/")
+        distribution = AnsibleDistribution.objects.get(base_path=distro_base_path)
+
+        url = "{host}/{prefix}/{distro_base_path}/{filename}".format(
+            host=settings.CONTENT_ORIGIN.strip("/"),
+            prefix=prefix,
```
awcrosby, line 460: could both of these params pull from settings here to make it easier to read?
```suggestion
        distribution = AnsibleDistribution.objects.get(base_path=distro_base_path)

        url = "{host}/{prefix}/{distro_base_path}/{filename}".format(
            host=settings.CONTENT_ORIGIN.strip("/"),
            prefix=settings.CONTENT_PATH_PREFIX.strip("/"),
```

newswangerd: done

newswangerd:
Turns out my remotes were messed up and I've been pushing changes directly to git@github.com:pulp/pulp_ansible.git. I'm closing this PR and re-opening it from my forked copy of pulp ansible #849

PR #802, opened 2022-01-27
Use proxy auth credentials of a Remote when syncing content

### LINKED ISSUE #801: Synchronization using authenticated proxy is failing
Using proxy authentication returns 407.

galaxy_ng issue: https://issues.redhat.com/browse/AAH-1243

fao89:
closes #801

### DIFF
--- a/pulp_ansible/tests/functional/api/collection/v3/test_sync.py
+++ b/pulp_ansible/tests/functional/api/collection/v3/test_sync.py
@@ -203,6 +203,54 @@ class SyncCollectionsFromPulpServerTestCase(TestCaseUsingBindings, SyncHelpersMi
         msg = "absent.not_present does not exist"
         self.assertIn(msg, task_result["error"]["description"], task_result["error"]["description"])
 
+    def test_sync_with_proxy_auth(self):
+        """Test sync collections from pulp server."""
+        if not os.getenv("PULP_PROXY_TEST"):
+            self.skipTest("Proxy isn't set.")
+
+        second_body = gen_ansible_remote(
+            url=self.distribution.client_url,
+            requirements_file=self.requirements_file,
+            sync_dependencies=False,
+            proxy_url="http://ciproxy:8899",
+            proxy_username="foo",
+            proxy_password="bar",
+        )
+        second_remote = self.remote_collection_api.create(second_body)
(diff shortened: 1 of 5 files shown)


pulp_ansible/tests/functional/api/collection/v3/test_sync.py
```diff
@@ -203,6 +203,54 @@ def test_sync_with_missing_collection(self):
...
+        second_body = gen_ansible_remote(
+            url=self.distribution.client_url,
+            requirements_file=self.requirements_file,
+            sync_dependencies=False,
+            proxy_url="http://127.0.0.1:8899",
+        )
+        second_remote = self.remote_collection_api.create(second_body)
+        self.addCleanup(self.remote_collection_api.delete, second_remote.pulp_href)
+
+        with self.assertRaises(PulpTaskError) as cm:
```
mdellweg, line 245: What's the reason behind the name `cm`? If it's short for "context manager" i would rather call it `ctx` or `exc_ctx` because its the context managed by the context manager `self.assertRaises(PulpTaskError)`.

mdellweg (approved):

pulp_ansible/tests/functional/api/collection/v3/test_sync.py
fao89: I just copied the structure from another test

fao89:
Should we disable this single commit validation?
It disturbs more than helps

pulp_ansible/tests/functional/api/collection/v3/test_sync.py
```diff
@@ -203,6 +203,54 @@ def test_sync_with_missing_collection(self):
         msg = "absent.not_present does not exist"
         self.assertIn(msg, task_result["error"]["description"], task_result["error"]["description"])
 
+    def test_sync_with_proxy_auth(self):
```
bmbouter, line 206: This is really great. We needed some kind of test, any kind of test and you did that.

The challenge we'll have with this test is that it can't run locally without the CI deploying the proxy container. To solve that I propose we start and stop proxy.py (same code) via a pulp-smash pytest fixture. What do you think about that? Is that something you'd be willing to do? If not, maybe I can if that's helpful. I'm hoping we can switch to that style right away before tests like this get copied to more places.

PR #812, opened 2022-01-27
Enable mirror galaxy on nightly job

fao89:
[noissue]

### DIFF
--- a/.github/workflows/scripts/post_before_script.sh
+++ b/.github/workflows/scripts/post_before_script.sh
@@ -22,3 +22,8 @@ cmd_prefix chmod a+x /var/lib/pulp/sign-metadata.sh
 KEY_FINGERPRINT="6EDF301256480B9B801EBA3D05A5E6DA269D9D98"
 TRUST_LEVEL="6"
 echo "$KEY_FINGERPRINT:$TRUST_LEVEL:" | cmd_stdin_prefix gpg --import-ownertrust
+
+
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/post_before_script.sh
```diff
@@ -22,3 +22,8 @@ cmd_prefix chmod a+x /var/lib/pulp/sign-metadata.sh
 KEY_FINGERPRINT="6EDF301256480B9B801EBA3D05A5E6DA269D9D98"
 TRUST_LEVEL="6"
 echo "$KEY_FINGERPRINT:$TRUST_LEVEL:" | cmd_stdin_prefix gpg --import-ownertrust
+
+
+if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
```
bmbouter, line 27: Apologies Fabricio it seems I always have another way to suggest today.

I'd like us to try to do this differently. I want the plugin template to have it's nightly job run tests marked with the `nightly` pytest mark and have the functional tests leave out the `nightly` ones.

bmbouter: What do you think?

fao89: sounds good

bmbouter: Is making this something you want to try or should I? I'm keep asking because I want to get more people involved in this transition.

PR #825, opened 2022-02-01
Add proxy tests

bmbouter:


gerrod3 (changes requested):
Do we need proxy testing for V3 syncing?

.github/workflows/scripts/install.sh
```diff
@@ -75,9 +75,6 @@ services:
     image: "pulp:${TAG}"
     volumes:
       - ./settings:/etc/pulp
-  - name: ciproxy
-    image: ghcr.io/abhinavsingh/proxy.py:latest
-    command: "--basic-auth foo:bar --hostname 0.0.0.0"
```
mdellweg, line 80: Is this the current state of the plugin template? I'd hate to do ping pong with the pulp bot.

pulp_ansible/tests/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+    return RemotesCollectionApi(ansible_bindings_client)
+
+
+# Object Generation Fixtures
+
+
+@pytest.fixture
+def ansible_repo(ansible_repo_api_client, gen_object_with_cleanup):
+    """Creates an Ansible Repository and deletes it at test cleanup time."""
+    return gen_object_with_cleanup(ansible_repo_api_client, gen_repo())
```
mdellweg, line 58: Maybe this should be a factory too.

pulp_ansible/tests/functional/api/collection/v2/test_proxy.py
```diff
@@ -0,0 +1,94 @@
...
+    ansible_remote = gen_ansible_collection_remote(
+        url="https://galaxy.ansible.com",
+        requirements_file="collections:\n  - testing.k8s_demo_collection",
+        sync_dependencies=False,
+        proxy_url=http_proxy.proxy_url,
+    )
+
+    _sync_and_assert(
+        ansible_remote, ansible_repo, ansible_repo_api_client, ansible_collection_version_api_client
+    )
```
mdellweg, line 40: Just wondering, whether this could be handled with `mark.parameterize`. And it's probably a no.

pulp_ansible/tests/functional/api/collection/v2/test_proxy.py
```diff
@@ -0,0 +1,94 @@
...
+)
+
+
+def _sync_and_assert(
+    remote, ansible_repo, ansible_repo_api_client, ansible_collection_version_api_client
+):
+    body = AnsibleRepositorySyncURL(remote=remote.pulp_href)
+    monitor_task(ansible_repo_api_client.sync(ansible_repo.pulp_href, body).task)
+
+    # Check content is present, but no artifacts are there
```
mdellweg, line 15: This comment is wrong.

pulp_ansible/tests/functional/api/collection/v2/test_proxy.py
```diff
@@ -0,0 +1,94 @@
...
+    content_response = ansible_collection_version_api_client.list()
+    assert content_response.count == 1
```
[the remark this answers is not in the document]
bmbouter: Yeah let me fix this, I thought it was for that specific repo. Thanks!

Regarding V3 tests, I don't think they are helpful because the real point of these tests is to assert on the downloaders which is shared by both V2 and V3 tests.

.github/workflows/scripts/install.sh
bmbouter: Ahh, ty. I'll rebase onto it now.

pulp_ansible/tests/conftest.py
bmbouter: I suspect over time end up with both versions, but this is the kind I need now. I like having both options because if need exactly one you don't have to call the factory in your test itself. This is debatable though, more fixtures for less lines of test code...

pulp_ansible/tests/functional/api/collection/v2/test_proxy.py
bmbouter: I'm not understanding the idea.

pulp_ansible/tests/functional/api/collection/v2/test_proxy.py
bmbouter: I will fix, ty.

pulp_ansible/tests/functional/api/collection/v2/test_proxy.py
gerrod3: Probably could be done with parameterize. https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize. Not sure how clean it would be.

bmbouter: I'm familiar with parameterize, but I'm not sure what the we're parameterize with this idea.

gerrod3: The three tests are identical besides the arguments to the remote. Those arguments could be parameterize, even including the case that fails with: https://docs.pytest.org/en/6.2.x/reference.html#pytest.param, resulting with one test that gets ran 3 times. Not sure how cleanly you could make it though...

gerrod3 (approved):

pulp_ansible/tests/functional/api/collection/v2/test_proxy.py
bmbouter: Mmmm I see. Yeah I tried to do something like this actually, but I ended up backing it out. It's a bit tricky because in addition to one of the tests expecting a failure and the other two not, the fixture type used is subtly different too. Two use `http_proxy_with_auth` and the other uses `http_proxy`. The pytest way to paramaterize fixtures is with the `indirect=True` feature, but to do that you have to create an additional fixture that operates kind of like a switcher. I ended up going with this because the other option seemed more complicated. I very much want to use the parameterize though, I haven't really gotten to yet.

pulp_ansible/tests/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+def ansible_remote_collection_api_client(ansible_bindings_client):
+    """Provides the Ansible Collection Remotes API client object."""
+    return RemotesCollectionApi(ansible_bindings_client)
+
+
+# Object Generation Fixtures
+
+
+@pytest.fixture
+def ansible_repo(ansible_repo_api_client, gen_object_with_cleanup):
```
mdellweg, line 56: gen_ansible_repo ?

mdellweg: > gen_ansible_repo ?

edit: Forget about this! I read incorrectly on my mobile. Small screen... Didn't see all comments.

pulp_ansible/tests/conftest.py
mdellweg: Yeah, it's a matter of building habits. So no require multiple repos here.

mdellweg (approved):

PR #831, opened 2022-02-03
Updates CI template

bmbouter:
Also mark the mirror_galaxy V2 sync test as a 'nightly' test.

[noissue]


fao89 (approved):

mdellweg (approved):
I like this change. But hand on heart, do we really realize a failing nightly?

bmbouter:
> I like this change. But hand on heart, do we really realize a failing nightly?

I get emails, but I don't always act on them same day. We should think more about how we can reduce the time to fix of breaking nightlies.

bmbouter:
This PR is working well because I can see `test_mirror_galaxy` was not collected whereas before it was collected but shown as skipped!

bmbouter:
I kicked off a nightly here post-merge, so we'll see if `test_mirror_galaxy` shows up there: https://github.com/pulp/pulp_ansible/actions/runs/1795849960

fao89:
you will see it being skipped again, the event is dispatch, and your change only works for schedule

bmbouter:
Ha! @fao89 so do you think this will work as expected when the actual nightly occurs? Also is there a better way for me to have this run under both conditions?

fao89:
you would have to add something like `or event_name == dispatch`
and I believe it will work because it is the way we do on the single container: 
https://github.com/pulp/pulp-oci-images/blob/latest/.github/workflows/publish_images.yaml#L27

bmbouter:
I'm a little unclear on the syntax. @fao89 would you be able to open a pull request maybe?

fao89:
here: https://github.com/pulp/pulp_ansible/pull/832

PR #834, opened 2022-02-07
Fix nightly mirror test

fao89:
[noissue]


mdellweg (approved):
That sounds reasonable

Issue #837, opened 2022-02-11
Allow upload of CollectionVersionSignatures via API


rochacbruno:
**Is your feature request related to a problem? Please describe.**

To approve collections on automation hub it will be required that the collection has at least one valid signature attached to it, users will be able to upload a `namespace.collection-version.asc` file containing the ascii armoured detached signature file.

**Describe the solution you'd like**

An endpoint/viewset to enable upload of CollectionVersionSignature

```yaml
POST: https://pulp/pulp/api/v3/content/ansible/collection_signatures/
---
signature_file: path/to/namespace.collection-version.asc
signed_collection: reference to CollectionVersion being signed
repository:  name of the repository where that signature must be added
```

The content of the `namespace.collection-version.asc` must feed the `.data` attribute of the collection signature object.
signed_collection gets the relation with existing collection version on the same `repository` that is informed.

**validation**

The endpoint must have a `validate` method that on the default implementation can do a simple validation such as matching the artifact name `namespace.collection-version.asc` against `namespace.collection-version.tar.gz` or just ensuring the file isn't empty.

On galaxy_ng we are going to subclass this ViewSet/Serializer and on the validate we want to add a GPG verification process so we avoid upload of invalid signatures, we plan to do that via settings

e.g:

```py
GALAXY_VERIFICATION_KEYRING="/path/to/galaxy.kbx"
```

And then using GPG we are going to verify the signature + collection against the keys on that keyring, at this point I am not sure if this specific verification logic can go to pulp_ansible source or better to keep on galaxy side.

gerrod3:
### Investigation into adding signature upload support
The current fields on the signature model include:
- signed_collection - the collection-version this signature points to
- data - the actual signature itself
- digest - the sha256 digest of the signature
- pubkey_fingerprint - the fingerprint of the public key used to create this signature
- signing_service - an optional signing service used to create this signature

The signature's uniqueness is determined by the `signed_collection` and `pubkey_fingerprint`, so these two pieces of information need to also be provided when uploading a signature.  Currently the signatures produced by the signing task use a collection-version's `MANIFEST.json` by passing it to the `signing_service` that the user supplies. The `signing_service` contains a user shell script that can use any cryptographic signature program to return a signature file.

### Some Key questions that arise from the current design:
1. Do we intend to only support(expect) OpenPGP signatures?
2. Do we care if the signatures are detached or encrypted?
3. What filename and format types should we expect?

### Option 1, No Convenience:
[...]

PR #838, opened 2022-02-15
Add CollectionSignatures upload

gerrod3:
fixes: #837


pulp_ansible/app/serializers.py
```diff
@@ -515,13 +515,31 @@ class CollectionVersionSignatureSerializer(NoArtifactContentSerializer):
...
     )
 
+    def validate(self, data):
+        """
+        Prepare the necessary data for signature creation. Minimal validation is performed.
+        """
+        data = super().validate(data)
+
+        if "file" not in data:
+            raise ValidationError(_("'file' must be specified."))
```
mdellweg, line 529: Can't we just mark the file filed as required, and let drf do the tedious part?

rochacbruno: line 532 explains why this can't rely on DRF, the validate is called outside the request context a second time.

Maybe this `if` must go inside that `if ¨request"` condition.

rochacbruno (approved):
LGTM

I am going to rewrite validation on galaxy_ng

gerrod3:
@rochacbruno I still would like to add 1 test for the upload. 

Are you sure you want to merge it with this current design? We can change it to be more in line with what you are currently doing(planning) on the galaxy_ng side?

rochacbruno:
@gerrod3 It would be good to add the validation builtin to pulp_ansible, would that work for the upstream?

The plan is:

- There will be a keyring `/var/lib/pulp/keys/galaxy.kbx` defined on `settings.COLLECTION_UPLOAD_VERIFICATION_KEYRING`
- That keyring will store the public key only
- That keyring will be used to verify considering whatever key is inside there (if multiple keys at least one must verify)

One options is to specify the key on the settings, but this is not flexible in case we need to replace that key or add more.

I also considered calling the `ansible-galaxy cli --verify` directly so we would use the same validation as the CLI and they use similar keyring approach.

For making it flexible we could or add a new model such as `VerificationService` or a settings key e.g: `settiings.COLLECTION_VERIFICATION_SCRIPT=/path/to/script.sh` 

What do you think is the best?

If that seems too much for now, I am ok on customizing all on galaxy side and we iterate later on it.

gerrod3:
@rochacbruno I think using the `ansible-galaxy cli --verify` would be the best approach to consolidate the code and ensure it works properly for the end user. Has that feature already been implemented into the galaxy cli and is there documentation for how I can use it?

rochacbruno:
@gerrod3 

We can specify `--keyring` so we could read that from `settings.COLLECTION_UPLOAD_VERIFICATION_KEYRING` and then do
```bash
ansible-galaxy collection verify my_namespace.my_collection \
    --signature file:///path/to/local/detached_signature.asc \
    --keyring /var/lib/pulp/keys/galaxy.kbx
```

pulp_ansible/app/serializers.py
```diff
@@ -515,13 +515,33 @@ class CollectionVersionSignatureSerializer(NoArtifactContentSerializer):
...
         view_name="signing-services-detail",
-        queryset=SigningService.objects.all(),
+        read_only=True,
         allow_null=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        """Ensure `file` field is required."""
+        super().__init__(*args, **kwargs)
+        self.fields["file"].required = True
```
mdellweg, line 525: I know this "works", but can you check in the api schema that it is really marked as required. If not, i think you need to redefine the field on this class.

pulp_ansible/app/tasks/signature.py
```diff
@@ -22,6 +23,34 @@
 from pulpcore.plugin.models import SigningService, ProgressReport
 from pulpcore.plugin.sync import sync_to_async_iterable, sync_to_async
 from pulp_ansible.app.tasks.utils import get_file_obj_from_tarball
+from rest_framework import serializers
+
+
+def verify_signature_upload(data):
+    """The task code for verifying collection signature upload."""
```
gerrod3, line 30: We would like to use the `ansible-galaxy colection verify` command, but I think it is [bugged](https://github.com/ansible/ansible/issues/77146), so this is a backup for right now.

pulp_ansible/app/tasks/signature.py
```diff
@@ -22,6 +23,38 @@
...
+    if verified.trust_level is None or verified.trust_level < verified.TRUST_FULLY:
+        # Skip verification if repository isn't specified, or it doesn't have a keyring attached
+        if verified.fingerprint is None or keyring is not None:
+            raise serializers.ValidationError(
+                _("Signature verification failed: {}").format(verified.status)
+            )
+
+    data["data"] = sig_data
+    data["digest"] = file.hashers["sha256"].hexdigest()
+    data["pubkey_fingerprint"] = verified.fingerprint
```
mdellweg, line 56: Ahh, slick. This is why we can make that field read only...
But it will be None, if we hit the condition in line 49. Would it be better if we could fall back to validation against a global keyring there?
Consider this:
1. you upload the signature; no repo -> kfp=None
2. you add the signature to a repo with a keyring -> kfp=None, because content is immutable
But it should create the same result as
1. you upload the signature to repo with keyring -> kfp is not None

Alternatively only allow upload in cases we can actually verify the sig.

gerrod3: The comment I left isn't very indicative of the actual behavior of the check. `verified.fingerprint` will always be set if the signature is a real PGP signature. Even if the signature isn't valid because it's expired or for the wrong file and even if the keyring doesn't contain the public key of the signature the field would be set. `verified.fingerprint` should only be `None` if the signature is beyond messed up and the check on line 49 would ensure that a `ValidationError` is raised. Also, if the keyring isn't set gnupg will use the default keyring from the default homedir of GPG, usually `~/.gnupg/pubring.kbx`. Also, one last thing, `pubkey_fingerprint` can not be set to `None`, it will result in a database error upon creation. So even if it gets passed my checks somehow, the user still wouldn't be able to create a signature that doesn't have a `pubkey_fingerprint`.

mdellweg (approved):

goosemania:
@rochacbruno , are we good to merge and release or are there potentially more changes needed?

rochacbruno (approved):
LGTM

pulp_ansible/app/tasks/signature.py
```diff
@@ -22,6 +23,38 @@
...
+def verify_signature_upload(data):
+    """The task code for verifying collection signature upload."""
+    file = data["file"]
+    sig_data = file.read()
+    file.seek(0)
+    collection = data["signed_collection"]
+    keyring = None
+    if (repository := data.get("repository")) and repository.keyring:
+        keyring = repository.keyring
+    gpg = gnupg.GPG(keyring=keyring)
```
rochacbruno, line 38: looks like this line should have been 
```py
gpg = gnupg.GPG(gnupghome=settings.ANSIBLE_CERTS_DIR, keyring=keyring)
```

@gerrod3

mdellweg: Can we please try to not make the gpg operations plugin specific? This will end in a configuration nightmare.
Maybe we should even provide a `pulpcore.plugin` interface that wraps around `gnupg.GPG` to make it consistent for plugins.
Plugins to bring to the table include ansible, container, deb and rpm at this time.

Issue #841, opened 2022-02-17
Signing task doesn't properly use the MANIFEST.json file for the signature creation.


gerrod3:
**Version**
pulpcore 3.17+, ansible- 0.12.0

**Describe the bug**
The current implementation of the signature task doesn't properly extract the current MANIFEST.json file from the collection tarball and use that to create that collection's signature. Instead it uses the manifest field on the collection-version's model and dumps the json to a temporary file and signs that. However, the json dumping of the manifest field could create a file with completely different formatting compared to the original MANIFEST.json file and thus produce an invalid signature. The signature task should use the exact MANIFEST.json file that the ansible-galaxy-cli will use for its verification.

PR #843, opened 2022-02-23
Fix improper signatures

gerrod3:
fixes: #841

### DIFF
--- a/functest_requirements.txt
+++ b/functest_requirements.txt
@@ -2,3 +2,4 @@ pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 orionutils
 pytest
 dynaconf
+https://github.com/ansible/ansible/archive/devel.tar.gz
--- a/pulp_ansible/app/tasks/signature.py
+++ b/pulp_ansible/app/tasks/signature.py
@@ -1,7 +1,7 @@
 import aiofiles
 import asyncio
-import json
 import hashlib
+import tarfile
 from gettext import gettext as _
 
 from pulpcore.plugin.stages import (
@@ -18,8 +18,10 @@ from pulp_ansible.app.models import (
 )
 
 from django.conf import settings
+from django.core.files.storage import default_storage as storage
 from pulpcore.plugin.models import SigningService, ProgressReport
 from pulpcore.plugin.sync import sync_to_async_iterable, sync_to_async
+from pulp_ansible.app.tasks.utils import get_file_obj_from_tarball
 
 
 def sign(repository_href, content_hrefs, signing_service_href):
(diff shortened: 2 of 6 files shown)


functest_requirements.txt
```diff
@@ -2,3 +2,4 @@ pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 orionutils
 pytest
 dynaconf
+ansible @ git+https://github.com/ansible/ansible.git@devel
```
gerrod3, line 5: The installation signature verification is currently only available on the devel branch.

pulp_ansible/app/tasks/signature.py
```diff
@@ -71,13 +73,20 @@ async def sign_collection_version(self, collection_version):
...
-                async with aiofiles.open(f"{d}/MANIFEST.json", "w") as t:
-                    await t.write(json.dumps(collection_version.manifest))
-                result = await self.signing_service.asign(t.name)
-                async with aiofiles.open(result["signature"], "rb") as sig:
-                    data = await sig.read()
+            # OpenPGP doesn't take filename into account for signatures, not sure about others
+            async with aiofiles.tempfile.NamedTemporaryFile(dir=".", mode="wb") as m:
+                cartifact = await sync_to_async(
+                    collection_version.contentartifact_set.select_related("artifact").first
+                )()
```
gerrod3, line 80: Not sure if there is a better way to get this content_artifact to extract the artifact.

mdellweg: No, this looks right to me.

pulp_ansible/app/tasks/signature.py
```diff
@@ -71,13 +73,20 @@ async def sign_collection_version(self, collection_version):
...
+                artifact_name = cartifact.artifact.file.name
+                artifact_file = storage.open(artifact_name)
+                with tarfile.open(fileobj=artifact_file, mode="r") as tar:
+                    man = get_file_obj_from_tarball(tar, "MANIFEST.json", artifact_name)
+                    await m.write(man.read())
```
mdellweg, line 85: Should we do some more sync_to_async around tar file operations?

mdellweg (commented):
I am wondering, whether we should verify all file digests in the manifest before actually signing it.

rochacbruno (approved):

functest_requirements.txt
mdellweg: To be upfront: This is only a test requirement. So we are fine with relying on some unreleased piece of code.

gerrod3:
> I am wondering, whether we should verify all file digests in the manifest before actually signing it.

We could, it would slow down the task but maybe it is something to look into for the future. I'm not sure I can look into adding it now.

Issue #840, opened 2022-02-17
Manifest field isn't set when uploading a collection


gerrod3:
**Version**
core-3.18.dev, ansible-0.13.dev

**Describe the bug**
When using the ansible collection-versions upload api (POST `/ansible/collections/`) the manifest field on the resulting collection-version is blank.

**To Reproduce**
`pulp ansible content upload --file test.collection-1.0.0.tar.gz`
`pulp ansible content list --namespace collection --name test --fields manifest`

**Additional context**
There's 4 different ways to upload a collection to Pulp and I'm not sure if this bug applies to all of them. More investigation is needed.

PR #847, opened 2022-03-02
Fix manifest and files field on collection upload

gerrod3:
fixes: #840

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -302,6 +302,13 @@ def import_collection(
 
     try:
         with temp_file.file.open() as artifact_file:
+            with tarfile.open(fileobj=artifact_file, mode="r") as tar:
+                manifest_data = json.load(
+                    get_file_obj_from_tarball(tar, "MANIFEST.json", temp_file.file.name)
+                )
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -302,13 +302,23 @@ def import_collection(
...
+                )
             url = _get_backend_storage_url(artifact_file)
             importer_result = process_collection(
                 artifact_file, filename=filename, file_url=url, logger=user_facing_logger
             )
             artifact = Artifact.from_pulp_temporary_file(temp_file)
             importer_result["artifact_url"] = reverse("artifacts-detail", args=[artifact.pk])
             collection_version = create_collection_from_importer(importer_result)
+            collection_version.manifest = manifest_data
+            collection_version.files = files_data
```
mdellweg, line 320: Did they not get set at all, or are the results of `create_collection_from_importer` wrong?

mdellweg (approved):

pulp_ansible/app/tasks/collections.py
gerrod3: They don't get set at all.

PR #849, opened 2022-03-07
Galaxy API Refactor stage 1

newswangerd:
This PR does the following:
- moves the existing collection views under `/plugin/ansible/`
- redirects the legacy v3 endpoints to their counterparts in `/plugin/ansible/`
- adds a new configuration option `ANSIBLE_DEFAULT_DISTRIBUTION_PATH` that allows users to configure a default distribution base path for the API
- adds a new configuration option `ANSIBLE_URL_NAMESPACE` that allows django URL namespace to be set on `reverse` so that urls can be configured to point correctly to the galaxy APIs when pulp ansible is deployed as part of automation hub
- adds the get `v3/artifacts/path/file` API endpoint from galaxy_ng
- adds the `ContentRedirectContentGuard` content guard from galaxy_ng

#728


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -560,12 +631,123 @@ class RepoMetadataViewSet(
...
+
+                kwargs = {**self.kwargs, "distro_base_path": path}
+
+            # don't pass request. redirects work with just the path and this solves the client
+            # redirect issues for aiohttp
+            url = reverse_lazy(
+                settings.ANSIBLE_URL_NAMESPACE + self.url,
+                kwargs=kwargs,
+                # request=self.request,
+            )
```
fao89, line 721: This is the redirect fix

pulp_ansible/app/galaxy/views.py
```diff
@@ -28,12 +28,18 @@ class GalaxyVersionView(views.APIView):
     authentication_classes = []
     permission_classes = []
 
-    def get(self, request, path):
+    v3_only = False
```
bmbouter, line 31: Is this expected to still be here?

pulp_ansible/app/models.py
```diff
@@ -325,3 +327,7 @@ class AnsibleDistribution(Distribution):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+
+
+def _gen_secret():
```
bmbouter, line 332: What's this for?

pulp_ansible/app/settings.py
```diff
@@ -24,3 +24,5 @@
 ANSIBLE_CONTENT_HOSTNAME = settings.CONTENT_ORIGIN + "/pulp/content"
 ANSIBLE_SIGNING_TASK_LIMITER = 10
 ANSIBLE_CERTS_DIR = "/etc/pulp/certs/"
+ANSIBLE_DEFAULT_DISTRIBUTION_PATH = None
+ANSIBLE_URL_NAMESPACE = ""
```
bmbouter, line 28: We need to have the docs updated with these settings prior to merging. https://docs.pulpproject.org/pulp_ansible/settings.html https://github.com/pulp/pulp_ansible/blob/main/docs/settings.rst

pulp_ansible/app/settings.py
```diff
@@ -24,3 +24,5 @@
 ANSIBLE_CONTENT_HOSTNAME = settings.CONTENT_ORIGIN + "/pulp/content"
 ANSIBLE_SIGNING_TASK_LIMITER = 10
 ANSIBLE_CERTS_DIR = "/etc/pulp/certs/"
```
bmbouter, line 26: @gerrod3 would you be willing to open a PR to update the docs for these also maybe?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -165,12 +171,14 @@ class CollectionViewSet(
     ViewSet for Collections.
     """
 
-    authentication_classes = []
-    permission_classes = []
```
mdellweg, line 169: This will make the views accessible for admin only, right? Unless we inject an access policy stating otherwise.

pulp_ansible/app/models.py
mdellweg: I think, a leftover of the redirect content guard. Should be removed.

newswangerd: oh woops. I missed that when I removed the content redirect guard.

pulp_ansible/app/galaxy/v3/views.py
bmbouter: That is my understanding.

pulp_ansible/app/galaxy/views.py
bmbouter: I think it's still confused about why it needs to be switchable yet I don't see where `v3_only` is ever set to `True`.

newswangerd: @bmbouter https://github.com/pulp/pulp_ansible/pull/849/files#diff-a18275319f35fc1fbea4b63cd3e4a3b8fc68d4a43e51f7702d39818c705d1e2aR212

bmbouter:
@newswangerd should these new settings be variables on the installer we document here? https://docs.pulpproject.org/pulp_ansible/installation.html#install-using-ansible

bmbouter:
@gerrod3 same question for the new variable you were providing re the install docs https://docs.pulpproject.org/pulp_ansible/installation.html#install-using-ansible

pulp_ansible/app/galaxy/views.py
newswangerd: The v1 and v2 APIs all require a distro base path in the url for the entire API, however the v3 API is designed for the base path to be nested further down in URL path. For example a v2/ legacy v3 path would look like `/pulp_ansible/galaxy/{path}/api/v3/collections` whereas the new v3 paths look like `/pulp_ansible/galaxy/default/api/v3/plugin/ansible/content/{path}/`. Since v1 and v2 both require a path at the root, it doesn't make sense to serve them from `/pulp_ansible/galaxy/default/`.

pulp_ansible/app/galaxy/v3/views.py
newswangerd: That's correct

newswangerd:
> @newswangerd are there updates to how we instruct users to configure their galaxy CLI? Those sections are here https://docs.pulpproject.org/pulp_ansible/workflows/roles.html#install-a-role-hosted-by-pulp-using-ansible-galaxy and here https://docs.pulpproject.org/pulp_ansible/workflows/collections.html#install-a-collection-hosted-by-pulp-using-ansible-galaxy

I added information about configuring auth

pulp_ansible/app/galaxy/v3/views.py
mdellweg: Don't we need to provide a default permissive access policy then to keep pulp_ansible upstream happy until we implement RBAC there?

newswangerd:
@bmbouter @mdellweg please re-review

pulp_ansible/app/galaxy/v3/views.py
newswangerd: @mdellweg @bmbouter  I'm more than happy to add these. Just let me know if you want them.

mdellweg: I believe we need them in order not to change current behaviour. And i think they should be like allow everything. Also galaxy overrides them anyway, right?
I just want a second opinion before i make it a requirement.

docs/workflows/collections.rst
```diff
@@ -256,22 +237,53 @@ config (e.g. ``~/.ansible.cfg`` or ``/etc/ansible/ansible.cfg``):
 .. code::
 
     [galaxy]
-    server: http://localhost/pulp_ansible/galaxy/my_content/
+    server_list = pulp
```
mdellweg, line 240: Is deprecated missing here on purpose?

Or asked the other way around: Can you still specify `-s deprecated` if it is not included here? Just reading the docs linked above didn't enlighten me about that.

docs/workflows/collections.rst
```diff
@@ -256,22 +237,53 @@ config (e.g. ``~/.ansible.cfg`` or ``/etc/ansible/ansible.cfg``):
...
+    password = my_password
 
-Then you can install without the ``-s`` url
+    [galaxy_server.deprecated]
+    server = http://localhost/pulp_ansible/galaxy/deprecated/
+    username = my_user
+    password = my_password
+
+Servers will be used in the order that they appear in ``server_list`` in the configuration file. Specific servers 
+can also be specified using the ``-s`` flag, either by their name or URL.
```
mdellweg, line 253:
```suggestion
can also be specified using the ``-s`` (or ``--server``) flag, either by their name or URL.
```

docs/workflows/collections.rst
```diff
@@ -256,22 +237,53 @@ config (e.g. ``~/.ansible.cfg`` or ``/etc/ansible/ansible.cfg``):
...
+
+Using a specific version
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Install your collection by name by specifying the distribution serving your Repository's content using the
+``-s`` option.
+
+``ansible-galaxy collection install testing.ansible_testing_content:4.0.6 -c``
+
+
```
mdellweg, line 286: This whole section confused me. Also i think it is misplaced below the `collection-publish` anchor.

requirements.txt
```diff
@@ -4,6 +4,6 @@ galaxy_importer>=0.4.1
 GitPython>=3.1.24
 jsonschema>=3.0
 packaging
-pulpcore>=3.16
+pulpcore>=3.18
```
mdellweg, line 7: Can you pinpoint a change that makes this necessary?

mdellweg (commented):
Sorry, so many questions again.

pulp_ansible/app/galaxy/v3/views.py
bmbouter: After some matrix convo, I believe we're all in agreement these lines should be removed.

docs/workflows/collections.rst
mdellweg: I should elaborate a bit. To me the command examples are not reflecting what the text around promises. Also what is the `-c`?

requirements.txt
newswangerd: It's because of this: https://github.com/pulp/pulp_ansible/pull/849/files#diff-e11a4087ff75182a1c5c58b3812e77def3dccfe96c12e2f6261760225d5c41d2R470. We're using the new content redirect guard.

We don't _technically_ need to bump the version the way that it's written now, but this should probably be re-written to import `RedirectContentGuard` and then compare using `if guard.TYPE == RedirectContentGuard.TYPE` instead of `if guard.TYPE == "content_redirect"`

newswangerd: @mdellweg lmk if we should leave it at 3.16 and keep the content guard comparison as is, or bump to 3.18 and import the content guard

docs/workflows/collections.rst
newswangerd: Yes. This demonstrates that the `deprecated` server won't be used by default, but you can still specify `-s dreprecated` if you need to download something from there. I'll clarify.

requirements.txt
mdellweg: Thank you. That is exactly the information i wanted to hear (i forgot about).
Keep it exactly the way it is.

docs/workflows/collections.rst
mdellweg: Thank you. We will wind up with better docs than over there...

docs/workflows/collections.rst
```diff
@@ -256,18 +257,34 @@ config (e.g. ``~/.ansible.cfg`` or ``/etc/ansible/ansible.cfg``):
...
+    [galaxy_server.deprecated]
+    server = http://localhost/pulp_ansible/galaxy/deprecated/
+    username = my_user
+    password = my_password
+
+Servers will be used in the order that they appear in ``server_list`` in the configuration file. Specific servers 
+can also be specified using the ``-s`` (or ``--server``) flag, either by their name or URL. In this example
+``ansible-galaxy`` will attempt to fetch content from ``pulp`` and fallback to ``my_fallback_server`` if it can't
+find anything. The ``deprecated`` server is not listed under ``server_list``, so ``ansible-galaxy`` won't pull
+content from it unless the server is specified explicitly with the ``-s`` flag.
```
mdellweg, line 279: :+1:

mdellweg (approved):
I think we have it all now.

CHANGES/728.feature
```diff
@@ -0,0 +1,10 @@
+Galaxy API Refactor stage 1
+
+- move the existing collection views under /plugin/ansible/
+- redirects the legacy v3 endpoints to their counterparts in /plugin/ansible/
+- adds a new configuration option ANSIBLE_DEFAULT_DISTRIBUTION_PATH that allows users to configure a default distribution base path for the API
+- adds a new configuration option ANSIBLE_URL_NAMESPACE that allows django URL namespace to be set on reverse so that urls can be configured to point correctly to the galaxy APIs when pulp ansible is deployed as part of automation hub
+- adds the get v3/artifacts/path/file API endpoint from galaxy_ng
+- enable RedirectContentGuard
+
+fixes: #728
```
bmbouter, line 10: This part doesn't go in the towncrier portion. Towncrier will automatically link it from the filename.

bmbouter: The `fixes ... ` text does need to stay in the commit tho.

docs/settings.rst
```diff
@@ -44,3 +44,18 @@ GALAXY_API_ROOT
...
+
+   Set the distribution base path to be used on the ``GALAXY_API_ROOT/default/api/`` endpoint.
+   By default, this is set to ``None``, which causes the API to return a 404 on the ``default``
+   endpoint.
+
+
+ANSIBLE_URL_NAMESPACE
+^^^^^^^^^^^^^^^^^^^^^
+
+   The Django URL namespace to be used when generating URLs that are returned by the galaxy
```
bmbouter, line 60: Is this a required config or does it have a default? If it's a required config we need to be sure the application refuses to start if it's unconfigured. We usually put those checks in the `settings.py` and raise the dynaconf ImproperlyConfigured if so.

pulp_ansible/app/settings.py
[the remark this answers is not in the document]
bmbouter: It doesn't look like we need additional validation though.

bmbouter (approved):
I left a few small docs clarifications to be updated pre-merge, but I'm going ahead and lgtm-ing now.

PR #854, opened 2022-03-11
FIx Changelog job

fao89:
[noissue]


fao89:
merging for testing plugin-template fix

Issue #879, opened 2022-03-29
Move the collection/collection version deletion endpoints over from galaxy_ng


newswangerd:
The only part of the CollectionViewset and CollectionVersionViewset that aren't currently supported in pulp ansible is the delete method. These methods and their corresponding tasks need to be moved over to pulp ansible:
- https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/viewsets/collection.py#L108
- https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/viewsets/collection.py#L186
- https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/tasks/deletion.py#L67
- https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/tasks/deletion.py#L45

PR #880, opened 2022-03-31
Add api endpoints to delete collections and collection versions.

newswangerd:
fixes: #879


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -359,6 +362,76 @@ def update(self, request, *args, **kwargs):
...
+
+
+def get_collection_dependents(parent):
+    """Given a parent collection, return a list of collection versions that depend on it."""
+    key = f"{parent.namespace}.{parent.name}"
+    dependents = []
+    for child in CollectionVersion.objects.exclude(collection=parent).filter(
+        dependencies__has_key=key
+    ):
+        dependents.append(child)
```
mdellweg, line 421:
```suggestion
    dependents = list(
        CollectionVersion.objects.exclude(collection=parent).filter(
            dependencies__has_key=key
        )
    )
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -359,6 +362,76 @@ def update(self, request, *args, **kwargs):
...
+    return dependents
+
+
+def get_dependents(parent):
+    """Given a parent collection version, return a list ofcollection versions that depend on it."""
```
mdellweg, line 426:
```suggestion
    """Given a parent collection version, return a list of collection versions that depend on it."""
```

pulp_ansible/app/tasks/deletion.py
```diff
@@ -0,0 +1,75 @@
...
+    2. Run orphan_cleanup to delete the CollectionVersion
+    3. Delete Collection if it has no more CollectionVersion
+    """
+    collection_version = CollectionVersion.objects.get(pk=collection_version_pk)
+    collection = collection_version.collection
+
+    _remove_collection_version_from_repos(collection_version)
+
+    log.info("Running orphan_cleanup to delete CollectionVersion object and artifact")
+    orphan_cleanup(content_pks=None, orphan_protection_time=10)
```
mdellweg, line 52: That number seems rather arbitrary. Also we should really specify the content pk to remove.

mdellweg (changes requested):

pulp_ansible/app/tasks/deletion.py
newswangerd: @rochacbruno do you know why you had this set to `orphan_protection_time=10`?

rochacbruno: @newswangerd I haven't, this came from this PR https://github.com/ansible/galaxy_ng/pull/1163

rochacbruno: @newswangerd That is to prevent race conditions

pulp_ansible/app/galaxy/v3/views.py
rochacbruno: This was created using a `for` to allow easier interactive debugging

pulp_ansible/app/tasks/deletion.py
```diff
@@ -0,0 +1,78 @@
...
+    for version in repo.versions.complete().order_by("-number")[1:]:
+        version.delete()
+
+
+def _remove_collection_version_from_repos(collection_version):
+    """Remove CollectionVersion from latest RepositoryVersion of each repo."""
+    for repo in collection_version.repositories.all():
+        add_and_remove(repo.pk, add_content_units=[], remove_content_units=[collection_version.pk])
+        _cleanup_old_versions(repo)
+    collection_version.save()
```
mdellweg, line 35: Content is immutable. We should not attempt to save it.

rochacbruno: @mdellweg can't we make `.save` on content to raise error when object already exists?

mdellweg: Why should it raise an error? It would just see if ther is the need for a update in the database. I would look if it had a PK. In this case we are about to delete that thing. We can be pretty sure it exists. No?

pulp_ansible/app/tasks/deletion.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    collection_version = CollectionVersion.objects.get(pk=collection_version_pk)
+    collection = collection_version.collection
+
+    _remove_collection_version_from_repos(collection_version)
+
+    log.info("Running orphan_cleanup to delete CollectionVersion object and artifact")
+    # Running orphan_protection_time=0 should be safe since we're specifying the content
+    # to be deleted. This will prevent orphan_cleanup from deleting content that is in
+    # the process of being uploaded.
```
mdellweg, line 53: TBH: This is the usecase, we implemented the `content_pks` parameter for.

pulp_ansible/app/tasks/deletion.py
```diff
@@ -0,0 +1,82 @@
...
+    orphan_cleanup(content_pks=[collection_version.pk], orphan_protection_time=0)
+
+    if not collection.versions.all():
```
mdellweg, line 56:
```suggestion
    if not collection.versions.exists():
```

Note to myself: This only cleans up if this was the last existing version of that collection.

mdellweg (approved):

pulp_ansible/app/tasks/deletion.py
```diff
@@ -0,0 +1,82 @@
...
+    1. Call _remove_collection_version_from_repos
+    2. Run orphan_cleanup to delete the CollectionVersion
+    3. Delete Collection if it has no more CollectionVersion
+    """
+    collection_version = CollectionVersion.objects.get(pk=collection_version_pk)
+    collection = collection_version.collection
+
+    _remove_collection_version_from_repos(collection_version)
+
+    log.info("Running orphan_cleanup to delete CollectionVersion object and artifact")
```
bmbouter, line 50: All logging in pulp_ansible should be i8ln wrapped with gettext

bmbouter: @newswangerd would you possibly we willing to fix the log statements?

pulp_ansible/tests/functional/api/collection/v3/test_deletion.py
```diff
@@ -0,0 +1,194 @@
+from pulp_ansible.tests.functional.utils import gen_collection_in_distribution
+from pulp_ansible.tests.functional.utils import SyncHelpersMixin, TestCaseUsingBindings
+
+from pulpcore.client.pulp_ansible.exceptions import ApiException
+from pulp_smash.pulp3.bindings import monitor_task
+
+
+class CollectionDeletionTestCase(TestCaseUsingBindings, SyncHelpersMixin):
```
bmbouter, line 8: Nice tests!

pulp_ansible/tests/functional/utils.py
```diff
@@ -163,6 +167,52 @@ def wait_tasks():
         running_tasks = tasks.list(state="running")
 
 
+def gen_collection_in_distribution(
+    base_path, name=None, namespace=None, versions=None, **cfg_kwargs
+):
+    """
+    Generate a randomized collection and upload it to the specified repository.
+
+    Params:
```
bmbouter, line 176: This isn't a big deal, but Pulp's docstring style all over the code is different than this. https://docs.pulpproject.org/pulpcore/contributing/style-guide.html?highlight=docstring#in-code-documentation

bmbouter:
I'm really glad to see that this convergence between our two projects. Thank you!

pulp_ansible/app/tasks/deletion.py
rochacbruno: @bmbouter internationalization on logging is not recommended, logging must be consistent regardless of the language of the system. As logs are not for end user it would be better to keep in English only.

pulp_ansible/app/tasks/deletion.py
rochacbruno: you said it is immutable, if it is immutable I should not be allowed to call mutating methods.

pulp_ansible/app/tasks/deletion.py
bmbouter: I'm not familiar with that perspective, and I've always heard the reverse. If someone can't understand the logs how could they interpret them? How can we sort this out because if you look through Pulp's existing code it strives to i8ln all strings including log messages. Can you educate me some more on this?

rochacbruno: i18n is for front-ends, messages for the end user must be translated such as in API errors, CLI output, UI components.

Log audience are system administrators, auditing, developer when troubleshooting, if a support/qe/dev need to analyze a log taken from a customer running in French, the logs should be in English, consistent with the development language.

rochacbruno: @bmbouter imagine you take a bug issue from a customer running a system in Chinese, customer attaches log output. Isn't better to have log in English?

bmbouter: I hear you on the perspective, but right now we've got two perspectives. What evidence and guidance can we look at to help guide us on this?

rochacbruno: you can even search codebase to find where log where printed

bmbouter: Ultimately who are logs really for though? I really think rather than debating one over the other we should try to find out what other large, multi-lingual apps do. What does the python community expect. How about GNU?

bmbouter: Can you please bring this to discourse.pulpproject.org/ or to [open floor](https://hackmd.io/SVCMjpwXTfOMqF2OeyyLRw) because if what you're saying is right we should be doing this differently in all of Pulp. That's a good maturity point for an application, but we shouldn't really debate it here. Also, from a consistency perspective this is inconsistent to how it's been done broadly to date.

rochacbruno: @bmbouter the utility of logs are for debugging, auditing, monitoring, usually those professionals prefer English so it is consistent among different applications.

Log levels for example are "ERROR" or "WARNING" standardized and not "ERRO" "AVISO" if I switch my app front end to Portuguese.

rochacbruno: even log viewers tools filters by messages like ERROR, CRITICAL and there are tools that can do other kinds of filtering and analysis using English as base.

bmbouter: Maybe I haven't been clear, I see your point. Convincing me does very little though. Can you help me bring this to the larger audience of 20+ Pulp devs? Also can we stay consistent with the current style until we have a broad understanding that it's changed?

ipanova: +1 to bring to a broader audience.

Issue #909, opened 2022-04-12
500 when ansible-client acceses distribution that does not point to repository


ipanova:
**Version**
pulp-ansible 0.14

**Describe the bug**

**To Reproduce**
```
$ pulp ansible distribution create --name "test" --base-path "test"
Started background task /pulp/api/v3/tasks/6cc0bcd8-e767-47ec-85a4-fc229ba72c6c/
Done.
{
  "pulp_href": "/pulp/api/v3/distributions/ansible/ansible/66fbcc7d-4f9a-4dcd-ac0d-f917b1d70716/",
  "pulp_created": "2022-04-12T09:43:01.329738Z",
  "base_path": "test",
  "content_guard": null,
  "name": "test",
  "repository": null,
  "repository_version": null,
  "client_url": "https://pulp3-source-fedora34.fluffy.example.com/pulp_ansible/galaxy/test/",
  "pulp_labels": {}
}

(pulp) [vagrant@pulp3-source-fedora34 pulpcore]$ ansible-galaxy install -vv elasticsearch,6.2.4 -c -s https://pulp3-source-fedora34.fluffy.example.com/pulp_ansible/galaxy/test/
ansible-galaxy [core 2.12.4]
  config file = None
  configured module search path = ['/home/vagrant/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/pulp/lib64/python3.9/site-packages/ansible
  ansible collection location = /home/vagrant/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/lib/pulp/bin/ansible-galaxy
  python version = 3.9.7 (default, Aug 30 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
  jinja version = 3.0.1
  libyaml = True
No config file found; using defaults
Starting galaxy role install process
- downloading role 'elasticsearch', owned by 
[WARNING]: - elasticsearch was NOT installed successfully: None (HTTP Code: 500, Message: Internal Server Error)
ERROR! - you can use --ignore-errors to skip failed roles and finish processing the list.
```

Traceback form the logs:
```
pulp [e4a12db90fe04f448772c444301dbb21]:  - admin [12/Apr/2022:09:43:11 +0000] "GET /pulp_ansible/galaxy/test/api/v1/roles/?owner__username=&name=elasticsearch HT>
[...]

PR #911, opened 2022-04-12
Fix 500 when distribution doesn't have a repository

gerrod3:
fixes: #909

### DIFF
--- a/pulp_ansible/app/galaxy/serializers.py
+++ b/pulp_ansible/app/galaxy/serializers.py
@@ -43,7 +43,7 @@ class GalaxyRoleVersionSerializer(serializers.Serializer):
         """
         Get source.
         """
-        distro_base = self.context["request"].parser_context["kwargs"]["path"]
+        distro_base = self.context["path"]
         distro_path = "".join([settings.CONTENT_ORIGIN, settings.CONTENT_PATH_PREFIX, distro_base])
 
         return "".join([distro_path, "/", obj.relative_path])
@@ -79,7 +79,7 @@ class GalaxyCollectionSerializer(serializers.Serializer):
         return (
             "{hostname}/pulp_ansible/galaxy/{path}/api/v2/collections/{namespace}/{name}/"
             "versions/".format(
-                path=obj.path,
+                path=self.context["path"],
                 hostname=settings.ANSIBLE_API_HOSTNAME,
                 namespace=obj.namespace,
                 name=obj.name,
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/galaxy/views.py
```diff
@@ -139,24 +165,16 @@ def get_queryset(self):
...
+        return Collection.objects.filter(versions__pk__in=self._distro_content).distinct()
 
     def post(self, request, path):
         """
         Queues a task that creates a new Collection from an uploaded artifact.
         """
         distro = get_object_or_404(AnsibleDistribution, base_path=path)
+        if not distro.repository and not distro.repository_version:
+            return HttpResponse(status=400, reason="Distribution has no repository.")
+
```
fao89, line 177: 👍🏽

fao89 (approved):

pulp_ansible/app/galaxy/serializers.py
```diff
@@ -79,7 +79,7 @@ def get_versions_url(self, obj) -> str:
         return (
             "{hostname}/pulp_ansible/galaxy/{path}/api/v2/collections/{namespace}/{name}/"
             "versions/".format(
-                path=obj.path,
+                path=self.context["path"],
```
mdellweg, line 82: Can you explain, why this should be better? To me it looks like the serializer would stop working if no context is given.

pulp_ansible/app/galaxy/views.py
```diff
@@ -20,6 +20,46 @@
 )
 
 
+class DistributionMixin:
```
mdellweg, line 23: This looks a lot like the `galaxy.v3.views.AnsibleDistributionMixin`. Only that the other one adds signatures. Right?

gerrod3: Correct, it's been copied pasted over.

pulp_ansible/app/galaxy/serializers.py
gerrod3: Before we were annotating the object with the path every time. So for list endpoints we had to iterate over each object and set the path. So the serializer technically needed a context anyway since path wasn't apart of the object, this way is just more efficient and consistent with the other v3 views.

mdellweg: Should we separate this improvement from the bugfix in case some things are requested for backports?

mdellweg (approved):
Approving, now.
Use your best judgement whether splitting the fix from the improvement is worthwhile.

PR #945, opened 2022-05-05
Added CollectionDownload log table

brumik:
fixes https://github.com/pulp/pulp_ansible/issues/946


pulp_ansible/app/models.py
```diff
@@ -211,6 +212,29 @@ class Meta:
...
+    # Data Fields
+    collection_version = models.ForeignKey(
+        CollectionVersion, on_delete=models.CASCADE, related_name="download_logs"
+    )
+    user = models.ForeignKey(
+        get_user_model(), on_delete=models.CASCADE, null=True, related_name="collection_download_logs"
+    )
+    ip = models.GenericIPAddressField()
+    org_id = models.CharField(max_length=64, null=True)
+    user_agent = models.CharField(max_length=64)
```
newswangerd, line 229: user agents can be pretty long, so lets increase the max length here

mdellweg: I'd prefer to collect the useragent verbatim here. Also why not make it a `TextField`?

bmbouter:
The unit test runner uses pytest-django so you can do database-stuff with unittests. Also you can use the pytest facilities like [`monkeypatch`](https://docs.pytest.org/en/6.2.x/monkeypatch.html).

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -550,6 +559,70 @@ class CollectionArtifactDownloadView(views.APIView):
...
+            else:
+                ip = request.META.get("REMOTE_ADDR")
+            return ip
+
+        def _get_user_agent(request):
+            user_agent = request.headers.get("user-agent")
+            if not user_agent:
+                return "unknown"
+
+            return user_agent
```
newswangerd, line 579:
```suggestion
        user_agent = request.headers.get("user-agent", "unknown")
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -550,6 +559,70 @@ class CollectionArtifactDownloadView(views.APIView):
...
+            return identity["internal"]["org_id"]
+
+        def _get_collection_from_filename(filename):
+            collection_namespace = filename.split("-")[-3]
+            collection_name = filename.split("-")[-2]
+            collection_version = filename.split("-")[-1][:5]
+
+            return CollectionVersion.objects.get(
+                namespace=collection_namespace, name=collection_name, version=collection_version
+            )
```
newswangerd, line 615:
```suggestion
        ns, name, version = filename.split("-", maxsplit=3)
        collection = get_object_or_404(CollectionVersion, namespace=ns, name=name, version=version)
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -550,6 +559,70 @@ class CollectionArtifactDownloadView(views.APIView):
...
+
+        CollectionDownloadLog.objects.create(
+            collection_version=_get_collection_from_filename(filename),
+            user=_get_user(request),
+            ip=_get_client_ip(request),
+            org_id=_get_org_id(request),
+            user_agent=_get_user_agent(request),
+            repository=AnsibleDistribution.objects.get(name=distro_base_path).repository,
```
newswangerd, line 623: use `get_object_or_404` instead of `objects.get` so this doesn't throw a 500 if the object doesn't exist.

CHANGES/946.feature
```diff
@@ -0,0 +1 @@
+Added CollectionDownloadLog table and logger
```
mdellweg, line 1:
```suggestion
Added CollectionDownloadLog table and logger.
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -550,6 +559,61 @@ class CollectionArtifactDownloadView(views.APIView):
...
+        ip = None
+        x_forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
+        ip = x_forwarded_for.split(",")[0] if x_forwarded_for else request.META.get("REMOTE_ADDR")
+
+        user_agent = request.headers.get("user-agent", "unknown")
+
+        # Getting collection version
+        ns, name, version = filename.split("-", maxsplit=3)
+        # Get off the ending .tar.gz
+        version = version[0:-7]
```
mdellweg, line 603: Maybe we should use `split(".")` instead.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -571,9 +635,23 @@ def get(self, request, *args, **kwargs):
...
         return redirect(url)
 
 
+class CollectionDownloadLogViewSet(viewsets.ReadOnlyModelViewSet):
+    """Collection download viewset."""
+
+    model = CollectionDownloadLog
+    serializer_class = CollectionDownloadLogSerializer
+    filterset_class = CollectionDownloadLogFilter
+    filter_backends = (DjangoFilterBackend,)
```
mdellweg, line 651: I think, this is set by default in the drf-settings already. This line should not be necessary.

mdellweg (commented):
This change collects a lot of sensitive information and makes it available to every logged in user. I think we need to find a way to restrict the access to the new endpoint.
Do we also want to hide it behind a feature flag? @bmbouter

brumik:
@mdellweg Originally we did not wanted to have an endpoint for retrieving the data. It was added later so maybe we can remove it. (but then I have no idea how to test this feature since download is not working in unit tests and functional tests cannot query db)

mdellweg:
> @mdellweg Originally we did not wanted to have an endpoint for retrieving the data. It was added later so maybe we can remove it. (but then I have no idea how to test this feature since download is not working in unit tests and functional tests cannot query db)

I see your point. But adding an (official) endpoint just for the sake of tests seems wrong, too. How do you intend to extract the information? Would for example the download count show on some query? With pulp-smash it should be possible to run some python code in the context of the app [0]. Maybe that would help.

[0] https://github.com/pulp/pulp-smash/blob/4b696fbce2b6ce4bb41da5145c653af3911d3deb/pulp_smash/utils.py#L153

brumik:
As far as I know, the application would use direct db queries to extract the data, but I am not entirely sure about the details.

goosemania:
this is for https://issues.redhat.com/browse/AAH-1558

mdellweg:
closed for #1118

PR #955, opened 2022-05-09
Fix S3 URL

fao89:
[noissue]
Ref: https://github.com/pulp/pulpcore/pull/2684


mdellweg:
As a bugfix, it should have a changelog.

mdellweg (approved):

PR #1033, opened 2022-06-15
Pin requirements

gerrod3:
[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-aiofiles>=0.8.0
+aiofiles~=0.8.0
 async_lru~=1.0
-galaxy_importer>=0.4.1
-GitPython>=3.1.24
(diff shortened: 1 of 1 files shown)


fao89 (commented):
it would still fail the CI for galaxy_importer and others,
you can run this locally: https://github.com/pulp/pulp_ansible/blob/main/.ci/scripts/upper_bound.py

fao89 (approved):

ipanova:
The way jsonschema dep is pinned here breaks the deployment when pulp_ansible and pulp_container are installed together like this 
`pkg_resources.VersionConflict: (jsonschema 4.6.0 (/usr/local/lib/python3.10/site-packages), Requirement.parse('jsonschema~=3.0'))`
We need to fix this and re-release pulp-ansible

PR #1034, opened 2022-06-15
Update files from CI

bmbouter:
This also rewrites the signature tests to use the pytest signing service
fixtures from pulpcore.

[noissue]


dkliban:
Please update the template_config.yaml with `aiohttp_fixtures_origin: 127.0.0.1` and reapply the plugin_template.

pulp_ansible/tests/functional/api/collection/v3/test_deletion.py
```diff
@@ -201,6 +202,7 @@ def test_delete_deprecated_content(self):
 
         assert len(latest_version.content_summary.present) == 0
 
+    @unittest.skip("needs to use signing fixtures from pulpcore")
```
bmbouter, line 205: @mdellweg and @gerrod3 I don't think we have time to convert this and then release. I'm proposing we accept this skip for this PR, release, and then I'll focus on converting this module to pytest as my next piece of work. I won't get distracted and forget to do it. Is this ok with you? There is one more skip test here too (somewhere else in the PR).

pulp_ansible/tests/functional/api/collection/test_signatures.py
```diff
@@ -1,230 +1,206 @@
...
+def test_collection_signatures_no_deletion(ansible_collection_signatures_client):
+    """Test the API doesn't provide a delete for a Collection Signature obj."""
+    with pytest.raises(AttributeError):
+        ansible_collection_signatures_client.delete
+
+
+# Tests for syncing Collection Signatures
+
+
+def distro_serving_one_signed_one_unsigned_collection(
```
gerrod3, line 128: Think this needs a fixture mark.

bmbouter: It's a fixture not a test, so I'm not sure I can even mark it?

gerrod3: Wdym? If it's a pytest fixture doesn't it need the `@pytest.fixture` mark?

mdellweg:
This has been adopted in another PR.

Issue #1035, opened 2022-06-15
Nightly Signature Tets Failing with gpg error


bmbouter:
[The nightlies](https://github.com/pulp/pulp_ansible/actions/workflows/nightly.yml) are failing (and maybe all PRs) with:

```
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_00_create_signed_collections ERROR                                                                                                                                                           [ 14%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_01_upload_signed_collections ERROR                                                                                                                                                           [ 15%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_02_read_signed_collection ERROR                                                                                                                                                              [ 16%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_03_read_signed_collections ERROR                                                                                                                                                             [ 17%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_04_partially_update ERROR                                                                                                                                                                    [ 18%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_05_fully_update ERROR                                                                                                                                                                        [ 18%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_06_delete ERROR                                                                                                                                                                              [ 19%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CRUDCollectionVersionSignatures::test_07_duplicate ERROR                                                                                                                                                                           [ 20%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CollectionSignatureSyncing::test_sync_signatures ERROR                                                                                                                                                                             [ 21%]
devel/pulp_ansible/pulp_ansible/tests/functional/api/collection/test_signatures.py::CollectionSignatureSyncing::test_sync_signatures_only ERROR
```

When I run one of those tests locally with:

[...]

PR #1036, opened 2022-06-15
Fix signature tests

gerrod3:
fixes: #1035

Also, includes a small fix for an error masking bug in import_collection.


bmbouter:
oh awesome!!!!!

PR #1040, opened 2022-06-17
Sort lists before assertions.

ipanova:
[noissue]

### DIFF
--- a/pulp_ansible/tests/functional/api/collection/v3/test_redirects.py
+++ b/pulp_ansible/tests/functional/api/collection/v3/test_redirects.py
@@ -79,7 +79,7 @@ class CollectionContentGuardDownloadTestCase(TestCaseUsingBindings, SyncHelpersM
         r1 = c1.list(self.distribution.base_path, self.distribution.base_path)
         r2 = c2.list(self.distribution.base_path)
 
-        assert r1 == r2
+        assert sorted(r1) == sorted(r2)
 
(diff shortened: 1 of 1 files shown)


pulp_ansible/tests/functional/api/collection/v3/test_redirects.py
```diff
@@ -79,7 +79,7 @@ def _verify_all_collections(self, new_api, old_api):
         r1 = c1.list(self.distribution.base_path, self.distribution.base_path)
         r2 = c2.list(self.distribution.base_path)
 
-        assert r1 == r2
+        assert sorted(r1) == sorted(r2)
```
mdellweg, line 82: @newswangerd It looks like the redirected api may report things in different orders (sometimes). Is it ok to adjust the test for this? Or should it be seen as a bug in the api?

ipanova: https://github.com/pulp/pulp_ansible/runs/6917875772?check_suite_focus=true#step:16:1230 here is the failure. This list sorting would not work because it has nested dict structure. @newswangerd  i am leaving this into your hands to decide what exactly you need/want to assert

newswangerd: the ordering should be identical. I don't know why that would change on a redirect.

PR #1069, opened 2022-06-30
Fix skipped tests

mdellweg:


mdellweg:
wait for https://github.com/pulp/pulpcore/pull/2913

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -1,110 +1,99 @@
...
-        monitor_task(self.repo_api.sign(repository_href, signing_body).task)
-        repo = self.repo_api.read(repository_href)
-        self.assertEqual(repo.latest_version_href[-2], "2")
-
-        keyring_arg = f"--keyring {get_client_keyring()}"
-        self.perform_install_test(ANSIBLE_DEMO_COLLECTION, distribution, keyring_arg)
+from pulp_ansible.tests.functional.utils import gen_ansible_remote
+
+
+@pytest.fixture()
```
bmbouter, line 17:
```suggestion
@pytest.fixture
```

pulp_ansible/tests/functional/cli/test_collection_install.py
```diff
@@ -1,110 +1,99 @@
...
+    )
+    repo = ansible_repo_factory(remote=remote.pulp_href)
+
+    # Sync
+    repository_sync_data = AnsibleRepositorySyncURL()
+    sync_response = ansible_repo_api_client.sync(repo.pulp_href, repository_sync_data)
+    monitor_task(sync_response.task)
+
+    # Publish
+    distribution = ansible_distribution_factory(repo)
```
bmbouter, line 37:
```suggestion
    return ansible_distribution_factory(repo)
```

.github/workflows/scripts/post_before_script.sh
```diff
@@ -2,13 +2,11 @@
...
 
-echo "machine pulp
-login admin
-password password
-" > ~/.netrc
+# Add the QE (public) key to the container env for `test_sign_locally_then_upload_signature`.
+# Please, remove once the verification was made independent of the system keyring.
 
-chmod og-rw ~/.netrc
+cmd_prefix bash -c "curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-KEY-pulp-qe | gpg --import"
```
bmbouter, line 8: What's the motivation for this part? Aka, what breaks if this isn't here?

mdellweg: The very test mentioned three lines above: `test_sign_locally_then_upload_signature`

bmbouter, line 8: Thanks. With the fixtures from pulpcore I don't think this is needed. I'm sure I'm not understand it. I expected all keyring config to not be global and to be provided from the pulpcore fixtures.

gerrod3: Basically the signature upload verification code is bugged right now, it doesn't work if a keyring isn't specified and the key is not in the global (default) keyring. This is why the test passed on my machine, but not on the CI. @mdellweg is creating a prototype replacement to fix this.

gerrod3 (approved):

bmbouter (approved):

PR #1071, opened 2022-06-30
Pin requirements

gerrod3:
[noissue]


bmbouter:
Is it possible we can declare ranges instead of ~= style? We've been having a lot of issues lately with pulp components not being compatible with other software installed in the env. Basically the upper bounds could just be the latest semver safe release for each dep (if they use semver). Does that make sense?

requirements.txt
```diff
@@ -1,9 +1,9 @@
...
-packaging
-pulpcore>=3.18
-PyYAML
-semantic_version
+aiofiles>=0.8.0,<0.9
+async_lru>=1.0,<1.1
+galaxy_importer>=0.4.5,<0.5
+GitPython>=3.1.24,<3.2
+jsonschema>=4.4,<4.7
+packaging>=21,<22
```
bmbouter, line 6: What motivated the lower bound here?

requirements.txt
```diff
@@ -1,9 +1,9 @@
...
-jsonschema~=4.4
-packaging
-pulpcore>=3.18
-PyYAML
-semantic_version
+aiofiles>=0.8.0,<0.9
+async_lru>=1.0,<1.1
+galaxy_importer>=0.4.5,<0.5
+GitPython>=3.1.24,<3.2
+jsonschema>=4.4,<4.7
```
bmbouter, line 5: I think in practice we do support much older ones. We kind of consolidated on 4.4 project-wide, but the lower bound could be lower. This was already set though, so I'm not asking for a change here, just a FYI.

requirements.txt
```diff
@@ -1,9 +1,9 @@
...
+packaging>=21,<22
+pulpcore>=3.20,<3.25
+PyYAML<6.0
+semantic_version>=2.9,<2.11
```
bmbouter, line 9: What motivated the lower bound here?

bmbouter:
After this is merged can you also open a similar one against `main` only without the pulpcore change?

requirements.txt
gerrod3: Going to have to be more specific with the package name. `packaging`'s changelog hasn't updated much from version 21.0 released last July.

requirements.txt
gerrod3: `semantic_version` added python 3.7+ support in 2.9, so it should be the lowest we allow for a python 3.8 project.

requirements.txt
fao89: https://github.com/pulp/pulp_container/issues/772

PR #1062, opened 2022-06-27
[WIP] legacy v1 api for roles

jctanner:
This has been based on 0.13 because main/master was incompatible with the pulp core head / 3.20 release at the time.

PR #1072, opened 2022-06-30
[WIP] api/v1/imports support for legacy roles

jctanner:
Deprecates https://github.com/pulp/pulp_ansible/pull/1062 and patches main instead of 0.13 now that main's CI is functional again.


jctanner:
CI appears to be broken again.

jctanner:
Without a route for /api that returns the list of versions, the cli will fail import like so ...

```
vagrant@galaxy-dev:~$ ansible-galaxy role import -vvvv geerlingguy ansible-role-docker -c -s https://pulp:443/
ansible-galaxy [core 2.12.5]
  config file = None
  configured module search path = ['/home/vagrant/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/dist-packages/ansible
  ansible collection location = /home/vagrant/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible-galaxy
  python version = 3.8.10 (default, Mar 15 2022, 12:22:08) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True
No config file found; using defaults
Initial connection to galaxy_server: https://pulp:443/
Opened /home/vagrant/.ansible/galaxy_token
Calling Galaxy at https://pulp:443/
Calling Galaxy at https://pulp:443/api
Initial connection to galaxy_server: https://pulp:443/
Calling Galaxy at https://pulp:443/
Calling Galaxy at https://pulp:443/api
ERROR! Error when finding available api versions from cmd_arg (https://pulp:443/) (HTTP Code: 404, Message: Not Found)
```

mdellweg:
> [stream] CI appears to be ~broken again~ unstable.

FWIW, it's not marked as "required".

pulp_ansible/app/galaxy/v1/utils.py
```diff
@@ -0,0 +1,280 @@
...
+import subprocess
+import tempfile
+import yaml
+
+
+def get_path_git_root(path):
+    """
+    Find the root directory of a git checkout.
+    """
+    cmd = "git rev-parse --show-toplevel"
```
mdellweg, line 13: Can we use `git.Repo` for these?

jctanner: Yep, I've actually switched over to that in the task, so this file can probably be completely deleted once I get a better sense of where we are going.

pulp_ansible/app/galaxy/v1/tasks.py
```diff
@@ -0,0 +1,184 @@
...
+# from pulp_ansible.app.tasks.roles import GITHUB_URL
+
+# from galaxy_importer.loaders.content import RoleLoader
+# from galaxy_importer.utils import markup as markup_utils
+
+# from git import GitCommandError
+from git import Repo
+
+
+def legacy_role_import(*args, **kwargs):
```
bmbouter, line 32: In the final version, I prefer having these be named args, kwargs.

jctanner: yeah, i hope to give it fully written out args/kwargs with typedef and docstrings ... just waiting for us to settle on the ins and outs of these functions.

pulp_ansible/app/galaxy/v1/tasks.py
```diff
@@ -0,0 +1,184 @@
...
+            "role_name": role_name,
+            "name": role_name,
+            "namespace": github_user,
+            "role_version": github_reference,
+        }
+
+        # pulp_ansible's Roles are actuallly Role Versions
+        # The roles list viewset aggregates the versions together
+        # by using a {namespace}.{name} as the primary key.
+        role = Role(
```
bmbouter, line 140: I believe it would be good to put the `Role` and `ContentArtifact` into one transaction. `Artifacts` saving can take a long time if Pulp is configured with an object storage that has high latency, so I think you want to make that one outside of the transaction and `save()` it. Pulp's orphan cleanup handles deleting unattached Artifacts so it's safe to make it ahead of time even if the execution stops just after it. Additionally, if it did fail, and the import was attempted again, and it made the same Artifact you could catch the duplicate error from the DB and reuse the existing artifact.

bmbouter: The last part about catching the duplicate is a rare case and can be ignored for now I think.

pulp_ansible/app/galaxy/v1/tasks.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # Create a real artifact
+        artifact = Artifact.init_and_validate(tarfn)
+        artifact.save()
+
+        # This is a pulp'ism for the source url
+        # but we will generate the download_url in the
+        # serializer so this wouldn't ever be
+        # used ... ?
+        relative_path = "%s/%s/%s.tar.gz" % (
```
bmbouter, line 168: I believe the content app needs this at serving time, otherwise it can't know. Internally the files aren't stored with filenames like `something.tar.gz` it's stored in a sha256 content addressable storage. I suspect (maybe) if this wasn't right it would not work?

pulp_ansible/app/galaxy/v1/views.py
```diff
@@ -0,0 +1,136 @@
...
+    serializer_class = GalaxyRoleSerializer
+    authentication_classes = []
+    permission_classes = []
+
+    def get_queryset(self):
+        """
+        Get the list of items for this view.
+        """
+        # distro = get_object_or_404(AnsibleDistribution, base_path=self.kwargs["path"])
+        distro = get_object_or_404(AnsibleDistribution, base_path="legacy")
```
bmbouter, line 38: This is likely just something to come later, but I imagined `base_path` would come from the LEGACY_REPOSITORY_NAME somehow.

pulp_ansible/app/galaxy/v1/views.py
```diff
@@ -0,0 +1,136 @@
...
+            "github_repo": request.data.get("github_repo"),
+            "github_reference": request.data.get("github_reference", ""),
+            "repository_id": request.data.get("repository_id"),
+            "alternate_role_name": request.data.get("alternate_role_name"),
+        }
+        role_name = kwargs["alternate_role_name"] or kwargs["github_repo"].replace(
+            "ansible-role-", ""
+        )
+        task = dispatch(legacy_role_import, exclusive_resources=[legacy_repo], kwargs=kwargs)
+        hashed = abs(hash(str(task.pulp_id)))
```
bmbouter, line 123: Using `hash` I think is a good way to work around the integer versus UUID issue.

pulp_ansible/app/galaxy/v1/views.py
jctanner: Yeah, I missed this one.

pulp_ansible/app/galaxy/v1/tasks.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # Use Content to tie the role and artifact together
+        ca1 = ContentArtifact.objects.create(
+            artifact=artifact, content=role, relative_path=relative_path
+        )
+        ca1.save()
+
+        # Add the role to the legacy repository via a new version.
+        legacy = AnsibleRepository.objects.get(pulp_id=ansible_repo_id)
+        add_and_remove(int(legacy.pk), [role.pk], [])
```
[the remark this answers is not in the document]
bmbouter: I'm surprised this works with the `int` if it does.

pulp_ansible/app/galaxy/v1/constants.py
```diff
@@ -0,0 +1,2 @@
+LEGACY_DISTRIBUTION_PATH = "legacy"
+LEGACY_REPOSITORY_NAME = "legacy"
```
bmbouter, line 2: I imagine these will be configurable as settings and these would be the defaults. Is that your plan also?

To do that define them [here](https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/settings.py) and then document them [here](https://github.com/pulp/pulp_ansible/blob/main/docs/settings.rst).

pulp_ansible/tests/functional/cli/test_role_import.py
```diff
@@ -0,0 +1,97 @@
...
+from pulp_ansible.tests.functional.utils import TestCaseUsingBindings
+from pulp_ansible.tests.functional.constants import (
+    ANSIBLE_DOCKER_ROLE_NAMESPACE,
+    ANSIBLE_DOCKER_ROLE_NAME,
+    ANSIBLE_DOCKER_ROLE_REPO,
+)
+from pulp_ansible.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+
+
+class ImportRoleTestCase(TestCaseUsingBindings):
```
bmbouter, line 16: I'd be ok to accept this style for now as the other tests also have it. It would be much nicer as pytest, but if that's time-prohibitive in this PR, I understand.

pulp_ansible/app/galaxy/v1/constants.py
jctanner: Makes sense. I'll give it a shot.

pulp_ansible/tests/functional/cli/test_role_import.py
jctanner: Can you point me at an example functional test here or elsewhere that meets the criteria? I thought TestCaseUsingBindings was what you were steering me towards yesterday, so I must have misunderstood the expectation.

bmbouter: Sure! This [PR shows](https://github.com/pulp/pulp_ansible/pull/1069/files) a rewrite similar to what I'm talking about. It has these kinds of qualities:

* It's pytest native, so it no longer inherits from unittest at all (`TestCaseUsingBindings` inherits from `unittest`).
* It uses pytest fixtures for it's arrangement and cleanup. You can see the pytest fixtures for  [pulp_ansible here](https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/conftest.py), [pulpcore here](https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py) and [pulp-smash](https://github.com/pulp/pulp-smash/blob/main/pulp_smash/pulp3/pytest_plugin/__init__.py) provides some too. `pytest --fixtures` will show you the full list.
* It's bindings based. Your code here is too, but receive the bindings objects from pytest fixtures.
* Use native `assert` instead of self.AssertEqual, etc
* Generate any data you need at runtime, so you can run a test over and over without every having to call orphan cleanup. For this we use [`build_collection` from oriontools](https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/conftest.py#L131).
* It's ok to leave behind Artifacts (since they are unique per test run) and any kind of Content (also unique per test run).
* Have no tests depend on other tests.
* Mark any test that can be run in parallel with `@pytest.mark.parallel` and any test that should run on in the nightly as `@pytest.mark.nightly`.

If you grep for `pytest` through the test folder you'll find many already, and it likely has examples of a lot of the parts you want to do here.

pulp_ansible/app/galaxy/v1/tasks.py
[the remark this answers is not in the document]
mdellweg: Maybe because uuid is also just a disguised number.

```python
In [3]: int(uuid.uuid4())
Out[3]: 120903135329304400464259694501259332
```

But i agree it should not be casted here.

pulp_ansible/app/galaxy/v1/tasks.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Use Content to tie the role and artifact together
+    ca1 = ContentArtifact.objects.create(
+        artifact=artifact, content=role, relative_path=relative_path
+    )
+    ca1.save()
+
+    # Add the role to the legacy repository via a new version.
+    legacy = AnsibleRepository.objects.get(pulp_id=ansible_repo_id)
+    add_and_remove(legacy.pk, [role.pk], [])
```
bmbouter, line 216: L#215 and L#216 should not be in this transaction I believe.

pulp_ansible/app/galaxy/v1/tasks.py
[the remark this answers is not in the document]
jctanner: fixed

pulp_ansible/app/galaxy/v1/tasks.py
jctanner: fixed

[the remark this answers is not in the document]
jctanner: fixed

pulp_ansible/app/galaxy/v1/tasks.py
```diff
@@ -0,0 +1,229 @@
...
+        artifact.save()
+
+        # Create the role and bind with the artifact as a "content artifact"
+        role, content_artifact = _create_role_and_content_artifact(
+            metadata, relative_path, artifact
+        )
+
+        # Add the role to the legacy repository via a new version.
+        legacy = AnsibleRepository.objects.get(pulp_id=ansible_repo_id)
+        add_and_remove(legacy.pulp_id, [role.pulp_id], [])
```
gerrod3, line 193: My previous comment was meant to show that this query is unnecessary. You can just pass the `ansible_repo_id` to the `add_and_remove` call.

pulp_ansible/tests/functional/cli/test_role_import.py
jctanner: I think this is fixed now, although maybe a bit messy.

PR #1076, opened 2022-07-01
Add requirement pins

gerrod3:
[noissue]
@bmbouter dependabot will update the requirements appropriately or do I need to change something?


requirements.txt
```diff
@@ -1,9 +1,9 @@
...
-pulpcore>=3.18
-PyYAML
-semantic_version
+aiofiles>=0.8.0,<0.9
+async_lru>=1.0,<1.1
+galaxy_importer>=0.4.5,<0.5
+GitPython>=3.1.24,<3.2
+jsonschema>=4.4,<4.7
+packaging>=21.3,<22
+pulpcore>=3.20
```
mdellweg, line 7: I think, with the new deprecation policy, we should specify `pulpcore>=3.20,<3.25` here. The reason we kept this without the upper bound was that we had to change it with every pulpcore release. And that very reason is addressed by the new policy.

bmbouter: I agree. @mdellweg how can we identify this more broadly for all plugin writers?

bmbouter:
I reran the CI job twice and it failed on that 1 test, 1 job both times. I don't know why.

requirements.txt
mdellweg: We could move the upper bounds check from the release pipeline to the main ci.

mdellweg (approved):

requirements.txt
bmbouter: We should do this @dkliban @fao89 FYI. Basically we want to address release-time issues ahead of release time FTW.

fao89: it is a matter of calling https://github.com/pulp/pulp_ansible/blob/main/.ci/scripts/upper_bound.py at lint time

bmbouter: @fao89 how would you feel about making that PR?

fao89: I can do it, but people need to be aware as it would break all the CIs

fao89: https://github.com/pulp/plugin_template/pull/666

fao89:
is anything blocking the merge?

PR #1088, opened 2022-07-05
Add `gpgkey` to ansible repository

### LINKED ISSUE #1086: Signature validation depends on an externally provided keyring
This is bad for several reasons:
- The keyring must be provisioned and made available on all pulp nodes/containers/service units (installer nightmare).
- The repository serializer references a local file on the server to select the keyring (bad api).
- It is impossible to provide a key to verify against via the api.
- It is impossible to use an ephemeral keyring in the tests.

The proposed solution is to add a (text-)field on the repository that can hold a (maybe multiple) private key. Verification of an uploaded signature will then only depend on information in the applications database.

mdellweg:
fixes #1086

Required PR: https://github.com/pulp/pulpcore/pull/2931


docs/workflows/signature.rst
```diff
@@ -7,14 +7,14 @@ how to add signature support.
...
+In order to verify signature validity on uploads you will need to store your trusted key on the
+repositories ``gpgkey`` attribute.
 
 .. note::
-    You can upload signatures without supplying Pulp your keyring, but ``pulp_ansible`` will not
-    perform validity checks on the uploaded signature.
+   You can upload signatures without supplying Pulp any key, but ``pulp_ansible`` will not
+   perform validity checks on the uploaded signature. You will also have to configure the
+   ``ANSIBLE_SIGNATURE_REQUIRE_VERIFICATION`` setting to ``False``. By default and once a key is
+   provided, all signatures impossible to verify are rejected.
```
rochacbruno, line 17: dies it mean this setting is `True` by default?

mdellweg: It currently is, but that's up for debate.

rochacbruno: I am ok having is as True

If the default is False, then we need to add it to galaxy, but being True require no change on our side.

mdellweg: (If we can agree on `False` i would not need to adjust the cli tests right away.)

mdellweg: I do not feel comfortable to decide alone. Selecting `True` will change the behaviour for installations using default values. Setting to `False` will accommodate our current CI test suite. But also `True` is heading towards "safe by default".

@gerrod @bmbouter @newswangerd

mdellweg: We decided to keep the default true, but i'll try setting it to False in the CI for now to see if the CLI tests start passing again.

mdellweg:
Still depending on a core pr...

docs/workflows/signature.rst
newswangerd: Whatever you decide is fine with me. We can always change the default on our end if we want.

gerrod3 (approved):
Looks good on the pulp_ansible side.

docs/settings.rst
```diff
@@ -26,6 +26,15 @@ ANSIBLE_CONTENT_HOSTNAME
    CONTENT_ORIGIN, this would default to "https://example.com/pulp/content".
 
 
+ANSIBLE_SIGNATURE_REQUIRE_VERIFICATION
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+   By default, `pulp_ansible` rejects uploaded signatures if they cannot be verified against a
```
bmbouter, line 32:
```suggestion
   By default, ``pulp_ansible`` rejects uploaded signatures if they cannot be verified against a
```

requirements.txt
```diff
@@ -4,6 +4,6 @@ galaxy_importer>=0.4.5,<0.5
 GitPython>=3.1.24,<3.2
 jsonschema>=4.4,<4.7
 packaging>=21.3,<22
-pulpcore>=3.20,<3.25
+pulpcore>=3.21.dev,<3.25
```
bmbouter, line 7: I can see how this is required, but I guess we'll just have to remember to change this at release time?

bmbouter: I meant more to remember to change s/3.21.dev/3.21/ at release time, but I do think this is better than not changing it now.

bmbouter (approved):
Thanks!

Issue #1105, opened 2022-07-06
Unlock repository during collection import


newswangerd:
**Is your feature request related to a problem? Please describe.**
Right now collection imports lock the entire repository that the collection is being imported into (https://github.com/pulp/pulp_ansible/blob/1528769b01db6c3c0e4c6501e522fc7e706ae27e/pulp_ansible/app/galaxy/v3/views.py#L514). This makes it so that only one collection can be imported at a time. This is problematic because the import task can take a long time, and doesn't need to maintain a lock on the repository for the full duration of the import.

**Describe the solution you'd like**
Remove the repository lock during collection import and have the collection import launch a subtask (with the repository lock) to move the collection into the selected repository after the import is successful.

**Additional context**
This is needed for repository management in galaxy_ng to get rid of inbound repositories.

PR #1115, opened 2022-07-08
Remove repository lock from V3 Collection Import endpoint

gerrod3:
fixes: #1105

Right  now this is the bare minimum changes required to remove the repository lock from the initial import task. However, I think we should look into changing the `CollectionImport` model to accommodate for the new task group which would require a migration. @newswangerd 

TODO:

- [ ] Add/modify current tests
- [ ] Decide on how much info the `CollectionImport` model should report

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -34,6 +34,7 @@ from pulpcore.plugin.models import (
     Remote,
     RepositoryContent,
     Task,
+    TaskGroup,
 )
 from pulpcore.plugin.stages import (
     ArtifactDownloader,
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -338,11 +343,22 @@ def import_collection(
     CreatedResource.objects.create(content_object=collection_version)
 
     if repository_pk:
-        repository = AnsibleRepository.objects.get(pk=repository_pk)
-        content_q = CollectionVersion.objects.filter(pk=collection_version.pk)
-        with repository.new_version() as new_version:
-            new_version.add_content(content_q)
-        CreatedResource.objects.create(content_object=repository)
+        if not lockless:
```
mdellweg, line 346: You could try to see if the repository was in the locks of this task and decide that way.
WDYT?

PR #1118, opened 2022-07-13
Download log

mdellweg:
Attempting to supersede https://github.com/pulp/pulp_ansible/pull/945

this is for https://issues.redhat.com/browse/AAH-1558

AAH-1558


pulp_ansible/app/models.py
```diff
@@ -211,6 +222,28 @@ class Meta:
...
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        null=True,
+        related_name="download_logs",
+    )
+    ip = models.GenericIPAddressField()
+    extra_data = models.JSONField(null=True)
+    user_agent = models.TextField()
+    repository = models.ForeignKey(
```
newswangerd, line 242: Would repository version be more useful?

newswangerd (commented):
This looks good to me.

Now that I have a better understanding of repository versions, I think storing the repo version in the download log would be more useful than the repository.

pulp_ansible/app/models.py
mdellweg: Repository versions might be a bit more volatile if you are using `retain_repo_versions`. But if you like, you can get both.

newswangerd:
@mdellweg I'm running into circular dependecy issues. I think Levente got around it with this: https://github.com/pulp/pulp_ansible/pull/945/files#diff-ede042cbf5a92f6001b82d26b276d87a1f7fe3fa28e8a27913b219a83880e2a2R223

newswangerd (approved):
Everything seems to be working. Just need to apply the changes I mentioned above to resolve the circular import.

Issue #1106, opened 2022-07-06
As a user, I want to repair collection metadata


newswangerd:
**Is your feature request related to a problem? Please describe.**
We occasionally get collection versions in pulp ansible with missing or broken metadata. This can happen when syncing from a remote that doesn't contain the metadata on the API (such as galaxy.ansible.com). This can also happen if there are bugs in the galaxy importer or if the collection was imported using an older version of the galaxy importer. As a user, I would like a way to regenerate the metadata (this includes documentation, content counts, etc) for these collections from the collection tarball if it exists in pulp.

**Describe the solution you'd like**
Provide a repair metadata API call that allows a user to rebuild the collection metadata from the artifact for a specific collection version. This call should re run the galaxy importer on the collection tarball and load the resulting metadata into the the collection versions table.

Additionally it would be nice (but not necessary) to be able to call this repair task on:
- All collection versions in a repository
- All collection versions for a given collection. 

**Describe alternatives you've considered**
A django management command could also work, but we wouldn't be able to provide this as a feature on the galaxy ng UI if that's the case, so an API call would be vastly preferable.

cc @mdellweg @gerrod3

newswangerd:
@jctanner is going to work on this one.

PR #1119, opened 2022-07-16
Add sha256 uniqueness constraint to CollectionVersion

gerrod3:
Have to do lots of testing for this one, the migration is not pretty.


pulp_ansible/app/migrations/0042_collectionversion_sha256.py
```diff
@@ -0,0 +1,103 @@
...
+import subprocess
+
+
+def add_sha256_to_current_models(apps, schema_editor):
+    """Adds the sha256 to current CollectionVersion models."""
+    CollectionVersion = apps.get_model('ansible', 'CollectionVersion')
+    ContentArtifact = apps.get_model('core', 'ContentArtifact')
+    RemoteArtifact = apps.get_model('core', 'RemoteArtifact')
+    Artifact = apps.get_model('core', 'Artifact')
+    collection_bulk = {}
```
mdellweg, line 17: Getting it to work is the first prio of course. But in the end, we want this migration to work in batches.
On a really large installation, one by one is way to slow, but all in one will blow up the ram.

gerrod3: After asking if anyone uses GitRemotes (which can create on-demand Collections), it seems it is not widespread, so I simplified the migration to only migrate collections that have artifacts in the system. If on-demand collections are present an exception is raised with the collections that failed.

pulp_ansible/app/migrations/0044_collectionversion_sha256_migrate.py
```diff
@@ -0,0 +1,62 @@
...
+    if len(collection_bulk) > 0:
+        find_and_update_sha256()
+        if len(collections_to_update) > 0:
+            with transaction.atomic():
+                CollectionVersion.objects.bulk_update(collections_to_update, ["sha256",])
+
+    # If there are on-demand collections then the next migration will fail, so error here with
+    # helpful message on how to fix. No work will be performed by this migration on a second-run.
+    if len(collections_on_demand) > 0:
+        raise Exception(
```
ipanova, line 46: I don't think it is a good practice to raise exception in the middle on the migration because a) it already performed some changes to the db b) rollback might not be easiest in the middle of failed upgrade.

Is it possible to fetch sha256 from the RemoteArtifact of the on_demand collection content?

gerrod3: > I don't think it is a good practice to raise exception in the middle on the migration because a) it already performed some changes to the db b) rollback might not be easiest in the middle of failed upgrade.

Well I tried to design the migration to be able to be re-ran without re-doing the work it has already done and reverting this migration should be as easy as simply removing the new sha256 column. 

> Is it possible to fetch sha256 from the RemoteArtifact of the on_demand collection content?

Yes if sha256 was set on the RemoteArtifact, but it isn't required and the GIt sync code does not set the sha256 when doing an on_demand sync.

pulp_ansible/app/models.py
```diff
@@ -169,7 +172,7 @@ def relative_path(self):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
-        unique_together = ("namespace", "name", "version")
+        unique_together = ("sha256",)
```
mdellweg, line 175: So the artifact contains the information about namespace, name and version.
Do you see any conceivable way (maybe through on-demand sync) that pulp could be fooled about that information? Is this a problem we intend to solve, or do we in general trust upstream sources?

pulp_ansible/app/serializers.py
```diff
@@ -381,9 +381,9 @@ def validate(self, data):
             data = self.deferred_validate(data)
 
         sha256 = data["file"].hashers["sha256"].hexdigest()
-        artifact = Artifact.objects.filter(sha256=sha256).first()
-        if artifact:
-            ValidationError(_("Artifact already exists"))
+        cv = CollectionVersion.objects.filter(sha256=sha256).first()
+        if cv:
+            raise ValidationError(_("Collection Version already exists"))
```
mdellweg, line 386: Another story for another day, but i think we need this serializer to parse the artifact in deferred_validate.

pulp_ansible/app/migrations/0044_collectionversion_sha256_migrate.py
ipanova: I see.
My suggestion would be to first ( before doing any modifications to the data on the collections) identify whether there are any collections for which it is not possible to retrieve sha256 ( whether from artifact or remote_artifact), then find in which repos they are, and as last, raise error saying what repos need to be synced with immediate policy. Displaying a list of collection.pk is not very friendly and informative.

mdellweg: If we don't want anything to change in case we are unable to perform the whole migration, i would simply rely on the atomic transaction a migration is usually run in. Adding the extra check upfront introduces unnecessary complexity.

pulp_ansible/app/models.py
gerrod3: Which information being fooled? The name, namespace, version or the sha256? name, namespace, version is up to the user so it could be what ever they want. If you are talking about it differing from the downloaded artifact, then yes it is possible if an upstream messes with the database to change the metadata served or change which artifact is pointed to. I think in general we trust the upstream source (and since upstream will either be galaxy or pulp I think this is a safe assumption to make. Maybe this something we can consider in #1106 .

pulp_ansible/app/serializers.py
gerrod3: I think you are right, we should write plugin documentation for uploading content to Pulp for when and where fields should be processed and validated because I can never remember off the top of my head.

newswangerd:
I don't think this is a good idea at the moment. We don't really need this for repo management. This would be useful if we ever get asked to do multi tenancy, but so far we haven't had to do that. It might actually make repo management harder to do. Imagine looking up `foo.bar:1.0.0` and getting back 5 different collections with the same namespace, name and version. How would I begin to determine which one to use? It wouldn't be hard to accidentally end up with a bunch of identical collections lying around since `ansible-galaxy` doesn't reliably produce collections with the same SHA.

If we're going to do this, I would like to see a really solid reason for it. Otherwise, we're just creating a big headache for ourselves.

pulp_ansible/app/models.py
sivel: Is this the `sha256` of the tar.gz itself, or the `sha256` of the `MANIFEST.json`?

Since `ansible-galaxy collection build` always generates a new `MANIFEST.json` and `FILES.json`, we set the `mtime` to `time.time()` for those files in the tarball.  So if you simply run `ansible-galaxy collection build` 2 times, the `tar.gz` will have different `sha256`

Would this cause problems for you if someone tried to upload a rebuilt artifact?

As it is, the `sha256` of the tarball isn't necessarily equivalent of `(namepace, name, version)`.

gerrod3: This new `sha256` field is for the tarball `sha256` (the underlying artifact of the CollectionVersion).

Uploading a rebuilt artifact wouldn't break Pulp, it would just cause confusion for the user. With this change Pulp would gladly accept both the original and rebuilt collection since they would have different digests. Both could be stored in Pulp, but the repository you have them in would only hold the latest uploaded one. Doing a global content list for `(namespace, name, version)` could now return multiple results so the user would have to know the date uploaded or `sha256` to select the right one. 

However, having the new field be the `sha256` of the `MANIFEST.json` might be a better idea. In that scenario would the `sha256` of the `MANIFEST.json` of the second built artifact be the same as the first?

sivel: The `MANIFEST.json` will have the same sha, unless of course if something else was changed:

```console
$ shasum -a 256 sivel-toiletwater-0.0.10.tar.gz*
878311263fcd21f040022a33187d682a63a798e16a79d9bdf91d65e8af30b1a6  sivel-toiletwater-0.0.10.tar.gz.1
5fb2e8026e5dae65e1c40c19fa168883c524f3f99c90ec67a58ebe6ea920b51f  sivel-toiletwater-0.0.10.tar.gz.2
$ tar -Oxzvf sivel-toiletwater-0.0.10.tar.gz.1 MANIFEST.json | shasum -a 256
x MANIFEST.json
9a0812ebbd0107139c0e3e4684d585393cfce6a8d9fff4876e6cabd3e244ade7  -
$ tar -Oxzvf sivel-toiletwater-0.0.10.tar.gz.2 MANIFEST.json | shasum -a 256
x MANIFEST.json
9a0812ebbd0107139c0e3e4684d585393cfce6a8d9fff4876e6cabd3e244ade7  -
```

gerrod3: Thank you for testing this. Am I safe to assume that the `MANIFEST.json` is source of truth for two collections being the same? If so, I'll change the new field to be the `MANIFEST.json`'s `sha256` to remove the issue of uploading duplicate build artifacts.

sivel: I guess that is up for interpretation, but to clarify, the `MANIFEST.json` holds a `sha256` for `FILES.json`, which holds a `sha256` for every file in the artifact.  So the `sha256` of a `MANIFEST.json` which also holds `namespace`, `name`, and `version` should accurately represent whether 2 collections are different.

mdellweg: My question is whether the FILES.json is always ordered in the same way, thereby giving the same checksum if the same files are present.

sivel: The order of `FILES.json` is determined by filesystem order, effectively that of `os.listdir`.

PR #1130, opened 2022-07-21
[fix] Use Dynaconf lazy variables insted of re-importing the settings.

rochacbruno:
Instead of doing `from dynaconf import settings` which causes the whole
settings load to be executed again, it is better to use the support for
lazy evaluation and formatting https://www.dynaconf.com/dynamic/#format-token
so it is possible to define a variable that is a template to be formed from
interpolation with other in settings variables.

So inside a settings file that is managed by dynaconf like:

```python
from dynaconf import settings
FOO = settings.BAR + "/zaz"
```

It is better to do

```python
FOO = "@format {this.BAR}/zaz"
```

or if Jinja powers is needed

```python
FOO = "@jinja {{ this.BAR }}/zaz"
```

This way there is no need to reload all the settings and avoids circular dependencies.

[noissue]


rochacbruno:
This PR is still missing dynaconf fix first

rochacbruno:
Dynaconf fix still pending, will be fixed for v4.0 soon

rochacbruno:
this is probably fixed by https://github.com/dynaconf/dynaconf/pull/944 I will test it

rochacbruno:
This PR now depends on Dynaconf 3.3.0 that will include this https://github.com/dynaconf/dynaconf/pull/1040

mdellweg (approved):

mdellweg:
Do we need a versioned dependency on dynaconf here?

rochacbruno:
@mdellweg the feature used here is added on dynaconf 3.2.5 https://github.com/dynaconf/dynaconf/releases/tag/3.2.5 which is not available on pulpcore 3.49 https://github.com/pulp/pulpcore/blob/3.49/requirements.txt#L19 (the lower bound for pulp_ansible)

We would need to change main pulp_ansible to require at least pulpcore 3.50.

mdellweg:
> @mdellweg the feature used here is added on dynaconf 3.2.5 https://github.com/dynaconf/dynaconf/releases/tag/3.2.5 which is not available on pulpcore 3.49 https://github.com/pulp/pulpcore/blob/3.49/requirements.txt#L19 (the lower bound for pulp_ansible)
> 
> We would need to change main pulp_ansible to require at least pulpcore 3.50.

I thought, we could require that dynaconf version here until we bump pulpcore to >= 3.50.
But we can also wait. What would be more convenient?

rochacbruno:
I think we cannot require the version here because pulpcore 3.49 has upper bounds for dynaconf.

Maybe we can just wait the proper time to merget this, this is not blocking anything, it is just adding a bit of overhead on application startup time.

rochacbruno:
I think I am ready to fix it now.

rochacbruno:
After 2.5 years of waiting, this PR is finally mergeable! 🎉

mdellweg (approved):

PR #1174, opened 2022-08-16
Add endpoint to rebuild collection docs.

jctanner:
https://github.com/pulp/pulp_ansible/issues/1106
https://issues.redhat.com/browse/AAH-1820

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -357,6 +357,53 @@ def create_collection_from_importer(importer_result, metadata_only=False):
     return collection_version
 
 
+def rebuild_repository_collection_versions_metadata(
+    repository_pk, namespace=None, name=None, version=None
+):
+    """Rebuild metadata for all collection versions in a repo."""
+    repo = AnsibleRepository.objects.get(pulp_id=repository_pk)
+    repov = repo.latest_version()
+
+    qs = None
+    if namespace or name or version:
+        qkwargs = {}
+        if namespace:
+            qkwargs["namespace"] = namespace
+        if name:
+            qkwargs["name"] = name
+        if version:
+            qkwargs["version"] = version
+        qs = CollectionVersion.objects.filter(**qkwargs)
+
+    for cv in repov.get_content(content_qs=qs):
+        try:
+            rebuild_collection_version_meta(str(cv.pulp_id))
+        except Exception as e:
+            log.exception(e)
+
+
+def rebuild_collection_version_meta(collection_version_pk):
+    """Rebuild metadata for a single collection version."""
+    # find the collection version
+    collection_version = CollectionVersion.objects.get(pulp_id=collection_version_pk)
+
+    # where is the artifact?
+    ca = ContentArtifact.objects.filter(content_id=collection_version.content_ptr_id).first()
+    artifact = ca.artifact
+
+    # call the importer to re-generate meta
+    importer_result = process_collection(
+        artifact.file, filename=artifact.file.name, file_url=artifact.file.url, logger=None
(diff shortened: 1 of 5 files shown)


pulp_ansible/app/viewsets.py
```diff
@@ -327,6 +329,30 @@ def sync(self, request, pk):
         )
         return OperationPostponedResponse(result, request)
 
+    @extend_schema(
+        description="Trigger an asynchronous task to rebuild Ansible content meta.",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=True, methods=["post"], serializer_class=AnsibleRepositoryRebuildSerializer)
```
mdellweg, line 336: Defining an empty `inline_serializer` here should do the trick for drf-spectacular.

https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html?highlight=inline_serializer#drf_spectacular.utils.inline_serializer

pulp_ansible/app/tasks/collections.py
```diff
@@ -396,6 +396,37 @@ def create_collection_from_importer(importer_result, metadata_only=False):
...
+
+
+def rebuild_collection_version_meta(collection_version_pk):
+    """Rebuild metadata for a single collection version."""
+    # find the collection version
+    collection_version = CollectionVersion.objects.get(pulp_id=collection_version_pk)
+
+    # where is the artifact?
+    ca = ContentArtifact.objects.filter(content_id=collection_version.content_ptr_id).first()
+    artifact = ca.artifact
```
gerrod3, line 414: This is one way, the other way is:
```python
artifact = collection_version._artifacts.first()
```

pulp_ansible/app/viewsets.py
```diff
@@ -327,6 +329,30 @@ def sync(self, request, pk):
...
+        """
+        repository = self.get_object()
+        # serializer = AnsibleRepositoryRebuildSerializer(
+        #     data=request.data, context={"request": request, "repository_pk": repository.pk}
+        # )
+        # serializer.is_valid(raise_exception=True)
+
+        result = dispatch(
+            rebuild_repository_collection_versions_metadata,
+            exclusive_resources=[repository],
```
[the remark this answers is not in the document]
gerrod3: Probably safe to put repository lock in `shared_resources`. Question is should this also have an exclusive lock on all of `/pulp/api/v3/content/ansible/collection_versions/` since it's modifying the collections. It might be useless since I don't think there are any other tasks that require that lock even if they create or read collections.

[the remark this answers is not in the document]
mdellweg: I see this as repairing and not changing, and also yes we don't lock this in other places.

newswangerd:
Would it be possible to add namespace, name and version to the request body so that a user can rebuild the metadata for a specific set of collection versions?

pulp_ansible/app/tasks/collections.py
```diff
@@ -357,6 +357,55 @@ def create_collection_from_importer(importer_result, metadata_only=False):
...
+
+    # call the importer to re-generate meta
+    importer_result = process_collection(
+        artifact.file, filename=artifact.file.name, file_url=artifact.file.url, logger=None
+    )
+    importer_result["artifact"] = artifact
+    importer_result["artifact_url"] = reverse("artifacts-detail", args=[artifact.pk])
```
mdellweg, line 399: I don't see these being used.

mdellweg (approved):
Thoughts on (followup?) improvements:
- Allow the task to be called on a repository version.
- Count repaired and failed collections in a ProgressReport.

Issue #1175, opened 2022-08-17
Collection Upload doesn't raise a pretty error when uploading a duplicate Artifact.


gerrod3:
**Version**
0.14 and below

**Describe the bug**
When uploading a collection we *currently* don't allow duplicate files to be uploaded, so if the artifact exists it should reject the request. Currently this behavior is technically followed, but the error comes in the import task instead of at request time.

**Expected behavior**
400 error returned when trying to upload a duplicate collection with error message: "Artifact already exists"

**Additional context**
I'm going to try to unify the behavior more when I finished moving Namespace over.

PR #1176, opened 2022-08-17
Refactor CollectionVersion Upload to use pulpcore machinery 

gerrod3:
fixes: #1175

### DIFF
--- a/pulp_ansible/app/galaxy/mixins.py
+++ b/pulp_ansible/app/galaxy/mixins.py
@@ -19,3 +19,9 @@ class UploadGalaxyCollectionMixin:
             kwargs["repository_pk"] = repository.pk
 
         return dispatch(import_collection, exclusive_resources=locks, kwargs=kwargs)
+
+    def get_deferred_context(self, request):
+        context = {}
+        if "file" in request.data:
+            context["filename"] = request.data["file"].name
+        return context
--- a/pulp_ansible/app/galaxy/serializers.py
+++ b/pulp_ansible/app/galaxy/serializers.py
@@ -6,6 +6,7 @@ from drf_spectacular.types import OpenApiTypes
 from rest_framework.reverse import reverse
 from rest_framework import serializers
 
+from pulpcore.plugin.models import Artifact
 from pulp_ansible.app.models import Collection, CollectionVersion, Role
 from pulp_ansible.app.galaxy.v3.serializers import CollectionMetadataSerializer
 
@@ -187,3 +188,10 @@ class GalaxyCollectionUploadSerializer(serializers.Serializer):
     file = serializers.FileField(
         help_text=_("The file containing the Artifact binary data."), required=True
     )
+
+    def validate(self, data):
+        """Ensure duplicate artifact isn't uploaded."""
+        sha256 = data["file"].hashers["sha256"].hexdigest()
+        artifact = Artifact.objects.filter(sha256=sha256).first()
+        if artifact:
+            raise serializers.ValidationError(_("Artifact already exists"))
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -26,11 +26,14 @@ from rest_framework import viewsets, views
 from rest_framework.exceptions import NotFound
 from rest_framework import status
 
-from pulpcore.plugin.exceptions import DigestValidationError
-from pulpcore.plugin.models import PulpTemporaryFile, Content
+from pulpcore.plugin.models import Content
 from pulpcore.plugin.serializers import AsyncOperationResponseSerializer
-from pulpcore.plugin.viewsets import BaseFilterSet, OperationPostponedResponse
-from pulpcore.plugin.tasking import add_and_remove, dispatch
+from pulpcore.plugin.viewsets import (
+    BaseFilterSet,
+    OperationPostponedResponse,
+    SingleArtifactContentUploadViewSet,
+)
+from pulpcore.plugin.tasking import add_and_remove, dispatch, general_create
 
 from pulp_ansible.app.galaxy.v3.exceptions import ExceptionHandlerMixin
 from pulp_ansible.app.galaxy.v3.serializers import (
@@ -52,8 +55,9 @@ from pulp_ansible.app.models import (
     DownloadLog,
 )
 from pulp_ansible.app.serializers import (
-    CollectionOneShotSerializer,
     CollectionImportDetailSerializer,
+    CollectionOneShotSerializer,
+    CollectionVersionUploadSerializer,
 )
 
 from pulp_ansible.app.galaxy.mixins import UploadGalaxyCollectionMixin
@@ -458,13 +462,15 @@ class UnpaginatedCollectionViewSet(CollectionViewSet):
 
 
 class CollectionUploadViewSet(
-    ExceptionHandlerMixin, viewsets.GenericViewSet, UploadGalaxyCollectionMixin
+    ExceptionHandlerMixin, UploadGalaxyCollectionMixin, SingleArtifactContentUploadViewSet
 ):
     """
     ViewSet for Collection Uploads.
     """
 
-    serializer_class = CollectionOneShotSerializer
+    queryset = None
+    endpoint_pieces = None
+    serializer_class = CollectionVersionUploadSerializer
     pulp_tag_name = "Pulp_Ansible: Artifacts Collections V3"
 
     DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
@@ -472,6 +478,16 @@ class CollectionUploadViewSet(
     def urlpattern(*args, **kwargs):
         """Return url pattern for RBAC."""
         return "pulp_ansible/v3/collections/upload"
+
+    def _dispatch_upload_collection_task(self, args=None, kwargs=None, repository=None):
+        """
+        Dispatch an Upload Collection creation task.
+        """
+        locks = []
+        if repository:
+            locks.append(repository)
+
+        return dispatch(general_create, exclusive_resources=locks, args=args, kwargs=kwargs)
 
     @extend_schema(
         description="Create an artifact and trigger an asynchronous task to create "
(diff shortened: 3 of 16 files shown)


pulp_ansible/app/galaxy/views.py
```diff
@@ -180,6 +182,11 @@ def post(self, request, path):
         )
         serializer.is_valid(raise_exception=True)
 
+        sha256 = serializer.validated_data["file"].hashers["sha256"].hexdigest()
+        artifact = Artifact.objects.filter(sha256=sha256).first()
+        if artifact:
+            raise serializers.ValidationError(_("Artifact already exists"))
+
```
mdellweg, line 189: I think, this belongs in the serializer.validate.

pulp_ansible/app/serializers.py
```diff
@@ -383,7 +383,7 @@ def validate(self, data):
         sha256 = data["file"].hashers["sha256"].hexdigest()
         artifact = Artifact.objects.filter(sha256=sha256).first()
         if artifact:
-            ValidationError(_("Artifact already exists"))
+            raise ValidationError(_("Artifact already exists"))
```
mdellweg, line 386: :+1:

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
```diff
@@ -309,27 +309,17 @@ def test_collection_download(artifact, pulp_client, collection_detail):
     assert f.content == tarball
 
 
-def test_collection_upload_repeat(pulp_client, known_collection, pulp_dist):
+def test_collection_upload_repeat(pulp_client, artifact, pulp_dist, collection_upload):
```
mdellweg, line 312: what kind of a fixture is "artifact"? If it's a pulp_ansible one can we name it `collection_artifact`?

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
```diff
@@ -196,9 +195,10 @@ def test_collection_upload(collection_upload):
     assert "finished_at" in collection_upload
     assert "messages" in collection_upload
 
-    for key, value in collection_upload.items():
-        if key in COLLECTION_METADATA.keys():
-            assert COLLECTION_METADATA[key] == value, collection_upload
+    # TODO: Add this back when namespace, name, and version are apart of the CollectionImport
+    # for key, value in collection_upload.items():
+    #     if key in COLLECTION_METADATA.keys():
+    #         assert COLLECTION_METADATA[key] == value, collection_upload
```
mdellweg, line 201: If these do not assert anymore, i think we are seeing a change behavioural change unrelated to the bug / improvement advertised in the PR.

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
mdellweg: Are the changes in this file supposed to be part of this PR at all?

gerrod3: Yes, I realized the tests for collection upload were wrong and incorrectly passing when it shouldn't.

pulp_ansible/tests/functional/api/collection/v3/test_collection.py
gerrod3: The asserts were never being ran in the first place. `collection_upload` is a `CollectionImport` response and does not have the fields in `COLLECTION_METADATA` : `namespace, name, version`. So commenting them out does nothing I could have left it as is.

mdellweg: Are you saying the keys should have been there? Then we should not guard with the "if". If not can we just remove the block completely?

gerrod3: No idea if they should have been there. But they will be once I add `Namespace`s.

pulp_ansible/app/serializers.py
```diff
@@ -361,130 +373,259 @@ class Meta:
...
+        data.update(collection_info)
 
         return data
 
+    def create(self, validated_data):
+        """Final step in creating the CollectionVersion."""
+        tags = validated_data.pop("tags")
+        origin_repository = validated_data.pop("origin_repository")
+        # Create CollectionVersion from its metadata and adds to repository if specified
+        content = super().create(validated_data)
```
mdellweg, line 489: Does it make sense to add the add-to-repository-in-a-subtask feature to pulpcore?

pulp_ansible/app/tasks/upload.py
```diff
@@ -0,0 +1,88 @@
...
+from galaxy_importer.collection import import_collection
+from pulpcore.plugin.models import Task
+
+from pulp_ansible.app.models import Collection, CollectionImport, Tag
+from pulp_ansible.app.tasks.utils import CollectionFilename, get_file_obj_from_tarball
+
+log = logging.getLogger(__name__)
+
+
+def process_collection_artifact(artifact, namespace, name, version):
```
mdellweg, line 17: This is the thing we call in `deferred_validate` now? Great achievement!

pulp_ansible/app/serializers.py
```diff
@@ -361,130 +373,259 @@ class Meta:
...
+    )
 
     def validate(self, data):
-        """Validate that we have a file or can create one."""
-        if "artifact" in data:
-            raise serializers.ValidationError(_("Only 'file' may be specified."))
+        """Check and set the namespace, name & version."""
+        # This validation is for creating CollectionVersions
+        if self.instance:
+            return data
```
mdellweg, line 427: Shouldn't we still call `super` first?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -454,17 +457,30 @@ def urlpattern(*args, **kwargs):
...
 
-    serializer_class = CollectionOneShotSerializer
+    queryset = CollectionVersion.objects.all()
+    serializer_class = CollectionVersionUploadSerializer
     pulp_tag_name = "Pulp_Ansible: Artifacts Collections V3"
 
     DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
 
+    _declared_fields = serializer_class._declared_fields
+    INLINE_SERIALIZER = inline_serializer(
```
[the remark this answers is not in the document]
gerrod3: So I can hide the Pulp-specific fields: `artifact`, `upload`, and `repository` from the generated schema. I tried to just not include them in the `Meta` class, but then it would error when `artifact` or `repository` was set by the viewset's logic.

pulp_ansible/app/serializers.py
gerrod3: I'm thinking about it. I already know how I am going to implement the feature in pulp_ansible, but putting it into pulpcore would make it much more simple.

pulp_ansible/app/serializers.py
gerrod3: I'm thinking of removing this check and instead over-riding the `update` method to raise an error if called. `validate` should only be called on `create` or `update` so if `self.instance` is set then `update` will be called and we shouldn't allow it.

mdellweg: Hmm. Hard question. From the API we should never allow updates. But we want to allow for a repair command. And it would be super neat if that repair command could just use this serializer to actually _call_ the `process_collection` part again. No?

gerrod3: Yes that would be a good use. There would still need to be other fields that would need to be set if you wanted to call `super` in that case (`artifact`, `expected_namespace/name/version/`) else the serializer would fail validation. I'll remove the check and put a TODO comment to revisit this for a future issue (#1106).

gerrod3 (commented):
@mdellweg @newswangerd Ready for review.

Summary of changes:

- `/ansible/collections/` & Galaxy V2 Collection upload have been deprecated.
- `CollectionVersionUploadSerializer` now has all the old fields of `CollectionOneShotSerializer` and the entirety of the collection upload logic needed.
- Galaxy V3 Collection upload is now a `SingleArtifactContentUploadViewSet` and uses the `CollectionVersionUploadSerializer` to perform the collection upload.
- `CollectionVersionSerializer` now has `CollectionVersionUploadSerializer` as a base class so the Pulp CV api uses the same base upload logic as the Galaxy V3 api.
- Pulp CV api now accepts `artifact` and `upload` fields when uploading a collection.

pulp_ansible/app/galaxy/mixins.py
```diff
@@ -19,3 +19,11 @@ def _dispatch_import_collection_task(self, temp_file_pk, repository=None, **kwar
             kwargs["repository_pk"] = repository.pk
 
         return dispatch(import_collection, exclusive_resources=locks, kwargs=kwargs)
+
+    def init_content_data(self, serializer, request):
+        """Add serializer's validated data to the task payload."""
+        # Honestly this seems like a bug in pulpcore
```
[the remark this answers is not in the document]
mdellweg: I think this is by design. As the task will call the same serializer again (including the deferred validate steps) so it needs the raw data.

pulp_ansible/app/galaxy/serializers.py
```diff
@@ -187,3 +188,10 @@ class GalaxyCollectionUploadSerializer(serializers.Serializer):
     file = serializers.FileField(
         help_text=_("The file containing the Artifact binary data."), required=True
     )
+
+    def validate(self, data):
+        """Ensure duplicate artifact isn't uploaded."""
+        sha256 = data["file"].hashers["sha256"].hexdigest()
+        artifact = Artifact.objects.filter(sha256=sha256).first()
+        if artifact:
+            raise serializers.ValidationError(_("Artifact already exists"))
```
mdellweg, line 197: Is it safe to just take the existing artifact at this time? (The user uploaded the proper file, right?)

pulp_ansible/app/galaxy/v3/views.py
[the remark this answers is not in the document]
mdellweg: It think, this is dangerous. Just not adding them to the docs, means that the serializer still accepts them. And it shouldn't!?

pulp_ansible/app/serializers.py
```diff
@@ -361,151 +373,304 @@ class Meta:
...
+        # Super will call deferred_validate on second call
```
mdellweg, line 428: ... "in task context"; "not in request context"

pulp_ansible/app/serializers.py
```diff
@@ -361,151 +373,304 @@ class Meta:
...
+        data = super().validate(data)
 
-        if "request" not in self.context:
-            data = self.deferred_validate(data)
+        # First time calling validate
+        if "request" in self.context:
```
mdellweg, line 432: You are basically inverting the "deferred_validate" logic here. Can we just start with the raw data again in the task, so we can be sure that we validate everything in one go?

pulp_ansible/app/serializers.py
```diff
@@ -361,151 +373,304 @@ class Meta:
...
+                    collection = parse_collection_filename(filename)
+                except ValueError:
+                    raise ValidationError(
+                        _("Failed to parse Collection file upload '{}'").format(filename)
+                    )
+                data["expected_namespace"] = collection.namespace
+                data["expected_name"] = collection.name
+                data["expected_version"] = collection.version
+
+            if CollectionVersion.objects.filter(**{f: data[f"expected_{f}"] for f in fields}):
```
mdellweg, line 453:
```suggestion
            if CollectionVersion.objects.filter(**{f: data[f"expected_{f}"] for f in fields}).exists():
```

pulp_ansible/app/serializers.py
```diff
@@ -361,151 +373,304 @@ class Meta:
...
+                raise ValidationError(
```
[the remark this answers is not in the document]
mdellweg: The existing one? Only if we wanted to use it, IMHO.

pulp_ansible/app/serializers.py
gerrod3: Yes, I can rewrite it. Just need to take into account not having the artifact filename on the second validate.

pulp_ansible/app/galaxy/v3/views.py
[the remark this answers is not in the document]
gerrod3: You are right. I'll add validation for this serializer.

pulp_ansible/app/galaxy/serializers.py
mdellweg: Do we plan to (eventually?) change that task then too?

gerrod3: I eventually plan to remove the task as it is duplicate code now and eventually remove the deprecated endpoints that call it. Git sync still uses it as well so that will also require a separate PR to refactor it.

pulp_ansible/app/serializers.py
```diff
@@ -361,151 +373,298 @@ class Meta:
...
+    necessary. This serializer is meant to be compliant with
+    ``pulpcore.plugin.viewsets.SingleArtifactContentUploadViewSet`` and thus follows these steps on
+    creation:
+
+    1. ``SingleArtifactContentUploadViewSet.create()`` calls ``validate()`` with request in context
+    2. Task payload is created, converting uploaded-file (if present) to an artifact
+    3. ``pulpcore.plugin.tasks.general_create`` is dispatched with this serializer, the task payload
+        and a deferred context determine by the viewset (default an empty context).
+    4. ``general_create`` calls ``validate()`` again with deferred context now.
+        ``deferred_validate()`` is now called and the upload object (if present) is converted to an
```
mdellweg, line 389: Isn't this where we also call the import_collection to extract all the metadata? That process that may be offloaded into a different application cluster?

pulp_ansible/app/serializers.py
```diff
@@ -361,151 +373,304 @@ class Meta:
...
+
+    def validate(self, data):
+        """Run super() validate if creating, else return data."""
+        if self.creating:
+            return super().validate(data)
+        return data
+
+    def is_valid(self, raise_exception=False):
+        """
+        Allow this serializer to be used for validating before saving a model.
```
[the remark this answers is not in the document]
mdellweg: I'm still not sure we need this. Is it there to support code we plan to deprecate/delete? If i read correctly it allows the serializer to consume validated data, right? I do not think they are designed that way.
My thinking is if we hand the same raw request data that we get in the viewset to the task, then the serializer in the task can validate in the same way plus the additional deferred_validate step. No need for any serializer field to behave differently.
Not blocking on this now. But can you confirm that this will fall for another refactor to come?

mdellweg (approved):
I'm still a bit confused by the big number of serializers we have for creating CollectionVersions. Maybe we need individual ones for each version of the api, and that is fine.
We still need confirmation from the galaxy side before we can merge this.

pulp_ansible/app/serializers.py
gerrod3: Yes, I wrote these steps describing how the pulpcore machinery works since I feel it is not clear from just looking at the serializer what is going on.

pulp_ansible/app/serializers.py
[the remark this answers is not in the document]
gerrod3: Yes it most likely will be removed once the `import_collection` code is finally removed. `read_only` fields are not validated when initialized in a serializer and they don't appear in the `validate` method.

mdellweg (approved):

PR #1178, opened 2022-08-18
Reset sync timestamps on more remote attributes

### LINKED ISSUE #1177: Repositories last_synced_metadata_time is not reset on all remote updates
This can lead to unexpected optimized syncs.

mdellweg:
fixes #1177

### DIFF
--- a/pulp_ansible/app/models.py
+++ b/pulp_ansible/app/models.py
@@ -5,6 +5,7 @@ from django.db import models
 from django.db.models import UniqueConstraint, Q
 from django.contrib.postgres import fields as psql_fields
 from django.contrib.postgres import search as psql_search
+from django_lifecycle import AFTER_UPDATE, BEFORE_UPDATE, hook
 
 from pulpcore.plugin.models import (
     BaseModel,
@@ -293,6 +294,16 @@ class CollectionRemote(Remote):
             self._download_factory = AnsibleDownloaderFactory(self)
             return self._download_factory
 
+    @hook(
+        AFTER_UPDATE,
+        when_any=["url", "requirements_file", "sync_dependencies", "signed_only"],
+        has_changed=True,
+    )
+    def _reset_repository_last_synced_metadata_time(self):
+        AnsibleRepository.objects.filter(
+            remote_id=self.pk, last_synced_metadata_time__isnull=False
+        ).update(last_synced_metadata_time=None)
+
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
 
@@ -383,6 +394,10 @@ class AnsibleRepository(Repository):
                 )
 
                 new_version.remove_content(deprecations)
+
+    @hook(BEFORE_UPDATE, when="remote", has_changed=True)
+    def _reset_repository_last_synced_metadata_time(self):
+        self.last_synced_metadata_time = None
 
 
 class AnsibleDistribution(Distribution):
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -62,7 +62,7 @@ from pulp_ansible.app.models import (
     CollectionVersionSignature,
     Tag,
 )
-from pulp_ansible.app.serializers import CollectionRemoteSerializer, CollectionVersionSerializer
+from pulp_ansible.app.serializers import CollectionVersionSerializer
 from pulp_ansible.app.tasks.utils import (
     RequirementsFileEntry,
     get_file_obj_from_tarball,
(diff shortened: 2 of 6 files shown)


pulp_ansible/app/models.py
```diff
@@ -256,6 +257,16 @@ def download_factory(self):
             self._download_factory = AnsibleDownloaderFactory(self)
             return self._download_factory
 
+    @hook(
+        AFTER_UPDATE,
+        when_any=["url", "requirements_file", "sync_dependencies", "signed_only"],
+        has_changed=True,
+    )
+    def _reset_repository_last_synced_metadata_time(self):
```
ipanova, line 265: This approach does not seem to lock repositories, I believe this is problematic.

ipanova: I see at the bottom async_reserved_resources :+1:

pulp_ansible/app/viewsets.py
```diff
@@ -385,25 +385,13 @@ class CollectionRemoteViewSet(RemoteViewSet):
...
-        async_result = dispatch(
-            update_collection_remote,
-            exclusive_resources=lock,
-            args=(pk,),
-            kwargs={"data": request.data, "partial": partial},
+            remote_id=instance.pk, last_synced_metadata_time__isnull=False
         )
-        return OperationPostponedResponse(async_result, request)
+        lock.extend(repos)
+        return repos
```
ipanova, line 394: return lock

ipanova (commented):
looks good, left one comment https://github.com/pulp/pulp_ansible/pull/1178/files#r948987541

pulp_ansible/app/viewsets.py
mdellweg: right!

mdellweg:
I added a second pr with what i think improves/fixes the optimize logic when using "not-the-repo-remote" and also provide the correct date when performing a "not-optimized sync".

pulp_ansible/app/tasks/collections.py
```diff
@@ -504,13 +465,14 @@ def __init__(self, remote, repository, is_repo_remote, deprecation_before_sync,
         self.already_synced = set()
         self._unpaginated_collection_metadata = None
         self._unpaginated_collection_version_metadata = None
+        self.optimize = optimize
```
gerrod3, line 468: I think `self.is_repo_remote` is no longer used throughout the class looking at `_should_we_sync()`. If it is we can remove it.

pulp_ansible/app/viewsets.py
```diff
@@ -385,25 +385,13 @@ class CollectionRemoteViewSet(RemoteViewSet):
...
 
-    @extend_schema(
-        description="Trigger an asynchronous update task",
-        responses={202: AsyncOperationResponseSerializer},
-    )
-    def update(self, request, pk, **kwargs):
-        """Update remote."""
-        partial = kwargs.pop("partial", False)
-        lock = [self.get_object()]
+    def async_reserved_resources(self, instance):
```
gerrod3, line 388: This is very cool, I did not know this existed. Does this mean it will lock all those repositories on delete as well? (does delete lock on the instance?)

mdellweg: That's a good question. Need to check.

mdellweg: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/base.py#L543

Yes, it does.

pulp_ansible/app/models.py
```diff
@@ -347,6 +358,10 @@ def finalize_new_version(self, new_version):
 
                 new_version.remove_content(deprecations)
 
+    @hook(BEFORE_UPDATE, when="remote", has_changed=True)
+    def _reset_repository_last_synced_metadata_time(self):
+        self.last_synced_metadata_time = None
+
```
gerrod3, line 364: I don't think this triggers when the remote is deleted. A hook on the remote being deleted is also needed I think.

ipanova: if the remote is deleted then the repo loses it's reference to it and it becomes un-syncable anyway..

mdellweg: good observation

pulp_ansible/tests/functional/api/collection/test_sync.py
```diff
@@ -280,3 +281,64 @@ def test_v2_simple_dependency(self):
 
         content = self.cv_api.list(repository_version=f"{repo.pulp_href}versions/1/", name="posix")
         self.assertNotEqual(content.count, 0)
+
+
+@pytest.mark.skip("Skipped until fixture metadata has a published date")
```
mdellweg, line 286: Turns out galaxy does not provide the published info. So it cannot be tested this way. I would still keep the test, and we can maybe find another source to test this.
OTOH, is the "metadata" a specific file? We could just save it's digest and compare that...

gerrod3: If you wanted to test this you would have to perform a Pulp to Pulp sync. The published metadata is only apart of the v3 api.

newswangerd (approved):

PR #1179, opened 2022-08-18
Convert git tests to pytest

gerrod3:
Separated out from #1148

### DIFF
--- a/pulp_ansible/tests/functional/api/git/test_crud_remotes.py
+++ b/pulp_ansible/tests/functional/api/git/test_crud_remotes.py
@@ -1,40 +1,49 @@
-# coding=utf-8
 """Tests that CRUD remotes."""
+import pytest
 from pulp_smash.pulp3.bindings import monitor_task
 
 from pulp_ansible.tests.functional.utils import (
     gen_ansible_remote,
-    TestCaseUsingBindings,
 )
-from pulp_ansible.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+from pulpcore.client.pulp_ansible import ApiException
 
 
-class CRUDAnsibleGitRemotesTestCase(TestCaseUsingBindings):
-    """CRUD Ansible Git remotes."""
+@pytest.mark.parallel
+def test_crud_git_remote(ansible_remote_git_api_client, gen_object_with_cleanup):
+    """Create, Read, Partial Update, and Delete an Ansible Git remote."""
+    # Create Git remote
+    body = gen_ansible_remote(url="https://github.com/geerlingguy/ansible-role-adminer.git")
+    remote = gen_object_with_cleanup(ansible_remote_git_api_client, body)
 
-    def test_01_crud_remote(self):
-        """Create, Read, Partial Update, and Delete an Ansible Git remote."""
-        body = gen_ansible_remote(url="https://github.com/geerlingguy/ansible-role-adminer.git")
-        remote = self.remote_git_api.create(body)
-        self.addCleanup(self.remote_git_api.delete, remote.pulp_href)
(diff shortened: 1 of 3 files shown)


pulp_ansible/tests/functional/api/git/test_crud_remotes.py
```diff
@@ -1,40 +1,50 @@
...
-            self.assertEqual(body[k], getattr(remote, k))
+from pulpcore.client.pulp_ansible import ApiException
+
+
+@pytest.mark.parallel
+def test_crud_git_remote(ansible_remote_git_api_client):
+    """Create, Read, Partial Update, and Delete an Ansible Git remote."""
+    # Create Git remote
+    body = gen_ansible_remote(url="https://github.com/geerlingguy/ansible-role-adminer.git")
+    remote = ansible_remote_git_api_client.create(body)
```
mdellweg, line 16: This should be generate with cleanup. Even if the test set's out to remove it later. But if we fail in the middle we want the cleanup still.

pulp_ansible/tests/functional/conftest.py
```diff
@@ -99,6 +114,18 @@ def _ansible_repo_factory(**kwargs):
     return _ansible_repo_factory
 
 
+@pytest.fixture
+def sync_repo_factory(ansible_repo_api_client):
```
mdellweg, line 118: Would it be useful if this factory created the repository and sync it and then return?
In the current state it looks like a sync_helper to me. Also be careful with the name. pytest put's all fixture names in the same scope.

gerrod3: Renamed to `ansible_sync_factory`. Also added auto repo creation if not supplied.

pulp_ansible/tests/functional/conftest.py
```diff
@@ -99,6 +114,20 @@ def _ansible_repo_factory(**kwargs):
...
+    """A factory to perform a sync on an Ansible Repository and return its updated data."""
+
+    def _sync(ansible_repo=None, **kwargs):
+        body = AnsibleRepositorySyncURL(**kwargs)
+        if ansible_repo is None:
+            ansible_repo = ansible_repo_factory()
+        monitor_task(ansible_repo_api_client.sync(ansible_repo.pulp_href, body).task)
+        return ansible_repo_api_client.read(ansible_repo.pulp_href)
+
+    yield _sync
```
mdellweg, line 128: No need to make it a generator if we don't have cleanup code.
```suggestion
    return _sync
```

gerrod3: Would this apply to the other factories that just use `gen_object_with_cleanup` and have no other cleanup code as well?

mdellweg: yes, i think so.

pulp_ansible/tests/functional/api/git/test_crud_remotes.py
```diff
@@ -1,40 +1,50 @@
...
+    update_body = {"url": "https://github.com/geerlingguy/ansible-role-ansible.git"}
+    remote_update = ansible_remote_git_api_client.partial_update(remote.pulp_href, update_body)
+    monitor_task(remote_update.task)
+    remote = ansible_remote_git_api_client.read(remote.pulp_href)
+    assert remote.url == update_body["url"]
+    assert remote.metadata_only is False
+
+    # Ensure Git remote doesn't have a download policy
+    with pytest.raises(AttributeError):
+        getattr(remote, "policy")
```
mdellweg, line 33: How about `assert not hasattr()`

mdellweg (approved):

PR #1222, opened 2022-09-01
Switch token attribute to be encryped and hidden

### LINKED ISSUE #1221: Ansible tokens stored in plaintext
The collection remote for pulp_ansible stores tokens in plaintext instead of using pulp's encrypted field (https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/models.py#L234) and exposes them in read/write mode via the API (https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/serializers.py#L170) instead of marking it as write only.

mdellweg:
fixes #1221


gerrod3 (approved):

newswangerd (approved):

bmbouter:
@mdellweg did you want to add a test for this?

pulp_ansible/tests/functional/api/collection/test_remote.py
```diff
@@ -29,13 +29,14 @@ def test_update_auth_url(self):
             auth_url="https://example.com",
         )
         remote = self.remote_collection_api.create(body)
+        self.addCleanup(self.remote_collection_api.delete, remote.pulp_href)
+        assert not hasattr(remote, "token")
```
mdellweg, line 33: Added this for testing.

PR #1229, opened 2022-09-05
Fix the expected content-type in tests

mdellweg:
I have no idea if this is the right solution. Does any client tooling, whatsoever care about the reported content-type? If not, this assert is not even a useful constraint, and we should plainly remove it.
Who would know such a thing?


mdellweg:
This is probably the root cause: https://github.com/pulp/pulpcore/pull/3093

gerrod3 (approved):

PR #1237, opened 2022-09-12
Add client configuration api endpoint

### LINKED ISSUE #740: Implement configuration endpoint
Implement `v3/plugin/ansible/client-configuration/` api endpoint that returns

```
{
    "default_distribution_path": ANSIBLE_DEFAULT_DISTRIBUTION_PATH
}
```

`default_distribution_path` should return the value for `settings.ANSIBLE_DEFAULT_DISTRIBUTION_PATH`

newswangerd:
fixes #740

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -40,6 +40,7 @@ from pulp_ansible.app.galaxy.v3.serializers import (
     CollectionVersionListSerializer,
     RepoMetadataSerializer,
     UnpaginatedCollectionVersionSerializer,
+    ClientConfigurationSerializer,
 )
 from pulp_ansible.app.models import (
     AnsibleCollectionDeprecated,
@@ -922,3 +923,23 @@ def redirect_view_generator(actions, url, viewset, distro_view=True, responses={
             return self._get(request, *args, **kwargs)
 
     return GeneratedRedirectView.as_view(actions, url=url)
+
+
+class ClientConfigurationView(views.APIView):
+    """Return configurations for the ansible-galaxy client."""
+
+    DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
+
+    @extend_schema(responses=ClientConfigurationSerializer)
+    def get(self, request, *args, **kwargs):
+        """Get the client configs."""
+
+        data = ClientConfigurationSerializer(
+            {
+                "default_distribution_path": self.kwargs.get(
+                    "path", settings.ANSIBLE_DEFAULT_DISTRIBUTION_PATH
+                )
+            }
+        )
+
+        return Response(data.data)
--- a/pulp_ansible/tests/functional/api/collection/v3/test_client_configuration.py
+++ b/pulp_ansible/tests/functional/api/collection/v3/test_client_configuration.py
@@ -1,0 +1,13 @@
+def test_client_config_distro(
+    ansible_repo, ansible_distribution_factory, ansible_client_configuration_api_client
+):
+    distro = ansible_distribution_factory(ansible_repo)
+    res = ansible_client_configuration_api_client.get(path=distro.base_path)
+    assert res.default_distribution_path == distro.base_path
+
+
+def test_client_config_default_distro(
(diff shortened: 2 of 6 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -922,3 +923,23 @@ def update(self, request, *args, **kwargs):
...
+    DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
+
+    @extend_schema(responses=ClientConfigurationSerializer)
+    def get(self, request, *args, **kwargs):
+        """Get the client configs."""
+
+        data = ClientConfigurationSerializer(
+            {
+                "default_distribution_path": self.kwargs.get(
+                    "path", settings.ANSIBLE_DEFAULT_DISTRIBUTION_PATH
```
mdellweg, line 940: I think, we need to add the path as a query string to the schema.

pulp_ansible/tests/functional/api/collection/v3/test_client_configuration.py
```diff
@@ -0,0 +1,8 @@
+def test_client_default_base_path(
+    ansible_repo, ansible_distribution_factory, ansible_client_configuration_api_client
+):
+    distro = ansible_distribution_factory(ansible_repo)
+
+    res = ansible_client_configuration_api_client.get(path=distro.base_path)
```
mdellweg, line 6: How about another get without the path argument?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -922,3 +923,23 @@ def update(self, request, *args, **kwargs):
...
 
     return GeneratedRedirectView.as_view(actions, url=url)
+
+
+class ClientConfigurationView(views.APIView):
+    """Return configurations for the ansible-galaxy client."""
+
+    DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
+
+    @extend_schema(responses=ClientConfigurationSerializer)
```
mdellweg, line 933:
```suggestion
    @extend_schema(
        responses=ClientConfigurationSerializer,
        parameters=[
            OpenApiParameter(
                name='path',
                description='Distribution path',
                required=False,
                type=str,
                location=OpenApiParameter.QUERY,
    )
```

pulp_ansible/tests/functional/api/collection/v3/test_client_configuration.py
newswangerd: `ANSIBLE_DEFAULT_DISTRIBUTION_PATH` is none by default, and since I can't change the settings during tests, I don't think it's possible to the configuration setting for the default distribution

pulp_ansible/app/galaxy/v3/views.py
newswangerd: the path kwarg refers to this: https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/urls.py#L211

I don't think that goes in query params, right?

pulp_ansible/app/galaxy/v3/views.py
mdellweg: The path being part of the url pattern was not completely obvious to me at this point. Sorry.

pulp_ansible/tests/functional/api/collection/v3/test_client_configuration.py
mdellweg: What happens if we set it? Will all the tests break?

I'm currently pitching this to cover setting _and_ not setting feature flags in the ci:
https://github.com/pulp/plugin_template/pull/692

mdellweg:
Now that i "think" I fully understand the change, it looks like we provide an endpoint that will return a dict with a single key whose value is part of the url we used to get to the endpoint in the first place. Just out of curiosity, do we plan to add more client settings there in the future?

mdellweg (approved):
I'd be happy if you'd try the test suggestion, but I don't want to block this.
You answered all my other questions.

newswangerd:
> .

I'll give it a shot. If the API doesn't 404, then it should return None

Issue #1245, opened 2022-09-16
Syncing signed collections stores signatures as bytes


newswangerd:
**Version**
pulcore 3.21
pulp_ansible 0.15

**Describe the bug**
Signatures loaded during syncs are saved as bytes.

**To Reproduce**
Sync a signed collection. The signature shows up a b'' string on the API:

```
            "signature": "b'-----BEGIN PGP SIGNATURE-----\\nVersion: GnuPG v1\\n\\niQIVAwUAYyJcZ+GkvXCKgoqtAQhkbA//e2VrbAikRwO82C2ljuIOnjpN9OO6kgYN\\nHd6eABTWJ2m/N6jsHnINXtXnRYNWorZNnbNPnmZ0GtfUnh57U55I2y4COyZGOWq+\\nzbiNo2l44o+jiXomazr1f75beu92C0v+kW+nvgdre7OqbkNHI8295GiqiockcwdJ\\nxPCfojqUURLWqZumS8oWg1owSQPePiL7ZGsOfxBnRWQDyvqs2abJmrQY73bnhEBx\\nLsApx4W1rFi7yD012ejNEHsuguUpGcQxibyuHnxgP5ZDOtHFZbmAh0d5j29GnET/\\nevVrB/8HDoNivv4SD2zuV9LftbxR5VwGO1H5GtK4D7SSl0bHAloB2tZAgQrnkBif\\n3YjnA0gQwB6x5DaYFkZzlHkTwPvcqNM6ts827yHTiqf1jtVBrgjy8qYait9v251j\\nMQnT41x4XIVBIC2yiavF/kIIE9MOVqd0DZqLXt0mcXxvaK7ZHKhWed5FImEMzbgR\\nr+yI/YdRtDvNUaoqPNjtvXBf68t2/zJqtIf66CTYmguv2AqRSNQZBp902T+rX1s6\\n8NRNgDh6JQqy1oq7HwAZeiRFSYyShvGky76lWzNUWdX+aLdL3533hJwMjpoRe3/s\\nBiN61nR9DpvkY8fXVffwTJ+wQVVfRvclxVEkAKw5tczJU/aLiWp7qxkbDMlBYOTK\\nsjd1+WBrGTw=\\n=WIPN\\n-----END PGP SIGNATURE-----\\n'",
            "pubkey_fingerprint": "39B5D759361A97464C564A8FE1A4BD708A828AAD",
```

**Expected behavior**
The signature should be a regular string

PR #845, opened 2022-02-24
Add collection version signature model resource

mdellweg:
fixes #844

PR #1246, opened 2022-09-16
Fix signatures not being stored as text on syncs

gerrod3:
fixes: #1245

### DIFF
--- a/CHANGES/1245.misc
+++ b/CHANGES/1245.misc
@@ -1,0 +1,1 @@
+``CollectionVersionSignature`` data is now stored as plain text instead of bytes.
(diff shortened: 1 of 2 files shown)


CHANGES/1245.misc
```diff
@@ -0,0 +1 @@
+``CollectionVersionSignature`` data is now stored as plain text instead of bytes.
```
mdellweg, line 1: Isn't this a bugfix?

Do you think we need a data-repair command?

newswangerd: It hasn't shipped with anything yet, so this should be fine.

newswangerd (approved):

CHANGES/1245.misc
mdellweg: So do we need to backport any of this?

mdellweg: I believe not. This is really a followup to https://github.com/pulp/pulp_ansible/pull/845 .

PR #1252, opened 2022-09-20
Disable single commit check.

jctanner:
[noissue]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,21 +11,6 @@ on: {pull_request: {branches: ['*']}}
 jobs:
 
 
-  single_commit:
-    name: Assert single commit
-    if: github.base_ref == 'main'
-    steps:
-    - name: Checkout
-      uses: actions/checkout@v3
-      with:
-        fetch-depth: 30
-    - name: Checkout main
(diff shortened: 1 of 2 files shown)


awcrosby:
I believe what works pretty well in galaxy_ng… the PR can have multiple commits that show the history of things being worked, and shows CI passing/failing per commit, and in our repo settings we only have checked the `Allow squash merging` option which ensures a single commit per PR

mdellweg:
The point is, we don't want to enforce a single commit. And yes, i'm sad that this check cannot report as yellow instead of red. It's not marked as "required" though. Anyway, a change like this needs some discussion with the team.
cc @pulp/ansible

mdellweg:
I think you can edit the commit message on squash. So the person pushing the button would be in charge.

bmbouter:
@jctanner why close?

bmbouter:
I made this a formal proposal here: https://discourse.pulpproject.org/t/removing-the-one-commit-check-in-the-ci/639 The idea is to have pulp_ansible try it out and then we can report how we like it.

We need to determine which setting to try first for the allow squash merging config...

mdellweg (approved):

PR #1254, opened 2022-09-20
Enforce TDD.

jctanner:
more context: https://github.com/ansible/galaxy_ng/pull/1236

[noissue]

### DIFF
--- a/.github/tdd.py
+++ b/.github/tdd.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+
+import json
+import logging
+import os
+import subprocess
+import sys
+
+ORG_NAME = "pulp"
+REPO_NAME = "pulp_ansible"
+APP_NAME = "pulp_ansible"
+PRIMARY_BRANCH = "main"
+
+
+def get_current_branch():
+    pid = subprocess.run(
+        "git branch --show-current", shell=True, stdout=subprocess.PIPE, check=True
+    )
+    branch_name = pid.stdout.decode("utf-8")
+    branch_name = branch_name.strip()
+    return branch_name
+
+
+def get_changed_files(pr_branch, target_branch=PRIMARY_BRANCH):
+    if os.environ.get("GITHUB_PR_NUMBER"):
+        url = f"https://patch-diff.githubusercontent.com/raw/{ORG_NAME}/{REPO_NAME}"
+        url += f"/pull/{os.environ['GITHUB_PR_NUMBER']}.diff"
+        print(f"GET diff from {url}")
+        cmd = f"curl -s -o /tmp/pr.diff {url}"
+        pid = subprocess.run(cmd, shell=True)
+        assert pid.returncode == 0
+        with open("/tmp/pr.diff", "r") as f:
+            raw = f.read()
+        filenames = raw.split("\n")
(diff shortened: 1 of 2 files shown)


.github/tdd.py
```diff
@@ -0,0 +1,126 @@
...
+    if not pr_branch:
+        pr_branch = json.loads(os.environ.get("GITHUB_CONTEXT"))["ref"]
+    print(f"PR-BRANCH: {pr_branch}")
+
+    # branch the PR wants to change
+    target_branch = os.environ.get("GITHUB_BASE_REF", "master")
+    print(f"TARGET-BRANCH: {target_branch}")
+    if target_branch != "master":
+        print("TDD is only enforced on the master branch")
+        sys.exit(0)
```
gerrod3, line 117: Our default branch is named `main`.

.github/tdd.py
```diff
@@ -0,0 +1,126 @@
...
+
+    if not tests_found:
+        raise Exception("Tests should be added or modified with -every- PR!!!")
+
+
+def main():
+    logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+
+    # these users should not be checked
+    skip_users = ["dependabot[bot]", "patchback[bot]"]
```
gerrod3, line 96: Our automated CI (release process, plugin template updates) has [pulpbot](https://github.com/pulpbot) open PRs as well, think it should also be in this list.

.github/tdd.py
```diff
@@ -0,0 +1,126 @@
...
+        filenames = pid.stdout.decode("utf-8")
+        filenames = filenames.split("\n")
+        filenames = [x.strip() for x in filenames if x.strip()]
+    return filenames
+
+
+def verify_test_files_changed(changed_files):
+
+    # places where appplication source goes
+    app_paths = [f"{APP_NAME}/"]
```
gerrod3, line 53: I think there are some files that shouldn't require test changes.

Also, what should happen in the case where a thing is deprecated in `pulpcore.plugin` and the change is simply switching over to the new one?

gerrod3 (commented):
Overall I am in favor of this change, but I think there are some questions that need to be discussed first, mainly when this should be enforced.

mdellweg:
Sorry, but i think my verdict here will be a clear "no".

This check looks whether some files have been changed. This does not tell me anything about whether the added/removed/changed tests are meaningful. I could change a whitespace to make this check happy. And it may make a reviewer just nod on the green check mark.
Requesting a changed / added test is really the responsibility of the (human) reviewer. Happy to put this on a checklist though.

BTW: Test-*driven*-development is still something different to me.

awcrosby:
I’ll just add our rationale for adopting this in galaxy_ng… We took the mindset that almost all app changes should come with tests, this helped ensure tests were being thought about towards the start of the PR. It has really helped us increase the amount of tests written. Agreed test quality is still up to the author and reviewer.

We considered the ideal CI setup to be a tool to check that tests added in a PR relate to code touched in a PR. Or at least check coverage via a percentage of code that has been touched by tests, and for a given PR not let that percentage decrease. Even in both of those cases we would normally expect an edit to the dir where tests live. So due to complexity, we went with the simple check to start.

PR #1261, opened 2022-09-22
Update CI

mdellweg:
Configure branches to recieve regular CI updates

[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -61,7 +61,6 @@ pulp_settings:
   ansible_signature_require_verification: false
 pulp_settings_azure: null
 pulp_settings_s3: null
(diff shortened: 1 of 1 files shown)


template_config.yml
```diff
@@ -18,6 +18,7 @@ check_stray_pulpcore_imports: true
 cherry_pick_automation: false
 ci_env: {}
 ci_trigger: '{pull_request: {branches: [''*'']}}'
+ci_update_branches: ["0.15", , "0.14", "0.10", "0.7"]
```
mdellweg, line 21: I did some rough estimation here. What do we really need?

@pulp/ansible @pulp/galaxy

PR #1279, opened 2022-10-31
Use the new trigger to rebuild all existing search_vector values.

jctanner:
The previous change fixed the trigger, but it didn't actually "migrate" the old search_vector vaues. This change will force run the trigger on all existing collectionversions so that the search_vector is fixed on all of them.

### DIFF
--- a/pulp_ansible/app/migrations/0046_add_fulltext_search_fix.py
+++ b/pulp_ansible/app/migrations/0046_add_fulltext_search_fix.py
@@ -34,14 +34,10 @@ TS_VECTOR_SELECT = '''
     || setweight(to_tsvector(coalesce(description, '')), 'D')
 '''
 
-# Generates search vector for existing CollectionVersion objects in the database.
-POPULATE_COLLECTIONS_TS_VECTOR = f'''
-UPDATE ansible_collectionversion AS c
-SET search_vector = (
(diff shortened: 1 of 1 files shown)


pulp_ansible/app/migrations/0046_add_fulltext_search_fix.py
```diff
@@ -85,12 +82,11 @@ class Migration(migrations.Migration):
...
-        migrations.RunSQL(
-            sql=POPULATE_COLLECTIONS_TS_VECTOR,
-            reverse_sql=migrations.RunSQL.noop,
-        ),
         migrations.RunSQL(
             sql=CREATE_COLLECTIONS_TS_VECTOR_TRIGGER,
             reverse_sql=DROP_COLLECTIONS_TS_VECTOR_TRIGGER,
+        ),
+        migrations.RunSQL(
+            sql=REBUILD_COLLECTIONS_TS_VECTOR,
```
gerrod3, line 90: Does there need to be a noop for the reverse_sql? I think you can't down-migrate if there is no reverse.

jctanner: Probably.

gerrod3 (approved):

PR #1285, opened 2022-11-15
Update CI

gerrod3:
@mdellweg I think this has the correct changes to `post_before_script` to get the CI to pass.


mdellweg (approved):
Looks like it!
Thank you!

PR #1305, opened 2022-12-05
[WIP] Add initial cross-repo search endpoint.

jctanner:
https://issues.redhat.com/browse/AAH-1968
https://github.com/pulp/pulp_ansible/issues/739

Issue #739, opened 2021-11-17
Implement collection version search endpoint


pulpbot:
Implement a new collection version search endpoint that allows for searching collections across multiple repositories at `/v3/plugin/ansible/search/collection-versions/`

### Goal
- Replace the current functionality `_ui/v1/repo/{base_path}` and `_ui/v1/collection-versions` endpoints so that we can sunset the unsupported _ui APIs in galaxy_ng
- Provide an easy way to locate collections in the galaxy_ng UI when users have collections spread across multiple repositories

### Requirements
- Include the data that are in the current serializers for the existing _ui apis. 
  - namespace information
  - collection version metadata
  - signature data
  - distribution that the 
- Include the query params that are in the current _ui apis
  - `sign_state` filter collection versions by signed and unsigned
  - `deprecated`
  - `keywords` keyword search filter
  - `dependency` return a list of collections that depend on a given collection name
- Must be able to return all collection versions, or only the latest version of each collection in each repository
- Searched repositories must be limited to ones that the user has permission to view
- Must be able to filter to a single repository version (via a distribution base path), or a list of repository versions based on a set of pulp repository labels
- Collection versions should only be returned if they are in a repository version with a distribution attached to it
- Queryset for the API must be performant

Author: @newswangerd (newswangerd)

Original Redmine Issue: 9051, https://pulp.plan.io/issues/9051

goosemania:
Whoever works on it, please refer in the commit to the AH jira ticket, so they are linked properly. https://issues.redhat.com/browse/AAH-1968

PR #1316, opened 2023-01-03
[WIP] Implement the newswangerd method for cross repo search.

jctanner:

maybe replaces https://github.com/pulp/pulp_ansible/pull/1305 ?

### DIFF
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -375,11 +375,7 @@ class CollectionVersionSearchListSerializer(CollectionVersionListSerializer):
         model = models.CollectionVersion
 
     def get_collection_version(self, obj):
-        print(f'SERIALIZER OBJ: {obj}')
-        #return obj.content.cast()
         return obj.collection_version
-        #return obj
-        #return obj.content
 
     def get_pulp_id(self, obj):
         return self.get_collection_version(obj).pulp_id
@@ -422,8 +418,7 @@ class CollectionVersionSearchListSerializer(CollectionVersionListSerializer):
         )
 
     def get_is_deprecated(self, obj):
-        #return self.get_collection_version(obj).is_deprecated
-        return False
+        return obj.is_deprecated
 
     def get_tags(self, obj):
         return [x.name for x in self.get_collection_version(obj).tags.all()]
--- a/pulp_ansible/app/galaxy/v3/serializers.py
+++ b/pulp_ansible/app/galaxy/v3/serializers.py
@@ -375,8 +375,11 @@ class CollectionVersionSearchListSerializer(CollectionVersionListSerializer):
         model = models.CollectionVersion
 
     def get_collection_version(self, obj):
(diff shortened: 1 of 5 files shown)


jctanner:
I'm still working on the tests. Please take a look at the RepositoryContent hack and verify it's the path we want to go down.

pulp_ansible/app/galaxy/v3/viewsets.py
```diff
@@ -0,0 +1,69 @@
...
+            include_q = include_q | Q(repository=distro.repository, version_added__number__lte=rv)
+            exclude_q = exclude_q | Q(repository=distro.repository, version_removed__number__lte=rv)
+
+        qs = RepositoryContent.objects.filter(
+                content__pulp_type="ansible.collection_version"
+            ).exclude(
+                exclude_q
+            ).filter(
+                include_q
+            )
```
gerrod3, line 64: This logic makes sense for trying to keep the query count to 1, but have you considered using RepositoryVersion's `get_content` method? https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/repository.py#L654. With this method you can pass in an existing CollectionVersion queryset to avoid having to cast the objects in the serializer. You would have to OR together the queries across each distribution's repository's content, but it technically should only result in one query to the DB as well, however I'm not sure if it would be as performant as your way now.

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -333,3 +333,96 @@ class ClientConfigurationSerializer(serializers.Serializer):
...
+            "repository_name",
+            "tags",
+        )
+        model = models.CollectionVersion
+
+    def get_collection_version(self, obj):
+        return obj.content.cast()
+
+    def get_pulp_id(self, obj):
+        return self.get_collection_version(obj).pulp_id
```
rochacbruno, line 381: Is `self.get_collection_version` cached? or calling it multiple times for each `SerializerMethodField` will trigger a new query?

jctanner: No caching yet. I still have to hack through how the obj is initially given to the serializer and the subsequent method calls.

pulp_ansible/app/galaxy/v3/viewsets.py
gerrod3: My one worry with your current approach is that each `cast` call results in a db query, so as the number of results for a search increases, so does the number of queries required to show the results.

jctanner: Is .content.cast() the only way to access the CollectionVersion attributes?

gerrod3: Yes. It probably makes more sense to convert the RepositoryContent query to a CollectionVersion query to avoid casting for each result.

newswangerd (changes requested):
Can we close this since https://github.com/pulp/pulp_ansible/pull/1337 seems to be the way to go?

jctanner:
replaced by https://github.com/pulp/pulp_ansible/pull/1357

PR #1317, opened 2023-01-05
Repo management improvements - url filtering and last sync time

jerabekjiri:
fixes: #1344 
AAH issue: [AAH-2076](https://issues.redhat.com/browse/AAH-2076)

For repo management, we need little improvements for our galaxy_ng and UI (ansible-hub-ui)
- filtering by `url` in `CollectionRemote`
- last sync status (`finished_at`, `status`, `error`) for `CollectionRemoteSerializer` and `AnsibleRepositorySerializer`

(AnsibleRepository has `last_synced_metadata_time` field, but it's always null, even after sync)

### DIFF
--- a/CHANGES/1344.feature
+++ b/CHANGES/1344.feature
@@ -1,0 +1,2 @@
+Added last_sync_task field to CollectionRemote and AnsibleRepository
+Added filtering by url in CollectionRemote
--- a/CHANGES/1344.feature
+++ b/CHANGES/1344.feature
@@ -1,2 +1,2 @@
-Added last_sync_task field to CollectionRemote and AnsibleRepository
-Added filtering by url in CollectionRemote
+Added ``last_sync_task`` field to ``CollectionRemote`` and ``AnsibleRepository``.
+Added filtering by ``url`` in ``CollectionRemote``.
--- a/pulp_ansible/app/models.py
+++ b/pulp_ansible/app/models.py
@@ -268,6 +268,11 @@ class RoleRemote(Remote):
         default_related_name = "%(app_label)s_%(model_name)s"
 
 
+def _get_last_sync_task(pk):
+    sync_tasks = Task.objects.filter(name__contains="sync", reserved_resources_record__icontains=pk)
+    return sync_tasks.order_by("-pulp_created").first()
+
+
 class CollectionRemote(Remote):
     """
     A Remote for Collection content.
@@ -301,6 +306,10 @@ class CollectionRemote(Remote):
         except AttributeError:
             self._download_factory = AnsibleDownloaderFactory(self)
             return self._download_factory
+
+    @property
+    def last_sync_task(self):
+        return _get_last_sync_task(self.pk)
 
     @hook(
         AFTER_UPDATE,
@@ -365,6 +374,10 @@ class AnsibleRepository(Repository):
 
     last_synced_metadata_time = models.DateTimeField(null=True)
     gpgkey = models.TextField(null=True)
+
+    @property
+    def last_sync_task(self):
+        return _get_last_sync_task(self.pk)
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -19,6 +19,7 @@ from pulpcore.plugin.serializers import (
     DistributionSerializer,
     RepositoryVersionRelatedField,
     validate_unknown_fields,
+    TaskSerializer,
 )
 from rest_framework.exceptions import ValidationError
 
@@ -126,6 +127,7 @@ class AnsibleRepositorySerializer(RepositorySerializer):
     last_synced_metadata_time = serializers.DateTimeField(
         help_text=_("Last synced metadata time."), allow_null=True, required=False
     )
+    last_sync_task = TaskSerializer(read_only=True)
     gpgkey = serializers.CharField(
         help_text="Gpg public key to verify collection signatures against",
         required=False,
@@ -136,6 +138,7 @@ class AnsibleRepositorySerializer(RepositorySerializer):
         fields = RepositorySerializer.Meta.fields + (
             "last_synced_metadata_time",
             "gpgkey",
+            "last_sync_task",
         )
         model = AnsibleRepository
 
@@ -211,6 +214,8 @@ class CollectionRemoteSerializer(RemoteSerializer):
         default=False,
     )
 
+    last_sync_task = TaskSerializer(read_only=True)
+
     def validate(self, data):
         """
         Validate collection remote data.
@@ -262,6 +267,7 @@ class CollectionRemoteSerializer(RemoteSerializer):
             "token",
             "sync_dependencies",
             "signed_only",
+            "last_sync_task",
         )
         model = CollectionRemote
 
--- a/pulp_ansible/app/viewsets.py
+++ b/pulp_ansible/app/viewsets.py
@@ -19,6 +19,7 @@ from pulpcore.plugin.viewsets import (
     DistributionViewSet,
     BaseFilterSet,
     ContentFilter,
+    RemoteFilter,
     ContentViewSet,
     HyperlinkRelatedFilter,
     NamedModelViewSet,
@@ -171,6 +172,12 @@ class CollectionVersionFilter(ContentFilter):
     class Meta:
         model = CollectionVersion
         fields = ["namespace", "name", "version", "q", "is_highest", "tags"]
+
+
+class CollectionRemoteFilter(RemoteFilter):
+    class Meta:
+        model = CollectionRemote
+        fields = {"url": ["exact", "in", "icontains", "contains"], **RemoteFilter.Meta.fields}
 
 
 class CollectionVersionViewSet(UploadGalaxyCollectionMixin, SingleArtifactContentUploadViewSet):
@@ -407,6 +414,7 @@ class CollectionRemoteViewSet(RemoteViewSet):
     endpoint_name = "collection"
     queryset = CollectionRemote.objects.all()
     serializer_class = CollectionRemoteSerializer
+    filterset_class = CollectionRemoteFilter
 
     def async_reserved_resources(self, instance):
         if instance is None:


pulp_ansible/app/viewsets.py
```diff
@@ -173,6 +174,10 @@ class Meta:
         fields = ["namespace", "name", "version", "q", "is_highest", "tags"]
 
 
+class CollectionRemoteFilter(RemoteFilter):
+    url = filters.CharFilter(field_name="url")
```
gerrod3, line 178: I think you need to define a Meta class here. See instructions: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/repository.py#L213-L221

pulp_ansible/app/serializers.py
```diff
@@ -126,16 +127,21 @@ class AnsibleRepositorySerializer(RepositorySerializer):
...
+    last_synced_time = serializers.SerializerMethodField()
     gpgkey = serializers.CharField(
         help_text="Gpg public key to verify collection signatures against",
         required=False,
         allow_null=True,
     )
 
+    def get_last_synced_time(self, obj):
+        return get_last_synced_task_time(obj.pk)
+
```
gerrod3, line 139: I think we should fix the bug of `last_synced_metadata_time` always being `null` rather than adding a new field.

pulp_ansible/app/serializers.py
```diff
@@ -211,6 +217,11 @@ class CollectionRemoteSerializer(RemoteSerializer):
         default=False,
     )
 
+    last_synced_time = serializers.SerializerMethodField()
+
```
gerrod3, line 221: Can you explain why this field needs to be available for a remote? A remote can be used in multiple concurrent syncs at the same time, so not sure this field would provide much help. Maybe something that lists the last sync it was used in?

pulp_ansible/app/tasks/utils.py
```diff
@@ -216,3 +217,15 @@ def get_file_obj_from_tarball(tar, file_path, artifact_path, raise_exc=True):
...
+
+
+def get_last_synced_task_time(pk):
+    task = (
+        Task.objects.filter(reserved_resources_record__icontains=pk, name__icontains="sync")
+        .order_by("-pulp_last_updated")
+        .first()
+    )
+    if task:
+        return task.pulp_last_updated
```
gerrod3, line 229: I would suggest using `pulp_created`. Multiple syncs can use the same remote and an older sync could run longer than a newer sync thus end up being the "lastest" sync when in fact it is just taking longer to complete.

pulp_ansible/app/serializers.py
jerabekjiri: I wasn't sure about this one. I thought it rather could be related to the `rebuil_metadata` action

pulp_ansible/app/serializers.py
jerabekjiri: >Maybe something that lists the last sync it was used in?

I think you're correct, in our current solution, we have the `last_sync_task` which provides more than just the last synced time, but also state and error which we need
```
"last_sync_task": {
        "task_id": "f5e675e7-a429-4616-aea8-d82aae790610",
        "state": "completed",
        "started_at": "2023-01-09T12:05:58.642566Z",
        "finished_at": "2023-01-09T12:06:02.272053Z",
        "error": null
},
```

jerabekjiri:
I took the current logic for the `last_sync_status` from `galaxy_ng` codebase [ansible/galaxy_ng/blob/master/galaxy_ng/app/api/utils.py#L136](https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/utils.py#L136)

To serialize it I wanted to use the `MinimalTaskSerializer` which is exactly what we would want, but it's not exported in the plugin, therefore can't use it. WDYT? @gerrod3

gerrod3:
> To serialize it I wanted to use the `MinimalTaskSerializer` which is exactly what we would want, but it's not exported in the plugin, therefore can't use it. WDYT? @gerrod3

We can expose it in pulpcore and have it available in the next pulpcore release, this would require bumping your minimum pulpcore version. The other option is just copying the serializer for now and replace it later once we added it to the plugin api.

jerabekjiri:
> We can expose it in pulpcore and have it available in the next pulpcore release, this would require bumping your minimum pulpcore version. The other option is just copying the serializer for now and replace it later once we added it to the plugin API.

hmm... I noticed `MinimalTaskSerializer` is missing the `error` field which we use, so would it be possible to add it to the `MinimalTaskSerializer` or just use this custom response: https://github.com/pulp/pulp_ansible/blob/400b0109d6474fb5859b85f7dddb3d6294f42bdf/pulp_ansible/app/tasks/utils.py#L229-L237

mdellweg:
What if instead of the `MinimalTaskSerializer`, you used the `TaskSerializer`?
The content of error can be quite large. This kind of conflicts with "minimal" for me.

newswangerd:
related: https://github.com/pulp/pulpcore/issues/2863

pulp_ansible/app/viewsets.py
mdellweg: And please try without this line. The filter should be generated from the fields attribute.

pulp_ansible/app/viewsets.py
```diff
@@ -173,6 +176,14 @@ class Meta:
...
+
+    class Meta:
+        model = CollectionRemote
+        fields = ["url"]
```
mdellweg, line 184: Can you verify that we keep the filters from the superclass when we specify this?

pulp_ansible/app/viewsets.py
```diff
@@ -293,6 +304,8 @@ def sync(self, request, pk):
         elif isinstance(remote, GitRemote):
             sync_func = git_sync
 
+        set_guid(generate_guid())
```
mdellweg, line 307: Nope...

pulp_ansible/app/viewsets.py
jerabekjiri: hmm... so when I added the Meta class to the `CollectionRemoteFilter`, filters are missing inherited fields:
```
['url', 'pulp_label_select', 'ordering']
```

and without Meta class:
```
['name', 'name__in', 'name__icontains', 'name__contains', 'name__startswith', 'pulp_last_updated', 'pulp_last_updated__lt', 'pulp_last_updated__lte', 'pulp_last_updated__gt', 'pulp_last_updated__gte', 'pulp_last_updated__range', 'pulp_label_select', 'url', 'ordering']
```

jerabekjiri: For some reason with `Meta` class, the inheritance doesn't work and fields from the superclass are missing (only ordering is present).  Fields from `RemoteFilter` come as a dict type, so this feels like a hackish solution:
```
fields = {"url": ["exact", "in", "icontains", "contains"], **RemoteFilter.Meta.fields}
```

mdellweg: That is actually way less hackier than you think, and what i think it should be.

pulp_ansible/app/tasks/utils.py
```diff
@@ -216,3 +218,16 @@ def get_file_obj_from_tarball(tar, file_path, artifact_path, raise_exc=True):
         raise FileNotFoundError(f"{file_path} not found")
 
     return file_obj
+
+
+def get_last_sync_task(pk):
+    task = (
+        Task.objects.filter(reserved_resources_record__icontains=pk, name__icontains="sync")
```
mdellweg, line 225:
```suggestion
        Task.objects.filter(reserved_resources_record__contains=pk, name__contains="sync")
```
We don not need to search case insensitive.

pulp_ansible/app/tasks/utils.py
```diff
@@ -216,3 +218,16 @@ def get_file_obj_from_tarball(tar, file_path, artifact_path, raise_exc=True):
...
+        .order_by("-pulp_created")
+        .first()
+    )
+
+    if task:
+        return TaskSerializer(task, context={"request": None}).data
```
mdellweg, line 231: I'm wondering if we can make this function a property on the model called `last_sync_task` and then define the serializer field like `last_sync_task = TaskSerializer`.

https://www.django-rest-framework.org/api-guide/serializers/#dealing-with-nested-objects

pulp_ansible/app/models.py
```diff
@@ -365,6 +365,12 @@ class AnsibleRepository(Repository):
 
     last_synced_metadata_time = models.DateTimeField(null=True)
     gpgkey = models.TextField(null=True)
+    last_sync_task = models.ForeignKey(
+        Task,
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True,
+    )
```
gerrod3, line 373:
```suggestion
    last_sync_task = models.ForeignKey(
        Task,
        on_delete=models.CASCADE,
        null=True,
    )
```
`null=True` should be enough for a ForeignKey

pulp_ansible/app/tasks/collections.py
```diff
@@ -179,6 +179,17 @@ def sync(remote_pk, repository_pk, mirror, optimize):
     d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
     repo_version = d_version.create()
 
+    last_sync_task = (
+        Task.objects.filter(
+            reserved_resources_record__contains=repository.pk, name__contains="sync"
+        )
+        .order_by("-pulp_created")
+        .first()
+    )
```
gerrod3, line 188:
```suggestion
    last_sync_task = Task.current()
```
We have a special method for finding the task object of the current executing task!

gerrod3 (changes requested):

pulp_ansible/app/tasks/collections.py
jerabekjiri: okay, that's cool :+1:

pulp_ansible/app/migrations/0047_ansiblerepository_last_sync_task.py
```diff
@@ -0,0 +1,20 @@
...
+        ('core', '0099_versions_field'),
+        ('ansible', '0046_add_fulltext_search_fix'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='ansiblerepository',
+            name='last_sync_task',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='ansible_ansiblerepository', to='core.task'),
+        ),
```
gerrod3, line 19: You are going to need to change the migration file as well to remove `blank`.

jerabekjiri: oops, forget about that, thanks :)

mdellweg (changes requested):

pulp_ansible/app/models.py
```diff
@@ -268,6 +268,15 @@ class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
 
 
+def _get_last_sync_task(pk):
+    sync_task = Task.objects.filter(name__contains="sync", reserved_resources_record__icontains=pk)
+
+    if sync_task:
+        return sync_task.order_by("-pulp_created").first()
+
+    return None
```
mdellweg, line 277:
```suggestion
def _get_last_sync_task(pk):
    sync_tasks = Task.objects.filter(name__contains="sync", reserved_resources_record__icontains=pk)
    return sync_tasks.order_by("-pulp_created").first()
```

`first` has this logic builtin.

mdellweg (changes requested):

pulp_ansible/app/models.py
jerabekjiri: thanks, I thought this was throwing an error, but it was actually using `reserved_resources_record__contains` instead of  `reserved_resources_record__icontains`

```
pulp_1   | pulp [4f173387cd7a494c84129a1f7cde3c17]: django.request:ERROR: Internal Server Error: /api/automation-hub/pulp/api/v3/repositories/ansible/ansible/fddb35fa-b626-401b-a3a3-7f93d2d24dce/
pulp_1   | Traceback (most recent call last):
pulp_1   |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
pulp_1   |     return self.cursor.execute(sql, params)
pulp_1   | psycopg2.errors.CannotCoerce: cannot cast type uuid to text[]
pulp_1   | LINE 1: ...d" @> 'fddb35fa-b626-401b-a3a3-7f93d2d24dce'::uuid::text[]) ...
pulp_1   |                                                              ^
```

mdellweg (approved):

gerrod3 (approved):

mdellweg (changes requested):
Sorry, it's me now. This is missing an issue and changelogs. Sorry for not seeing that earlier.
And heads up we require this for _all_ features and bugfixes.

Also, BTW the Squash merge thing is really not working for me. It completely circumvents our commit message check. cc @bmbouter

CHANGES/1344.feature
```diff
@@ -0,0 +1,2 @@
+Added last_sync_task field to CollectionRemote and AnsibleRepository
+Added filtering by url in CollectionRemote
```
mdellweg, line 2:
```suggestion
Added ``last_sync_task`` field to ``CollectionRemote`` and ``AnsibleRepository``.
Added filtering by ``url`` in ``CollectionRemote``.
```

mdellweg (changes requested):

mdellweg (approved):

Issue #1323, opened 2023-01-16
`logging_cid` is null on dispatching Sync task 


jerabekjiri:
**Version**
pulpcore: `3.22.0`
pulp_ansible: `0.17.0`
galaxy_ng: `4.7.0`

**Describe the bug**
Sync task fails with status code 500

```
pulp_1   | pulp [None]: django.request:ERROR: Internal Server Error: /api/automation-hub/pulp/api/v3/repositories/ansible/ansible/3dc6a18f-d1b1-44dd-a6ed-0b00198199b2/sync/
pulp_1   | Traceback (most recent call last):
pulp_1   |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
pulp_1   |     return self.cursor.execute(sql, params)
pulp_1   | psycopg2.errors.NotNullViolation: null value in column "logging_cid" of relation "core_task" violates not-null constraint
pulp_1   | DETAIL:  Failing row contains (f9d4f0b4-16f0-47ff-b55b-db437b3c41cf, 2023-01-16 15:38:50.537753+00, 2023-01-16 15:38:50.537775+00, waiting, pulp_ansible.app.tasks.collections.sync, null, null, null, null, null, null, null, null, {"mirror": false, "optimize": true, "remote_pk": "5421af5a-5d70-..., {/api/automation-hub/pulp/api/v3/repositories/ansible/ansible/3d...).
pulp_1   |
pulp_1   |
pulp_1   | The above exception was the direct cause of the following exception:
pulp_1   |
pulp_1   | Traceback (most recent call last):
pulp_1   |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
pulp_1   |     response = get_response(request)
pulp_1   |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
pulp_1   |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
pulp_1   |   File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
pulp_1   |     return view_func(*args, **kwargs)
pulp_1   |   File "/usr/local/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
pulp_1   |     return self.dispatch(request, *args, **kwargs)
[...]

PR #1324, opened 2023-01-16
Set guid for `logging_cid`

jerabekjiri:
fixes: #1323

### DIFF
--- a/CHANGES/1323.bugfix
+++ b/CHANGES/1323.bugfix
@@ -1,0 +1,1 @@
+Generate and set ``logging_cid`` value for sync task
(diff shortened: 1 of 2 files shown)


mdellweg:
This is not how it's done. The guid must be set by the middleware. Could this be a misconfiguration of `django-guid`?

jerabekjiri:
> This is not how it's done. The guid must be set by the middleware. Could this be a misconfiguration of `django-guid`?

You are right :+1:, we need to add `django_guid.middleware.guid_middleware` to `MIDDLEWARE` in our Galaxy_ng settings. Thanks a lot :)

Issue #1325, opened 2023-01-17
Add CollectionVersionMark as a new content type


rochacbruno:
Relates to https://issues.redhat.com/browse/AAH-2043

**Describe the solution you'd like**

Adds a new model `CollectionVersionMark` that can be added to a repository and relates to a `CollectionVersion`

![collectionversionmark](https://user-images.githubusercontent.com/458654/212935301-3236addd-c2cf-4f6c-9c97-90d11b94ebe7.png)

**this image can be imported to excalidraw**

PR #1326, opened 2023-01-17
[feat] Add CollectionVersionMark as a content.

rochacbruno:
fixes: #1325

```python
In [83]: repo1 = AnsibleRepository.objects.first()

# A repo can have collection versions added
In [84]: repo1.content.filter(pulp_type="ansible.collection_version").first()
Out[84]: <Content (pulp_type=ansible.collection_version): collection_dep_a_fapdgczi>

# A repo also can have CollectionVersionMark
In [85]: mark = repo1.content.filter(pulp_type="ansible.collection_mark").first()
In [86]: mark.ansible_collectionversionmark
Out[86]: <CollectionVersionMark: pk=abac7714-b3a6-4bfc-8d98-36f75f5f2e1d>

# The mark has a foreign key to a CollectionVersion
In [87]: mark.ansible_collectionversionmark.marked_collection
Out[87]: <CollectionVersion: orionuser1.collection_dep_a_fapdgczi 1.0.0>

# The mark has a string value
In [88]: mark.ansible_collectionversionmark.value
Out[88]: 'validated'

# The common pulp metadata
In [89]: mark.ansible_collectionversionmark.pulp_created
Out[89]: datetime.datetime(2023, 1, 17, 15, 57, 34, 816490, tzinfo=<UTC>)

# And can be used for filtering/searching
In [90]: CollectionVersion.objects.filter(marks__value="validated").first()
Out[90]: <CollectionVersion: orionuser1.collection_dep_a_fapdgczi 1.0.0>

# NOTE: A new RepositoryVersion is created every time a mark is added to it.
```

URLS added

```
$ oci-env exec pulpcore-manager show_urls | grep mark
/pulp/api/v3/content/ansible/collection_marks/	pulp_ansible.app.viewsets.CollectionVersionMarkViewSet	content-ansible/collection_marks-list
/pulp/api/v3/content/ansible/collection_marks/<pk>/	pulp_ansible.app.viewsets.CollectionVersionMarkViewSet	content-ansible/collection_marks-detail
/pulp/api/v3/content/ansible/collection_marks/<pk>\.<format>/	pulp_ansible.app.viewsets.CollectionVersionMarkViewSet	content-ansible/collection_marks-detail
/pulp/api/v3/content/ansible/collection_marks\.<format>/	pulp_ansible.app.viewsets.CollectionVersionMarkViewSet	content-ansible/collection_marks-list

```


awcrosby:
@rochacbruno in your diagram will each mark for CV `999` exist in both repo `001` and `002` since marks are moved with a CV, and that CV exists in both repos?

mdellweg:
> @rochacbruno in your diagram will each mark for CV `999` exist in both repo `001` and `002` since marks are moved with a CV, and that CV exists in both repos?

The marks are a new type of content, they will only be present in the RV that explicitly includes them. Whether marks are updated on sync, or copied with the CV on a copy task is dependent on the semantic of that actual task. Only the removal of a CV from a repository will bubble into removing the attached marks automatically.

pulp_ansible/app/modelresource.py
```diff
@@ -75,6 +76,22 @@ class Meta:
         exclude = BaseContentResource.Meta.exclude + ("signing_service",)
 
 
+class CollectionVersionMarkResource(BaseContentResource):
```
mdellweg, line 79: Not sure if you _need_ to, but i would add it to the import order list at the bottom.

pulp_ansible/app/tasks/collections.py
```diff
@@ -640,7 +642,7 @@ async def _add_collection_version(self, api_version, collection_version_url, met
         await self.parsing_metadata_progress_bar.aincrement()
         await self.put(d_content)
 
-        if signatures:
+        if signatures or marks:
             collection_version = await d_content.resolution()
```
mdellweg, line 646: ohh, (not really a comment to this pr) this can significantly slow down sync, as this collection version basically jumps the whole pipeline without waiting to be processed in batches. We should refactor this soon.

pulp_ansible/app/modelresource.py
rochacbruno: done

pulp_ansible/app/tasks/collections.py
rochacbruno: lets tag @gerrod3 for a discussion on this topic on our next sync meeting

pulp_ansible/app/modelresource.py
mdellweg: Thank you. Actually, after thinking about it, we need to import the marks after the collection versions. So yes, we need to. ;)

pulp_ansible/app/models.py
```diff
@@ -467,8 +490,8 @@ def finalize_new_version(self, new_version):
             base_version=new_version.base_version
         ).filter(pulp_type=CollectionVersion.get_pulp_type())
 
-        # Remove any deprecated, signature and namespace content associated with the removed
-        # collection versions
+        # Remove any deprecated, signature, mark and namespace content
+        # associated with the removed collection versions
         for version in removed_collection_versions:
             version = version.cast()
```
mdellweg, line 496: Hrm. All we care for in the c_version is the pk. Right? I don't think we need to `cast` for that. And at that point we don't even need the loop anymore, right?
Would the following work?
```python
        signatures = new_version.get_content(
            content_qs=CollectionVersionSignature.objects.filter(signed_collection__in=removed_collection_versions)
        )
        new_version.remove_content(signatures)
```

mdellweg: Also do we maybe need to call `remove_duplicates` first? It may remove more CVs.

pulp_ansible/app/tasks/collections.py
mdellweg: Please add some meaningful sync tests. We want to be sure we do not regress when we refactor this.

pulp_ansible/tests/functional/conftest.py
```diff
@@ -62,6 +63,12 @@ def ansible_collection_signatures_client(ansible_bindings_client):
     return ContentCollectionSignaturesApi(ansible_bindings_client)
 
 
+@pytest.fixture
+def ansible_collection_mark_client(ansible_bindings_client):
```
mdellweg, line 67: :+1:

pulp_ansible/app/models.py
rochacbruno: I prefer to handle this in another PR, my changes are not touching this logic, I am just following the same pattern for the marks, what if we open a new issue based on this comment and then fix in another pr?

pulp_ansible/app/viewsets.py
```diff
@@ -230,6 +234,36 @@ class CollectionVersionSignatureViewSet(NoArtifactContentUploadViewSet):
     serializer_class = CollectionVersionSignatureSerializer
 
 
+class MarkFilter(ContentFilter):
```
mdellweg, line 237:
```suggestion
class CollectionVersionMarkFilter(ContentFilter):
```

pulp_ansible/tests/functional/api/collection/test_marks.py
```diff
@@ -0,0 +1,118 @@
...
+    3. Add a mark to the collection version
+    4. Assert the mark was added correctly
+    5. Remove the mark
+    6. Assert the mark was removed correctly
+    """
+    collection, collection_url = build_and_upload_collection()
+
+    # add it to a repo version
+    body = {"add_content_units": [collection_url]}
+    monitor_task(ansible_repo_api_client.modify(ansible_repo.pulp_href, body).task)
```
mdellweg, line 35: monitor_task is available as a fixture now.

Anyway, Yay for writing pytest test!

pulp_ansible/app/modelresource.py
```diff
@@ -132,6 +149,7 @@ class Meta:
     CollectionDeprecationResource,
     TagResource,
     CollectionVersionContentResource,
+    CollectionVersionMarkResource,
```
mdellweg, line 152: Do you think, it would be possible to inject a mark or two into the import-export tests?

mdellweg (commented):
Great work, almost done. (Just the suggested change is really mandatory now.)

mdellweg:
You need to redo / rebase your migration.

pulp_ansible/app/migrations/0048_merge_20230223_1246.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.2.18 on 2023-02-23 12:46
```
mdellweg, line 1: I prefer to keep the migration history linear. Sorry.
Renaming the file and switching the dependency lines works usually. But you also added no manual part, so just deleting and recreating may be easier.

mdellweg (approved):

Issue #1219, opened 2022-08-31
Implement "total" on sync progress report


newswangerd:
**Is your feature request related to a problem? Please describe.**
The progress report on the sync task has a "total" field that's always null. It would be nice if pulp_ansible could display this value so the user knows how much progress has been made.

**Describe the solution you'd like**
```
    "progress_reports": [
        {
            "message": "no_change: Checking if remote changed since last sync.",
            "code": "sync.no_change",
            "state": "completed",
            "total": null,                                 <------- Put the total count here
            "done": 0,
            "suffix": null
        },
        {
            "message": "Downloading Artifacts",
            "code": "sync.downloading.artifacts",
            "state": "running",
            "total": null,                                 <------- Put the total count here
            "done": 0,
            "suffix": null
        },
        {
            "message": "Associating Content",
            "code": "associating.content",
            "state": "running",
            "total": null,                                 <------- Put the total count here
            "done": 0,
            "suffix": null
        },
        {
            "message": "Parsing CollectionVersion Metadata",
            "code": "sync.parsing.metadata",
            "state": "running",
            "total": null,                                 <------- Put the total count here
            "done": 868,
            "suffix": null
        },
        {
            "message": "Downloading Artifacts",
            "code": "sync.downloading.artifacts",
            "state": "running",
            "total": null,                                 <------- Put the total count here
            "done": 267,
            "suffix": null
        }
    ],
```

newswangerd:
Also, dumb question, why are there two `Downloading Artifacts` fields with different values?

mdellweg:
I don't know much about the pulp_ansible sync pipeline details... So this is a general Pulp answer:
The sync pipeline is organized in stages that run in parallel, so downloading starts before all the metadata is digested. So it's generally hard to know the total numbers upfront. OTOH having a running total instead is more confusing. Maybe it's possible to inject the total number once the first stage is finished. But due to backpressure, this may be very late in the process anyway.
Are there two artifact download stages?

gerrod3:
The two `Downloading Artifacts` is interesting, not sure what is going on there. 

For v3 galaxy syncs if they are using the unpaginated endpoint we know the total number before filtering or dependencies are added. So depending on the sync settings used this number could be way off from the actual number synced. We can always update this total as we go along through the sync if you want to use it as a progress bar.

newswangerd:
[...]

PR #1335, opened 2023-01-23
Add sync total to collection sync first stage

gerrod3:
fixes: #1219

@newswangerd This is the best I can do for updating totals in the sync progress report. The other reports are apart of pulpcore and are designed without foreknowledge of the final total.

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -711,6 +711,8 @@ class CollectionSyncFirstStage(Stage):
                 break
             page_num = page_num + 1
 
+        self.parsing_metadata_progress_bar.total += len(tasks)
+        await self.parsing_metadata_progress_bar.asave(update_fields=["total"])
         await asyncio.gather(*tasks)
 
     async def _read_from_downloaded_metadata(self, name, namespace, requirement):
@@ -756,6 +758,8 @@ class CollectionSyncFirstStage(Stage):
                             )
                         )
                     )
+        self.parsing_metadata_progress_bar.total += len(tasks)
+        await self.parsing_metadata_progress_bar.asave(update_fields=["total"])
         await asyncio.gather(*tasks)
 
     async def _fetch_collection_metadata(self, requirements_entry):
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -969,6 +977,8 @@ async def run(self):
             else:
                 tasks.append(loop.create_task(self._find_all_collections()))
             await asyncio.gather(*tasks)
+            # Ensure PR 'total' is correct before stage finishes
+            pb.total = pb.done
```
mdellweg, line 981: Should this be an assertion? What can go wrong here without being fatal?

gerrod3: An assertion will cause the sync to fail pretty often with how I currently have it set up. For non-full syncs, the total typically overshoots the done number by a small amount. This is due to the fact that each item I add to the total (num of tasks spawned to process a collection version) does not equal the number of collection versions that gets put into the pipeline due to filtering.

mdellweg: Hmm this makes me uneasy as we are not only having a running total, but also sounds like counting a different thing. Should we count these with separate progress bars instead?

mdellweg: I'd guess in weird situations with aggressive filtering we'd end up with very different figures.

gerrod3: We are estimating how many collection versions we are possibly adding to the repository, not sure what need there is for a second progress bar.

gerrod3: We could but I see no point, and doing so gets away from the crux of the issue which is to implement a total count.

newswangerd: I think this is the right thing to do. It's okay if the estimates are a little wonky while the task is running (progress bars fudge numbers all the time), but the total should match the completed at the end.

mdellweg: So you two agree. I'll rest my case then.

mdellweg (approved):

PR #1336, opened 2023-01-23
Allow GitRemotes on AnsibleRepositories

gerrod3:
fixes: #1140


mdellweg (approved):
Looks obvious.
Are the remote types not checked on sync? I think we should do that.

gerrod3:
> Looks obvious. Are the remote types not checked on sync? I think we should do that.

No, we have an issue for it in pulpcore: https://github.com/pulp/pulpcore/issues/1860

PR #1337, opened 2023-01-24
x-repo search with a psql view

jctanner:

maybe replaces https://github.com/pulp/pulp_ansible/pull/1366 ?

Basic explanation: https://gist.github.com/jctanner/c3a77883aa957de9a0e50be7fc4a3811

I ran a benchmark with increasing levels of collectionsversions and repositories ...

image

The response time (duration) for the django-view scales linearly. It's ~.1 seconds for smaller amounts of collections and hovers around .5 seconds for larger amounts (up to 10K in this test).  

The filterset has a small performance impact when  filtering by signed or unsigned, but the duration is still reasonable.

Here is a correlation table for what affects the duration ...

```
version                    -0.193676
namespace                  -0.192813
name                       -0.191717
params                      0.086050
repository_count            0.304023
signed                      0.306051
collection_version_count    0.451162
duration                    1.000000
```


pulp_ansible/app/models.py
```diff
@@ -417,3 +417,44 @@ class AnsibleDistribution(Distribution):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+
+
+class RepositoryCollectionVersionIndex(models.Model):
```
rochacbruno, line 422: I think it would be helpful if the model name says explicitly that this is a `View` and not a table.

Because then we know that we can't call `objects.create` for example and other `CUD` methods.

Maybe `RepositoryCollectionVersionView`

pulp_ansible/app/models.py
```diff
@@ -417,3 +417,44 @@ class AnsibleDistribution(Distribution):
...
+
+    # the number of deprecations found for this namespace.name
+    # if >0, then it is "deprecated"
+    deprecation_count = models.IntegerField()
+
+    # the count of signatures that are attached to the collectionversion
+    # which are also in the repository as content
+    sig_count = models.IntegerField()
+
+    class Meta:
```
rochacbruno, line 458: I guess if you add `abstract = True` Django will stop generating migrations and CI will pass

rochacbruno: I was reading that Django would generate a migration even for unmanaged models (not sure why), but I think it has something to do with foreign keys.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,120 @@
...
+#     repositories containing the CV.
+#   * Many models and objects and concepts desired for cross repo search
+#     filtering are not directly related to repositories.
+#   * Using the queryset language to build the counts for signatures and
+#     deprecations were drifting from wizardry into impossibility.
+#   * Creating a real index model requires too many hooks in various
+#     obscure workflows for CRUD and new devs will always make mistakes.
+
+CREATE_REPOSITORY_COLLECTIONVERSION_VIEW = '''
+CREATE OR REPLACE VIEW repository_collection_version AS
```
rochacbruno, line 24: Could't this be a materialized view?

Using `VIEW` will make all the columns to be computed on every query.
Using `MATERIALIZED VIEW` will create a physical temporary table to store the results of the view and will compute it every time any indexed table is updated, making it faster.

I am not sure if there are any caveats on using it with pulp.

pulp_ansible/app/models.py
rochacbruno: Nevermind... I just tried and this does the job but looks like django won't populate the manager i.e. objects making it non queryable. 

ACtually you need to run makemigrations and commit the generated migration for the view, it is a **dummy** migration but is required.

jctanner: Got it, ty!

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: When I read the documentation at https://www.postgresql.org/docs/current/rules-materializedviews.html my iterpretation was that the only way to update* data in the materialized view is to call `REFRESH MATERIALIZED VIEW` periodically or anytime a table changes. I was explicitly trying to avoid writing hooks throughout the code that could catch all possible CRUD actions to then update the view.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,120 @@
...
+		SELECT
+			COUNT(*)
+		FROM
+			core_repositorycontent crc2
+		WHERE
+			crc2.content_id=cc.pulp_id
+			AND
+			crc2.repository_id=cr.pulp_id
+			AND acv.content_ptr_id=crc2.content_id
+	)>=1
```
gerrod3, line 94: If I am understanding this correctly we are selecting from repositories that have ansible collection versions in it? Is there any difference selecting from ansible repositories since they are the only repositories that can hold collection versions? 

Also, is it necessary to select from repos that have content in it or can this be dropped if we know we are pulling from ansible repos?

jctanner: Created a new view so this logic has all changed.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,144 @@
...
+                crc3.content_id=acvs.content_ptr_id
+        )>=1
+    ) as sig_count
+
+FROM
+    core_repositorycontent crc
+INNER JOIN
+    core_repository cr ON cr.pulp_id=crc.repository_id
+INNER JOIN
+    core_repositoryversion crv ON crv.repository_id=cr.pulp_id
```
gerrod3, line 112: How does this INNER JOIN work? Again normal pulp with full repo history can have multiple repo versions pointing to a repository, so I don't see how this join would work in that case, but maybe I'm missing something.

gerrod3 (commented):
Questions about the main FROM logic of the query, having a hard time wrapping my head around what is going on. I feel it shouldn't work the way it is intended.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,144 @@
...
+                core_repositorycontent crc3
+            WHERE
+                crc3.repository_id=crc.repository_id
+            AND
+                crc3.content_id=acvs.content_ptr_id
+        )>=1
+    ) as sig_count
+
+FROM
+    core_repositorycontent crc
```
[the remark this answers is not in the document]
jctanner: The query only includes rows from the core_repositorycontent table (crc) that have a matching content_id in the ansible_collectionversion table (acv) through the join on "acv.content_ptr_id=crc.content_id". Therefore, rows in the core_repositorycontent table that do not have a matching content_id in the ansible_collectionversion table will not be included in the final results of the query.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: The inner join on core_repositoryversion (crv) is joining the core_repositorycontent (crc) table with the core_repositoryversion table on the repository_id. This means that for each row in the core_repositorycontent table, the query will only include the corresponding rows from the core_repositoryversion table that have the same repository_id. This join operation allows the query to retrieve the version number of the repository which the collection version is in.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,144 @@
...
+INNER JOIN
+    core_repository cr ON cr.pulp_id=crc.repository_id
+INNER JOIN
+    core_repositoryversion crv ON crv.repository_id=cr.pulp_id
+INNER JOIN
+    ansible_collectionversion acv ON acv.content_ptr_id=crc.content_id
+INNER JOIN
+    HighestVersionLookup hvl ON hvl.highest_version_repository_id=crv.repository_id
+WHERE
```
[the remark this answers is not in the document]
jctanner: This is taken care of by the inner join on ansible_collectionversion.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,144 @@
...
+    highest_version=crv.number
```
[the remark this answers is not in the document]
jctanner: Yep, I need to do more testing with removed versions and make some assertions that removed CVs don't show up in the list (at least not from the repository where they were removed).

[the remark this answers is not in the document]
jctanner: Fixed.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: @gerrod3 maybe this will help break it down https://gist.github.com/jctanner/c3a77883aa957de9a0e50be7fc4a3811

jctanner:
@gerrod3 @newswangerd latest commit splits the view into 2 views, and properly accounts for the logic of rows in the repositorycontent table with version_added_id and version_removed_id.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,155 @@
...
+
+    cr.pulp_id as repository_id,
+    cr.name,
+    acvs.content_ptr_id,
+    acvs.signed_collection_id as collectionversion_id,
+
+    MAX(crv1.number) as sig_vadded,
+    coalesce(MAX(crv2.number), -1) as sig_vremoved,
+
+    True as is_signed
```
gerrod3, line 35: What is this for?

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,155 @@
...
+    core_repositoryversion crv2 on crv2.pulp_id=crc.version_removed_id
+left join
+    ansible_collectionversion acv on acv.content_ptr_id=acvs.content_ptr_id
+WHERE
+    acvs.content_ptr_id is not null
+GROUP BY
+    cr.pulp_id, acvs.content_ptr_id
+HAVING
+    MAX(crv1.number) > coalesce(MAX(crv2.number), -1)
+;
```
gerrod3, line 55: This is neat logic, should you replace this with the defined names above `sig_vadded` and `sig_vremoved`?

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,155 @@
...
+                SELECT
+                    COUNT(*)
+                FROM
+                    core_repositorycontent crc2
+                WHERE
+                    crc2.repository_id=cr.pulp_id
+                AND
+                    crc2.content_id=acvs.content_ptr_id
+            )>=1
+    ) as sigs,
```
gerrod3, line 105: You are currently not using sigs on the model, should it be removed? Also, I think the logic here doesn't account for the signature being removed.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,155 @@
...
+    coalesce((
+        SELECT
+            rcvs.is_signed
+        FROM
+            repository_collectionsignature rcvs
+        WHERE
+            cr.pulp_id=rcvs.repository_id
+            AND
+            acv.content_ptr_id=rcvs.collectionversion_id
+    ), False) as is_signed
```
gerrod3, line 116: I thought we were trying to avoid subqueries. Can this not have been done with a lookup table like before? Or is this faster?

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
```diff
@@ -0,0 +1,155 @@
...
+    (
+        SELECT
+            COUNT(*)
+        FROM
+            ansible_ansiblecollectiondeprecated acd
+        WHERE
+            acd.namespace=acv.namespace
+        AND
+            acd.name=acv.name
+    ) as deprecation_count,
```
gerrod3, line 84: This is wrong, deprecated is also a content type so it needs the exact same versioning logic as collection-versions and signatures. Although it would be weird if a user were to un-deprecate a collection, technically it can happen.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: fixed.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: I believe that it does work at the sql level, but for whatever reason django doesn't like it ...

```
django.db.utils.ProgrammingError: column "vadded" does not exist
```

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: I believe it is allowed in regular sql, but for whatever reason django doesn't like it ....

```
django.db.utils.ProgrammingError: column "sig_vadded" does not exist
```

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: That's just what I could get working for now ... I'll have to play around with the join and see if I can munge it in. I think it might be possible by throwing a CASE statement into the SELECT stanza which can default to False

jctanner:
This latest commit makes deprecations correct, but it has introduced slowness ...

image

I will try to convert the subqueries into joins and see if that makes a difference.

jctanner:
Subqueries converted to joins and already seeing the 28+ second queries go down to less than a second for 10K+ CVs.

Benchmark results pending.

pulp_ansible/app/migrations/0047_add_repo_cv_index.py
jctanner: done.

jctanner:
We are deferring filtering on is_latest to a future PR for the sake of time and getting this one out the door.

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def filter_by_namespace(self, qs, name, value):
+        return qs.filter(Q(collectionversion__namespace=value))
+
+    def filter_by_version(self, qs, name, value):
+        return qs.filter(Q(collectionversion__version=value))
+
+    def filter_by_dependency(self, qs, name, value):
+        """Return a list of collections that depend on a given collection name."""
+        qs = qs.filter(collectionversion__dependencies__has_key=value)
```
gerrod3, line 58: These are also available on the model, I don't think we need to look through the collection version model.

gerrod3 (changes requested):
Lot's of minor changes I would like to see. Haven't reviewed the tests yet, going to put that in a second review. The SQL queries are looking good, the performance is impressive too!

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,112 @@
...
+from django.db.models import Q
+
+
+class CollectionVersionSearchFilter(FilterSet):
+    """A custom filterset for cross-repo search."""
+
+    name = filters.CharFilter(method="filter_by_name")
+    namespace = filters.CharFilter(method="filter_by_namespace")
+    version = filters.CharFilter(method="filter_by_version")
+    distribution = filters.CharFilter(method="filter_by_repository_name")
```
newswangerd, line 15: This should be distribution_base_path. It should look up the distribution, find the associated repository and return it's content.

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,112 @@
+from django.contrib.postgres.search import SearchQuery
+from django.db.models import fields as db_fields
+from django.db.models.expressions import F, Func
+from django_filters import filters
+from django_filters import FilterSet
+from django.db.models import Q
+
+
+class CollectionVersionSearchFilter(FilterSet):
```
newswangerd, line 9: need to be able to filter by repository labels too. You can use this filterset: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/custom_filters.py#L325. Might need to add pulp labels in the postgres view

pulp_ansible/app/galaxy/v3/viewsets.py
```diff
@@ -0,0 +1,32 @@
...
+    max_page_size = 1000
+
+
+class CollectionVersionSearchViewSet(viewsets.ModelViewSet):
+    """
+    A viewset for cross-repo searches.
+    """
+
+    serializer_class = CollectionVersionSearchListSerializer
+    pagination_class = CollectionVersionSearchViewSetPagination
```
newswangerd, line 27: This needs to use the same pagination class as the rest of the api endpoints (offset/limit)

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -333,3 +333,73 @@ class ClientConfigurationSerializer(serializers.Serializer):
...
+        """Shortcut in case the collectionversion property moves around."""
+        return obj.collectionversion
+
+    def get_is_highest(self, obj):
+        return self.get_collection_version(obj).is_highest
+
+    def get_href(self, obj) -> str:
+        """Get href."""
+
+        ctx = _get_distro_context({"path": obj.reponame, "distro_base_path": obj.reponame})
```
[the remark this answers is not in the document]
newswangerd: We do... This needs to use a distribution instead of a distro name.

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -333,3 +333,73 @@ class ClientConfigurationSerializer(serializers.Serializer):
...
+            "is_signed",
+            # "tags",
+            "signatures",
+        )
+
+    def get_collection_version(self, obj):
+        """Shortcut in case the collectionversion property moves around."""
+        return obj.collectionversion
+
+    def get_is_highest(self, obj):
```
newswangerd, line 386: Can you remove get_is_highest and get_collection_version for now? It's slowing the API endpoint way down and isn't currently implemented yet.

This change speeds the response up by about 2 seconds

pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -333,3 +333,73 @@ class ClientConfigurationSerializer(serializers.Serializer):
...
+            "version",
+            "created_at",
+            "updated_at",
+            "requires_ansible",
+            "dependencies",
+            "href",
+            "is_highest",
+            "is_deprecated",
+            "is_signed",
+            # "tags",
```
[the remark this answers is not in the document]
newswangerd: yes. Anything that calls `self.get_collection_version` slows it way down.

newswangerd (changes requested):

jctanner:
closing in favor of https://github.com/pulp/pulp_ansible/pull/1357

PR #1338, opened 2023-01-30
Implement RBAC

bmclaughlin:
Added Role Based Access Control.

New default roles (creator, owner, viewer) have been added for ``AnsibleRepository``, ``AnsibleDistribution``,
``CollectionRemote``, ``RoleRemote``, and ``GitRemote``.

New detail role management endpoints (``my_permissions``, ``list_roles``, ``add_role``,
``remove_role``) have been added to ``AnsibleRepository``, ``AnsibleDistribution``, ``CollectionRemote``,
``GitRemote``, and ``RoleRemote``.


pulp_ansible/app/models.py
```diff
@@ -198,6 +209,9 @@ class Meta:
                 condition=Q(is_highest=True),
             )
         ]
+        permissions = [
+            ("manage_roles_collectionversion", "Can manage roles on collection versions")
+        ]
```
mdellweg, line 214: We do not manage permissions on the content level, but on the repository level.
One reason is that content is shared between repositories and so multiple users can own the same content implicitly. Another (soft) reason is that Pulp expects to have _a lot_ of content, and this would blow up uncontrolled.

pulp_ansible/app/viewsets.py
```diff
@@ -260,6 +596,55 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+            {
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": "create",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:ansible.add_ansiblerepository",
```
gerrod3, line 612: This condition should be `has_model_perms`. When I tried this out with pulp_file I got a 500 error when trying to create a repo. Also, when testing locally you have to run `pulpcore-manager migrate` to ensure any updates to the access policy get updated in the db, we have a hook that checks for updates.

pulp_ansible/app/models.py
```diff
@@ -45,6 +46,13 @@ def relative_path(self):
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
         unique_together = ("version", "name", "namespace")
+        permissions = [
```
newswangerd, line 49: Can we use the default django permissions for some of these? Django comes with create, update, view and delete by default

pulp_ansible/app/models.py
```diff
@@ -59,6 +67,9 @@ def __str__(self):
 
     class Meta:
         unique_together = ("namespace", "name")
+        permissions = [
+            ("manage_roles_collection", "Can manage roles on Collections"),
```
newswangerd, line 71: I don't know that we need to manage roles on a collection level. This will typically be handled at the namespace or repository level.

@mdellweg thoughts?

pulp_ansible/app/models.py
```diff
@@ -417,3 +436,8 @@ class AnsibleDistribution(Distribution):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+
+        permissions = [
+            ("manage_roles_ansibledistribution", "Can manage roles on ansible distributions"),
+            ("modify_ansible_repo_content", "Can modify ansible repository content"),
```
newswangerd, line 442: I would just defer to the repository permissions for managing the content of the repository.

pulp_ansible/app/models.py
```diff
@@ -345,7 +361,7 @@ class Meta:
         unique_together = ("namespace", "name")
 
 
-class AnsibleRepository(Repository):
+class AnsibleRepository(Repository, AutoAddObjPermsMixin):
```
newswangerd, line 364: We also need to add a flag for private repositories. I know we have a separate issue for that, so if you want to tackle it in a different PR, that's totally fine

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -68,6 +69,61 @@
...
+            "principal": "authenticated",
+            "effect": "allow",
+            "condition": [
+                "has_model_perms:ansible.delete_collection",
+                "has_model_or_obj_perms:ansible.modify_ansible_repo_content",
+            ],
+        },
+        {
+            "action": ["download"],
+            "principal": "authenticated",
```
newswangerd, line 90: @mdellweg should we allow unauthenticated downloading by default since that's how pulp ansible has worked until now? It doesn't matter to me if we change this now, since this is the default for galaxy, but this is a breaking change for pulp

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -194,7 +250,9 @@ class CollectionViewSet(
     filterset_class = CollectionFilter
     pagination_class = LimitOffsetPagination
 
-    DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
+    queryset_filtering_required_permission = "ansible.view_ansiblerepository"
```
newswangerd, line 253: We should only require users to have view permissions on the repository when the repository is private.

pulp_ansible/app/viewsets.py
```diff
@@ -111,6 +167,77 @@ class CollectionViewset(NamedModelViewSet, mixins.RetrieveModelMixin, mixins.Lis
...
+                "condition": "has_model_perms:ansible.add_collection",
+            },
+            {
+                "action": "destroy",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:ansible.delete_collection",
+                    "has_model_or_obj_perms:ansible.view_collection",
+                    "has_model_or_obj_perms:ansible.change_ansiblerepository",
```
newswangerd, line 192: @mdellweg @gerrod3 does this api endpoint allow you to upload a collection without specifying a repo?

If so, we probably don't need to require repo permissions here.

pulp_ansible/app/viewsets.py
```diff
@@ -183,6 +310,49 @@ class CollectionVersionViewSet(UploadGalaxyCollectionMixin, SingleArtifactConten
...
```
newswangerd, line 335: same question as the collections endpoint

newswangerd (commented):
I haven't tested this yet. The code looks solid. I think we still need to hash out how the access policies should work. Looking to @gerrod3 and @mdellweg for help here.

pulp_ansible/app/models.py
bmclaughlin: Yes, they should already be in the db since `django.contrib.auth` is in `INSTALLED_APPS`.  I'll remove the extraneous code.

pulp_ansible/app/models.py
bmclaughlin: @mdellweg  answered this above, I missed removing this instance.

pulp_ansible/app/models.py
gerrod3: I agree, let's try to reduce the duplicate permissions.

pulp_ansible/app/viewsets.py
gerrod3: If you follow the example in pulp_file (https://github.com/pulp/pulp_file/blob/main/pulp_file/app/viewsets.py#L88-L92), then you won't be able to upload unless it is to a repository that you have perms on.

gerrod3: I should also mention that this viewset isn't under the `/pulp/api/v3/` endpoints, it actually appears under `/ansible/collections/`. Personally I wouldn't add RBAC to these endpoints as the upload endpoint is deprecated and the list endpoint is kind of useless. If galaxy uses this viewset then maybe it would make sense to add a policy, otherwise I would leave it how it was.

pulp_ansible/app/viewsets.py
gerrod3: See my answer from collections endpoint. 

@bmclaughlin Collections Version doesn't have a destroy or update endpoint.

pulp_ansible/app/viewsets.py
```diff
@@ -224,6 +388,43 @@ class CollectionVersionSignatureViewSet(NoArtifactContentUploadViewSet):
...
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": "create",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_perms:ansible.add_collection",
+            },
```
gerrod3, line 405: I would repeat the same access policy I linked above for pulp_file content here. CollectionVersionSignatures also don't have a update or destroy endpoint.

pulp_ansible/app/viewsets.py
```diff
@@ -234,19 +435,96 @@ class CollectionDeprecatedViewSet(ContentViewSet):
...
```
gerrod3, line 452: Same here, CollectionDeprecated does not have a destroy or update endpoint.

pulp_ansible/app/viewsets.py
```diff
@@ -267,6 +585,55 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+            {
+                "action": [
+                    "modify",
+                    "partial_update",
+                    "rebuild_metadata",
+                    "repair",
+                    "sign",
+                    "sync",
+                    "update",
+                ],
```
gerrod3, line 618: We typically add a permission for each extra action that can be performed on a repository. https://github.com/pulp/pulp_file/blob/main/pulp_file/app/viewsets.py#L143-L170.  If this is too granular you can combine some actions into one permission, but I think it would be helpful having separate sync and sign permissions.

pulp_ansible/app/viewsets.py
```diff
@@ -373,12 +740,51 @@ def sign(self, request, pk):
...
+class AnsibleRepositoryVersionViewSet(RepositoryVersionViewSet, RolesMixin):
     """
     AnsibleRepositoryVersion represents a single file repository version.
     """
 
     parent_viewset = AnsibleRepositoryViewSet
+    queryset_filtering_required_permission = "ansible.view_ansiblerepository"
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
```
gerrod3, line 752: RepositoryVersionViewSet doesn't need a `queryset_filtering_required_permission`. I would follow the example in pulp_file for this one: https://github.com/pulp/pulp_file/blob/main/pulp_file/app/viewsets.py#L242

pulp_ansible/app/models.py
```diff
@@ -24,7 +25,7 @@
 log = getLogger(__name__)
 
 
-class Role(Content):
+class Role(Content, AutoAddObjPermsMixin):
```
gerrod3, line 28: You should remove the `AutoAddObjPermsMixin` from here and CollectionVersion model.

pulp_ansible/app/viewsets.py
```diff
@@ -88,6 +89,59 @@ class RoleViewSet(ContentViewSet):
...
+    queryset_filtering_required_permission = "ansible.view_role"
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_repository_model_or_obj_perms:ansible.view_role",
+            },
```
gerrod3, line 101: I would follow the same advice I gave for CollectionVersions here as well. Also you should remove the `LOCKED_ROLES`. This is not needed and I think it is breaking your tests.

pulp_ansible/app/viewsets.py
```diff
@@ -267,6 +459,91 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+                "action": ["sync"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:ansible.sync_ansiblerepository",
+                    "has_model_or_obj_perms:ansible.view_ansiblerepository",
+                ],
+            },
+        ],
+    }
```
gerrod3, line 534: You are missing two important sections for the policy: `creation_hooks` and `queryset_scoping`. I would add these sections for this policy and for remote and distribution as well. `queyset_scoping` should also be added to the content policiess as well. 
pulp_file example: https://github.com/pulp/pulp_file/blob/main/pulp_file/app/viewsets.py#L178-L185

pulp_ansible/tests/functional/api/test_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+    except ApiException as e:
+        assert e.status == outcome, f"{e}"
+    else:
+        assert status == outcome, f"User performed {action} when they shouldn't been able to"
+        return response
+
+
+def test_ansible_repository_rbac(gen_user):
+    ansible_client = gen_ansible_client()
+    repo_api = RepositoriesAnsibleApi(ansible_client)
```
gerrod3, line 39: Ok I figured out why the tests are failing, it's because you are not using the pytest client fixtures. You can find all the defined clients here: https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/conftest.py
```suggestion
def test_ansible_repository_rbac(gen_user, ansible_repo_api_client):
```

pulp_ansible/tests/functional/api/test_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Create testing
+    repo = try_action(user_creator, repo_api, "create", 201, gen_repo())
+    try_action(user_reader, repo_api, "create", 403, gen_repo())
+    try_action(user_helpless, repo_api, "create", 403, gen_repo())
+
+    # View testing
+    try_action(user_creator, repo_api, "read", 200, repo.pulp_href)
+    try_action(user_reader, repo_api, "read", 200, repo.pulp_href)
+    try_action(user_helpless, repo_api, "read", 200, repo.pulp_href)
```
gerrod3, line 58:
```suggestion
    try_action(user_helpless, repo_api, "read", 404, repo.pulp_href)
```
Queryset scoping prevents unauthorized users from even knowing the object exists.

pulp_ansible/tests/functional/api/test_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Update testing
+    update_args = [repo.pulp_href, gen_repo()]
+    try_action(user_creator, repo_api, "update", 202, *update_args)
+    try_action(user_reader, repo_api, "update", 403, gen_repo())
+    try_action(user_helpless, repo_api, "update", 403, gen_repo())
```
gerrod3, line 64:
```suggestion
    try_action(user_reader, repo_api, "update", 403, *update_args)
    try_action(user_helpless, repo_api, "update", 404, *update_args)
```

gerrod3 (changes requested):
Hopefully this helps you get the tests passing. I didn't leave comments on all the tests, just the first to hopefully unblock you.

pulp_ansible/app/models.py
```diff
@@ -421,7 +434,7 @@ def _reset_repository_last_synced_metadata_time(self):
         self.last_synced_metadata_time = None
 
 
-class AnsibleDistribution(Distribution):
+class AnsibleDistribution(Distribution, AutoAddObjPermsMixin):
```
gerrod3, line 437: I think you should also add `manage_roles_ansibledistribution` permission here.

pulp_ansible/app/viewsets.py
```diff
@@ -88,6 +89,47 @@ class RoleViewSet(ContentViewSet):
...
     filterset_class = RoleFilter
+    queryset_filtering_required_permission = "ansible.view_role"
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_repository_model_or_obj_perms:ansible.view_role",
```
gerrod3, line 100: When `queryset_scoping` is specified, **Content** Viewsets will scope the returned results based on permissions the user has on repositories, so these two lines are not needed.

```suggestion

    DEFAULT_ACCESS_POLICY = {
        "statements": [
            {
                "action": ["list", "retrieve"],
                "principal": "authenticated",
                "effect": "allow",
```

pulp_ansible/app/viewsets.py
```diff
@@ -88,6 +89,47 @@ class RoleViewSet(ContentViewSet):
...
+            },
+            {
+                "action": ["update"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_repository_model_or_obj_perms:ansible.change_role",
+                    "has_repository_model_or_obj_perms:ansible.view_role",
+                ],
+            },
```
gerrod3, line 129: There are no destroy or update actions for Content in Pulp APIs, so I would remove these.

pulp_ansible/app/viewsets.py
```diff
@@ -190,6 +234,28 @@ class CollectionVersionViewSet(UploadGalaxyCollectionMixin, SingleArtifactConten
     serializer_class = CollectionVersionSerializer
     filterset_class = CollectionVersionFilter
     ordering_fields = ("pulp_created", "name", "version", "namespace")
+    queryset_filtering_required_permission = "ansible.view_collection"
```
gerrod3, line 237: Same here, permissions for viewing content is based on the repository.
```suggestion
```

pulp_ansible/app/viewsets.py
```diff
@@ -224,6 +290,38 @@ class CollectionVersionSignatureViewSet(NoArtifactContentUploadViewSet):
     filterset_class = SignatureFilter
     queryset = CollectionVersionSignature.objects.all()
     serializer_class = CollectionVersionSignatureSerializer
+    queryset_filtering_required_permission = "ansible.view_ansiblerepository"
```
gerrod3, line 293: The parent Content viewset has special logic for figuring out what permission is needed on the repository, so this isn't needed either.
```suggestion
```

pulp_ansible/app/viewsets.py
```diff
@@ -224,6 +290,38 @@ class CollectionVersionSignatureViewSet(NoArtifactContentUploadViewSet):
...
+            {
+                "action": "update",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:ansible.change_collection",
+                    "has_model_or_obj_perms:ansible.view_collection",
+                    "has_model_or_obj_perms:ansible.change_ansiblerepository",
+                ],
+            },
```
gerrod3, line 321: Remove, there is no update action.

pulp_ansible/app/viewsets.py
```diff
@@ -224,6 +290,38 @@ class CollectionVersionSignatureViewSet(NoArtifactContentUploadViewSet):
...
+                "effect": "allow",
+            },
+            {
+                "action": "create",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:ansible.change_collection",
+                    "has_model_or_obj_perms:ansible.view_collection",
+                    "has_model_or_obj_perms:ansible.change_ansiblerepository",
```
gerrod3, line 309: CollectionSignatures are a content type in pulp_ansible, so I believe this should be the same as the first two conditions for CollectionVersion create.

pulp_ansible/app/viewsets.py
```diff
@@ -234,19 +332,88 @@ class CollectionDeprecatedViewSet(ContentViewSet):
     endpoint_name = "collection_deprecations"
     queryset = AnsibleCollectionDeprecated.objects.all()
     serializer_class = CollectionSerializer
+    queryset_filtering_required_permission = "ansible.view_collection"
```
gerrod3, line 335: CollectionDeprecated is also a content type in pulp_ansible.
```suggestion
```

pulp_ansible/app/viewsets.py
```diff
@@ -234,19 +332,88 @@ class CollectionDeprecatedViewSet(ContentViewSet):
...
+                "effect": "allow",
+            },
+            {
+                "action": "update",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:ansible.change_collection",
+                    "has_model_or_obj_perms:ansible.view_collection",
+                    "has_model_or_obj_perms:ansible.change_ansiblerepository",
```
gerrod3, line 351: Change action to `create` and change the condition to match the first two conditions of CollectionVersion create.

pulp_ansible/app/viewsets.py
```diff
@@ -234,19 +332,88 @@ class CollectionDeprecatedViewSet(ContentViewSet):
...
     endpoint_name = "role"
     queryset = RoleRemote.objects.all()
     serializer_class = RoleRemoteSerializer
-
-
-class GitRemoteViewSet(RemoteViewSet):
+    queryset_filtering_required_permission = "ansible.view_roleremote"
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
```
gerrod3, line 370: You added the `RolesMixin`, but forgot to add the actions from it to the policy.

pulp_ansible/app/viewsets.py
```diff
@@ -256,9 +423,56 @@ class GitRemoteViewSet(RemoteViewSet):
     endpoint_name = "git"
     queryset = GitRemote.objects.all()
     serializer_class = GitRemoteSerializer
-
-
-class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
+    queryset_filtering_required_permission = "ansible.view_gitremote"
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
```
gerrod3, line 429: Same here, need `RolesMixin` actions here.

pulp_ansible/app/viewsets.py
```diff
@@ -267,6 +481,112 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+                "condition": [
+                    "has_model_or_obj_perms:ansible.sign_ansiblerepository",
+                    "has_model_or_obj_perms:ansible.view_ansiblerepository",
+                ],
+            },
+            {
+                "action": ["sync"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
```
gerrod3, line 553: Need to add three more conditions `has_remote_param_model_or_obj_perms:ansible.view_(collection/role/git)remote`. This makes sure the user has permission to use the remote to sync.

pulp_ansible/app/viewsets.py
```diff
@@ -373,13 +693,51 @@ def sign(self, request, pk):
         return OperationPostponedResponse(result, request)
 
 
-class AnsibleRepositoryVersionViewSet(RepositoryVersionViewSet):
+class AnsibleRepositoryVersionViewSet(RepositoryVersionViewSet, RolesMixin):
```
gerrod3, line 696: Remove the `RolesMixin` and its actions. We don't assign roles/perms to repository versions, it's all based off the parent repository.

pulp_ansible/app/viewsets.py
```diff
@@ -416,6 +774,54 @@ class CollectionRemoteViewSet(RemoteViewSet):
     serializer_class = CollectionRemoteSerializer
     filterset_class = CollectionRemoteFilter
 
+    queryset_filtering_required_permission = "ansible.view_collectionremote"
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
```
gerrod3, line 780: Need to also add the `RolesMixin` actions here.

pulp_ansible/app/viewsets.py
```diff
@@ -483,6 +889,59 @@ class AnsibleDistributionViewSet(DistributionViewSet):
...
+        ],
+        "queryset_scoping": {"function": "scope_queryset"},
+    }
+
+    LOCKED_ROLES = {
+        "ansible.ansibledistribution_creator": ["ansible.add_ansibledistribution"],
+        "ansible.ansibledistribution_owner": [
+            "ansible.view_ansibledistribution",
+            "ansible.change_ansibledistribution",
+            "ansible.delete_ansibledistribution",
```
gerrod3, line 940: Don't forget to add the `ansible.manage_roles_ansibledistribution` permission here after you add it to the model.

gerrod3 (changes requested):
Lots of minor changes to the policies, haven't gotten to all of the tests yet.

pulp_ansible/app/viewsets.py
bmclaughlin: @gerrod, when I add the three conditions my sync test fails (500, `RuntimeError: Permission ansible.view_roleremote is not suitable for objects of class <class 'pulp_ansible.app.models.CollectionRemote'>.`) due to all of the conditions needing to be met.  When only adding the condition for the type of remote being used (CollectionRemote in this case) it works as expected.

gerrod3: Sounds like a bug in pulpcore, it doesn't account for repositories that can be synced from multiple remotes. I would just remove the check for now, we can add it back later once we fix it in pulpcore.

mdellweg: Can't we add or statements in the conditions? Sure, we would need to prevent a 500 and get a 403 instead.

bmclaughlin: Thanks @mdellweg, I tracked down `condition_expression` and implemented that to return 403 instead of the 500.

gerrod3:
@bmclaughlin I think the test is failing because of the new access policy on the unpaginated-collection-version endpoint. Before, anyone (no authentication) could access the endpoint, but now you need to be at least authenticated to get a response. When the pulp-to-pulp sync happens it is sending the requests as an unauthenticated user. If the new access policy behavior is intended then I suggest changing that failing test to have it add user credentials to the created remote used for the sync.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -201,7 +201,22 @@ class CollectionViewSet(
...
+                "action": "destroy",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:ansible.delete_collection",
+            },
+            {
+                "action": ["update", "partial_update"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:ansible.change_collection",
```
newswangerd, line 216: we shouldn't be using the collection version or collection permissions directly, since we don't ever track permissions on content. Instead permissions to perform operations on content in the repo should be based off of repository level permissions. For now you could probably use "modify repo content" as a proxy for this, but it may be worth it to create some more granular permissions such as "modify repo collections", "modify repo namespaces" etc.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -201,7 +201,22 @@ class CollectionViewSet(
...
     pagination_class = LimitOffsetPagination
 
-    DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": "destroy",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:ansible.delete_collection",
```
newswangerd, line 210: This should just be model permissions. This action will remove the collection from every repository in the system which is more of an admin level function. (Also, this is the exception to the "don't use collection permissions" rule)

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -480,7 +495,21 @@ class CollectionUploadViewSet(
     serializer_class = CollectionVersionUploadSerializer
     pulp_tag_name = "Pulp_Ansible: Artifacts Collections V3"
 
-    DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
+    queryset_filtering_required_permission = "ansible.view_ansiblerepository"
```
newswangerd, line 498: i think you can remove this. I'm not sure it does anything.

pulp_ansible/app/viewsets.py
```diff
@@ -195,6 +220,27 @@ class CollectionVersionViewSet(UploadGalaxyCollectionMixin, SingleArtifactConten
...
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": "create",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_required_repo_perms_on_upload:ansible.modify_ansible_repo_content",
```
newswangerd, line 235: repo level permissions don't make sense here because these apis aren't scoped to a repository to make changes on. Its probably okay to just use the v3 apis for creating and modifying collections and reserve these apis for admin users (same goes for the other content views in this file)

pulp_ansible/app/viewsets.py
```diff
@@ -409,6 +710,44 @@ class AnsibleRepositoryVersionViewSet(RepositoryVersionViewSet):
 
     parent_viewset = AnsibleRepositoryViewSet
 
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_repository_model_or_obj_perms:ansible.view_ansiblerepository",
```
newswangerd, line 719: It might make sense to inherit the permissions from the repo versions repository for this. if you have permissions to modify a repository, that should include it's versions.

pulp_ansible/tests/functional/conftest.py
```diff
@@ -222,3 +222,14 @@ def _build_and_upload_collection(ansible_repo=None, **kwargs):
...
+
+@pytest.fixture
+def create_ansible_cfg(server, user):
+    return (
+        "[pulp]\nserver_list = pulp_ansible\n\n"
+        "[pulp.pulp_ansible]\n"
+        "server = {}\n"
+        "username = {}\n"
+        "password = {}"
+    ).format(server, user.username, user.password)
```
mdellweg, line 235: If you want to make this a factory fixture, you need to add one more level of wrapping it in a function you can return as the fixture.

pulp_ansible/tests/functional/conftest.py
```diff
@@ -225,11 +226,21 @@ def _build_and_upload_collection(ansible_repo=None, **kwargs):
...
 @pytest.fixture
-def create_ansible_cfg(server, user):
-    return (
-        "[pulp]\nserver_list = pulp_ansible\n\n"
-        "[pulp.pulp_ansible]\n"
-        "server = {}\n"
-        "username = {}\n"
-        "password = {}"
-    ).format(server, user.username, user.password)
+def create_ansible_cfg():
```
mdellweg, line 229: At this point, i think you can just make this fixture request the tmp_path fixture.
And let the factory return the path to the created file.
Remember, it is a pathlib.Path object. See this link for inspiration:
https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture

bmclaughlin: Neat!  I would concur except that I allowed for the passed in path so this could be used in conjunction with setting the `cwd` of an `ansible-galaxy` command in a directory where it would also pick up this ansible config allowing for authentication.  I think the passed in path is preferable to string parsing in that use case?

mdellweg: Do we need the directory to preexist? If not, then we could just call the the fixture `ansible_dir_factory` and return the path of the tempdir with the configuration.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -482,7 +506,20 @@ class CollectionUploadViewSet(
...
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            _CAN_VIEW_REPO_CONTENT,
+            {
+                "action": ["create"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_perms:ansible.add_collection",
+                    "has_model_or_obj_perms:ansible.view_ansiblerepository",
```
gerrod3, line 518: I'm not sure this is doing what you think it is doing. `has_model_or_obj_perms` calls the viewset's `get_object` method if the user doesn't have model perms, and this viewsets `get_object` method behavior is undefined (possibly will throw an error) because `queryset = None`. I think this condition should use the new one you wrote, `v3_can_view_repo_content`.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -673,7 +710,17 @@ class AnsibleNamespaceViewSet(
         "metadata_sha256": ["exact", "in"],
     }
 
-    DEFAULT_ACCESS_POLICY = _PERMISSIVE_ACCESS_POLICY
+    # TODO: write an actual access policy
```
gerrod3, line 713: Is this to be done in a follow up PR?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -776,7 +823,20 @@ class CollectionVersionViewSet(
...
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            _CAN_VIEW_REPO_CONTENT,
+            {
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:ansible.delete_collection",
+                    "has_model_or_obj_perms:ansible.view_collection",
```
gerrod3, line 835: This viewset is for CollectionVersions which is a content type that we don't assign permissions to. So this probably needs a special condition method to find the Collection object that the CollectionVersion is a part of and then check the user's perms on that Collection.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -925,6 +985,8 @@ class CollectionImportViewSet(
     queryset = CollectionImport.objects.prefetch_related("task").all()
     serializer_class = CollectionImportDetailSerializer
 
+    queryset_filtering_required_permission = "ansible.view_ansiblerepository"
+
```
gerrod3, line 989: Since this viewset is not a `NamedModelViewSet`, I don't think `queryset_filtering_required_permission` does anything. If you want to do some queryset filtering I would just define a `get_queryset` method for this viewset and perform the filter there. See the last two paragraphs here: https://docs.pulpproject.org/pulpcore/plugin_dev/plugin-writer/concepts/rbac/queryset_scoping.html

pulp_ansible/app/models.py
```diff
@@ -541,3 +554,5 @@ class AnsibleDistribution(Distribution):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+
```
gerrod3, line 557: Can you remove the extra line here?

pulp_ansible/app/settings.py
```diff
@@ -27,3 +27,8 @@
 ANSIBLE_DEFAULT_DISTRIBUTION_PATH = None
 ANSIBLE_URL_NAMESPACE = ""
 ANSIBLE_COLLECT_DOWNLOAD_LOG = False
+
+DRF_ACCESS_POLICY = {
+    "dynaconf_merge_unique": True,
+    "reusable_conditions": ["pulp_ansible.app.global_access_conditions"],
+}
```
gerrod3, line 34: Nice!

pulp_ansible/app/viewsets.py
```diff
@@ -98,6 +99,28 @@ class RoleViewSet(ContentViewSet):
     serializer_class = RoleSerializer
     filterset_class = RoleFilter
 
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_repository_model_or_obj_perms:ansible.view_role",
```
gerrod3, line 108: You don't need a condition here, Roles is a content type and doesn't have permissions directly assigned to them.

pulp_ansible/app/viewsets.py
gerrod3: This is a Pulp RBAC design choice. When a plugin adds RBAC, content becomes accessible only through permissions on repositories. This includes uploads, you can only upload to repositories you have permission for. The reason being that users can only see content for repositories they have permissions on, so they can not see orphan content unless they are an admin. So now if you are uploading to this endpoint and are a non-admin you must specify the repository to upload to or you will get a 403.

pulp_ansible/app/viewsets.py
```diff
@@ -331,6 +523,115 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:ansible.delete_ansiblerepository",
+            },
+            {
+                "action": ["update", "partial_update"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:ansible.change_ansiblerepository",
+                    "has_model_or_obj_perms:ansible.view_ansiblerepository",
```
gerrod3, line 553: @mdellweg What do you think about removing the `view_ansiblerepository` perm condition on these actions since scope queryset already does the same check?

pulp_ansible/app/viewsets.py
```diff
@@ -331,6 +523,115 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+            {
+                "action": ["sync"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition_expression": [  # required to be single string
+                    "has_model_or_obj_perms:ansible.sync_ansiblerepository "
+                    "has_model_or_obj_perms:ansible.view_ansiblerepository "
+                    "(has_remote_param_model_or_obj_perms:ansible.view_collectionremote "
+                    "or has_remote_param_model_or_obj_perms:ansible.view_gitremote "
+                    "or has_remote_param_model_or_obj_perms:ansible.view_roleremote)"
```
gerrod3, line 600: I thought there should atleast be an `and` in between the statements.

pulp_ansible/app/viewsets.py
```diff
@@ -508,6 +809,44 @@ class AnsibleRepositoryVersionViewSet(RepositoryVersionViewSet):
...
+            {
+                "action": ["rebuild_metadata"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_repository_model_or_obj_perms:ansible.rebuild_metadata_ansiblerepository",
+                ],
+            },
+        ],
+        "queryset_scoping": {"function": "scope_queryset"},
```
gerrod3, line 847: I think `queryset_scoping` should be removed. This endpoint is already scoped by permission on the parent repository. Don't need to filter it any further.

pulp_ansible/app/viewsets.py
```diff
@@ -611,6 +1004,66 @@ class AnsibleDistributionViewSet(DistributionViewSet):
...
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": "create",
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_perms:ansible.add_ansibledistribution",
+                ],
```
gerrod3, line 1022: Need another condition check: `"has_repo_or_repo_ver_param_model_or_obj_perms:ansible.view_ansiblerepository",`

gerrod3 (changes requested):

pulp_ansible/app/viewsets.py
```diff
@@ -613,7 +613,12 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin, R
                 "parameters": {"roles": "ansible.ansiblerepository_owner"},
             },
         ],
-        "queryset_scoping": {"function": "scope_queryset"},
+        "queryset_scoping": {
+            "function": "get_ansible_repository_qs",
```
gerrod3, line 617: Nice use of the queryset_scoping feature!

pulp_ansible/app/viewsets.py
```diff
@@ -632,6 +637,24 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin, R
         "ansible.ansiblerepository_viewer": ["ansible.view_ansiblerepository"],
     }
 
+    def get_ansible_repository_qs(self, qs, repo_perm):
+        """
+        Returns a queryset by filtering out private repositories unless the user has the required
+        permission to view.
```
gerrod3, line 643: I just want to check, but galaxy wants the behavior for repositories to be "visible to any authenticated user if it is public, else hidden unless you have view permission"? This differs from Pulp's general policy of "only user with permissions can see the repo".

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -710,14 +710,26 @@ class AnsibleNamespaceViewSet(
...
     DEFAULT_ACCESS_POLICY = {
         "statements": [
             _CAN_VIEW_REPO_CONTENT,
             {
-                "action": "*",
-                "principal": "*",
+                "action": "create",
+                "principal": "authenticated",
+                "effect": "allow",
+                "conditions": "has_model_or_obj_perms:ansible.modify_ansible_repo_content",
```
gerrod3, line 720: I think any galaxy V3 endpoint using this condition `has_model_or_obj_perms:ansible.modify_ansible_repo_content` is incorrect. Remember this access condtion call's the view's `get_object` and checks the permission against the returned object. So for most of the V3 endpoints it will be checking a repository permission against the collection/collection-version object and thus would fail. I think you should expand your custom access condition (or create another one) to allow for checking the repo permission against the actual repo object.

pulp_ansible/app/global_access_conditions.py
```diff
@@ -1,15 +1,17 @@
+from pulp_ansible.app import models
+
+
 def v3_can_view_repo_content(request, view, action):
```
gerrod3, line 4: So here I would either modify this method to accept another argument for the **repository** permission you want to check against. Or just create another method with similar logic for extracting the repository and testing the permission against it and that should be the access condition used in your V3 access policies.

gerrod3 (changes requested):
One more change I think needs to be fixed in the V3 access policy. Nice work with the `private` field and custom `queryset_scoping`!

pulp_ansible/app/global_access_conditions.py
```diff
@@ -0,0 +1,38 @@
...
+
+def v3_can_view_repo_content(request, view, action):
+    """
+    Check if the repo is private, only let users with view repository permissions
+    view the collections here.
+    """
+
+    if "distro_base_path" in view.kwargs:
+        distro_base_path = view.kwargs["distro_base_path"]
+        repo = models.AnsibleDistribution.objects.get(base_path=distro_base_path).repository.cast()
```
gerrod3, line 12: This doesn't handle the scenario where the distribution is pointing at a repository version, or nothing at all!
```suggestion
        distro_base_path = view.kwargs["distro_base_path"]
        distro = models.AnsibleDistribution.objects.select_related("repository", "repository_version").get(base_path=distro_base_path)
        if not repo := distro.repository:
            if not repo_ver := distro.repository_version:
                 return False
            repo = repo_ver.repository
```

pulp_ansible/app/models.py
```diff
@@ -482,7 +490,22 @@ def last_sync_task(self):
...
+            (
+                "copy_collection_version_ansiblerepository",
+                "Can copy collection versions on the repository",
+            ),
+            (
+                "move_collection_version_ansiblerepository",
+                "Can move collection versions on the repository",
+            ),
+            ("manage_roles_ansiblerepository", "Can manage roles on repositories"),
+            ("modify_ansible_repo_content", "Can modify repository content"),
```
gerrod3, line 507: @newswangerd What do you think about these separate permissions for copy and move? They seem to overlap with modify, but modify is also being used for upload.

pulp_ansible/tests/functional/api/collection/v3/test_namespace.py
```diff
@@ -170,38 +170,51 @@ def test_namespace_syncing(
...
-    collections_string = "\n  - ".join((f"{c.namespace}.{c.name}" for c in collections[0:2]))
-    requirements = f"collections:\n  - {collections_string}"
-    remote = ansible_collection_remote_factory(
-        url=distro1.client_url, requirements_file=requirements
+    user = gen_user(
+        model_roles=[
+            "ansible.ansibledistribution_creator",
+            "ansible.ansiblerepository_creator",
+            "ansible.collectionremote_creator",
+        ]
```
gerrod3, line 181: Any particular reason you added a user to this test? They should run with admin permissions without specifying the user.

pulp_ansible/app/viewsets.py
gerrod3: I don't understand, your latest change reverted the behavior back to how the original `scope_queryset` method worked. Your custom method is making no real distinction between private and non-private repos now. What is the actual behavior a user should expect when they set the `private` flag to True?

gerrod3 (changes requested):

pulp_ansible/tests/functional/api/collection/v3/test_namespace.py
bmclaughlin: I added a user due to a 403 error that I ran into while fixing tests (see [current test failure](https://github.com/pulp/pulp_ansible/actions/runs/4471839876/jobs/7857260320?pr=1338#step:11:2697)). In my local logs I see the following which leads to the missing `path` attribute:
```
pulp_1   | Giving up download_wrapper(...) after 1 tries (aiohttp.client_exceptions.ClientResponseError: 403, message='Forbidden', url=URL('http://localhost:5001/pulp_ansible/galaxy/587105c2-32bf-4f58-ad8b-9b60c1ab6c6b/api/v3/collections/all/'))
pulp_1   | pulp [15fd9c46-dab0-41ea-8173-a07caaa77112]: backoff:ERROR: Giving up download_wrapper(...) after 1 tries (aiohttp.client_exceptions.ClientResponseError: 403, message='Forbidden', url=URL('http://localhost:5001/pulp_ansible/galaxy/587105c2-32bf-4f58-ad8b-9b60c1ab6c6b/api/v3/collections/all/'))
```

pulp_ansible/app/global_access_conditions.py
```diff
@@ -9,7 +9,13 @@ def v3_can_view_repo_content(request, view, action):
...
     if "distro_base_path" in view.kwargs:
         distro_base_path = view.kwargs["distro_base_path"]
-        repo = models.AnsibleDistribution.objects.get(base_path=distro_base_path).repository.cast()
+        distro = models.AnsibleDistribution.objects.select_related(
+            "repository", "repository_version"
+        ).get(base_path=distro_base_path)
+        if not (repo := distro.repository.cast()):
+            if not (repo_ver := distro.repository_version.cast()):
+                return False
+            repo = repo_ver.repository
```
gerrod3, line 18:
```suggestion
        if not (repo := distro.repository):
            if not (repo_ver := distro.repository_version):
                return False
            repo = repo_ver.repository
        repo = repo.cast()
```
You can't call `cast` on a repository version and you can't call it on a `None` object.

bmclaughlin: Doh.

pulp_ansible/app/models.py
newswangerd: I would just use the permissions for modify for copy and move. I would prefer to keep the permissions relatively primitive. Copy and move are just instances of modify and view

CHANGES/1290.feature
```diff
@@ -0,0 +1,7 @@
+Added Role Based Access Control.
+New default roles (creator, owner, viewer) have been added for ``AnsibleRepository``, ``AnsibleDistribution``,
+``CollectionRemote``, ``RoleRemote``, and ``GitRemote``.
+New detail role management endpoints (``my_permissions``, ``list_roles``, ``add_role``,
+``remove_role``) have been added to ``AnsibleRepository``, ``AnsibleDistribution``, ``CollectionRemote``,
+``GitRemote``, and ``RoleRemote``.
+
```
gerrod3, line 7: Can you fix this to be a just a new line at the end of the file?

gerrod3 (changes requested):
Nice tests! Last two things that need to be addressed before I approve:

1. https://github.com/pulp/pulp_ansible/pull/1338/files#r1141523751, I think you need to change this to my suggestion to ensure users with view perms on their private objects can see them
2. https://github.com/pulp/pulp_ansible/pull/1338/files#r1143944189 Based on David's reply I would remove the `copy` and `move` permissions and just have the access policy use the `modify_ansible_repo_content` permission for those actions.

gerrod3 (approved):
Nice! Thanks for all the hard work, LGTM!

PR #1348, opened 2023-02-04
Use semantic_version for version requirements

mdellweg:
Pythons pkg_resources Requirement introduced stricter PEP-440 checking
that is incompatible with ansible versions.

fixes https://github.com/pulp/pulp_ansible/issues/1349


gerrod3 (approved):
Should this be backported to 0.15?

mdellweg:
> Should this be backported to 0.15?

Maybe. Should we wait and see if the nightly pass with this?

PR #1355, opened 2023-02-10
Activate single commit check

mdellweg:
[noissue]


jctanner:
Why are you basically reverting https://github.com/pulp/pulp_ansible/pull/1252 ?

mdellweg:
It was an experiment, and it is not working out. The merging on pressing the button produces commit messages that are not checked by our ci and that way work against our backport automation. This moves the pain to later and leaves room for invalid commit messages. I will not merge this until discussed in the integration meeting.

PR #1357, opened 2023-02-13
Use a managed model to index collectionversions across repos.

jctanner:

maybe replaces https://github.com/pulp/pulp_ansible/pull/1337 ?
maybe replaces https://github.com/pulp/pulp_ansible/pull/1366 ?


jerabekjiri:
@jctanner in old API for the collection list/detail we have the dependencies field:
```
"dependencies": {
      "newswangerd.main_collection": ">=1.0.1"
}
```


also is sorting enabled? for UI we need to sort collections by:
- date created (`pulp_created`)
- namespace
- collection
- version

jctanner:
@jerabekjiri nope, but I'll work on that now.

jctanner:
@jerabekjiri okay it's added ...

order_by=pulp_created
order_by=-pulp_created
order_by=namespace
order_by=-namespace
order_by=name
order_by=-name
order_by=version
order_by=-version

Are all supported.

If you want to sort by "collection" aka namespace.name you can pass in a csv to the parameter like so `order_by=namespace,name`

jerabekjiri:
@jctanner thanks for adding sorting and dependencies :+1:,  tested and it's working well :)

any idea when the `repository_label` filter will be enabled?

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+def _is_indexable(repo, repo_v):
+    # don't bother updating the index with a distribution doesn't exist
+    # if no distro exists, remove the index
+    if not AnsibleDistribution.objects.filter(repository=repo, repository_version=repo_v).exists():
+        CVIndex.objects.filter(repository=repo, repository_version=repo_v).delete()
```
gerrod3, line 21: I think you should be using an OR Q object here. Distributions do not allow for repository and repository version to be set at the same time.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+    if is_latest:
+        repo_v = None
+    else:
+        repo_v = repository_version
+
+    if not _is_indexable(repository, repo_v):
+        return
+
+    cvs = repository_version.content.filter(pulp_type="ansible.collection_version")
+    cvs = [x.cast() for x in cvs]
```
gerrod3, line 62: I know this is ran in a separate task now so performance isn't the biggest issue, but I'm pretty sure each `cast` call results in a separate DB query.

Maybe use cvs query for an CollectionVersion query?
```suggestion
    cvs = repository_version.content.filter(pulp_type="ansible.collection_version").values_list("pk", flat=True)
    cvs = CollectionVersion.objects.filter(pk__in=cvs)
```

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+
+    namespaces = repository_version.content.filter(pulp_type="ansible.namespace")
+    namespaces = [x.cast() for x in namespaces]
+    namespaces = {x.name: x for x in namespaces}
+
+    # map out the namespace(s).name(s) for everything in the repo version
+    colset = set([(x.namespace, x.name) for x in cvs])
+
+    log.info("Repository collections ...")
+    log.info(colset)
```
gerrod3, line 81: Are these info statements still needed?

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+    deprecations = repository_version.content.filter(pulp_type="ansible.collection_deprecation")
+    deprecations = [x.cast() for x in deprecations]
+    deprecations_set = set([(x.namespace, x.name) for x in deprecations])
+
+    namespaces = repository_version.content.filter(pulp_type="ansible.namespace")
+    namespaces = [x.cast() for x in namespaces]
+    namespaces = {x.name: x for x in namespaces}
+
+    # map out the namespace(s).name(s) for everything in the repo version
+    colset = set([(x.namespace, x.name) for x in cvs])
```
gerrod3, line 78:
```suggestion
    colset = {(x.namespace, x.name) for x in cvs}
```

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+            )
+            cvi.delete()
+            continue
+        log.info(
+            (
+                f"KEEP {cvi.repository.name}"
+                + f".{cvi.collection_version.namespace}"
+                + f".{cvi.collection_version.name}"
+                + f".{cvi.collection_version.version} IN INDEX"
+            )
```
gerrod3, line 103: This seems like a lot of information to spam the logs with. Should this be on debug level?

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+    # iterate through each collection in the repository
+    for colkey in colset:
+        namespace = colkey[0]
+        name = colkey[1]
```
gerrod3, line 109:
```suggestion
        namespace, name = colkey
```

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # get all the versions for this collection
+        related_cvs = [x for x in cvs if x.namespace == namespace and x.name == name]
+
+        # compute the highest version (allow pre-releases too?)
+        versions = [x.version for x in related_cvs]
```
gerrod3, line 115: `semantic_version` should rate normal releases higher than pre-releases.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+        highest_version = versions[0]
+
+        # clean up any CV indexes with a version no longer in the repo ...
+        cvindexes = CVIndex.objects.filter(
+            repository=repository,
+            collection_version__namespace=namespace,
+            collection_version__name=name,
+        )
+        for cvi in cvindexes:
+            if cvi.collection_version.version not in versions:
```
gerrod3, line 127: I think you need to convert to a Version object first.
```suggestion
            if Version(cvi.collection_version.version) not in versions:
```

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,188 @@
...
+
+                crcv.highest_version = highest_version
+                crcv.is_highest = rcv.version == highest_version
+                crcv.is_signed = len(rcv_signatures) > 0
+                crcv.is_deprecated = is_deprecated
+                crcv.namespace_metadata = namespaces.get(rcv.namespace, None)
+
+                crcv.save()
+
+            except Exception as e:
```
gerrod3, line 152: Can we get a more specific Exception?

pulp_ansible/app/models.py
```diff
@@ -527,9 +535,9 @@ def finalize_new_version(self, new_version):
         remove_duplicates(new_version)
         validate_repo_version(new_version)
 
-    @hook(BEFORE_UPDATE, when="remote", has_changed=True)
-    def _reset_repository_last_synced_metadata_time(self):
-        self.last_synced_metadata_time = None
```
gerrod3, line 532: Why did you remove this hook? Is it not needed anymore?

pulp_ansible/app/galaxy/v3/viewsets.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # ordering_fields = ("collection_version.pulp_created",)
+
+    def get_queryset(self):
+        return (
+            CrossRepositoryCollectionVersionIndex.objects.select_related("repository")
+            .select_related("collection_version")
+            .select_related("repository_version")
+            .select_related("namespace_metadata")
+            .all()
```
gerrod3, line 79: This is a massive query. How is the performance?

gerrod3 (changes requested):

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+    distribution = filters.CharFilter(method="filter_by_distribution_id")
+    distribution_base_path = filters.CharFilter(method="filter_by_base_path")
+    dependency = filters.CharFilter(method="filter_by_dependency")
+    version_range = filters.CharFilter(field_name="version_range", method="version_range_filter")
+    repository_version = filters.CharFilter(
+        field_name="repository_version__number", method="repository_version_filter"
+    )
+    keywords = filters.CharFilter(field_name="q", method="filter_by_q")
+    repository_label = LabelFilter(label_field_name="repository__pulp_labels")
```
jerabekjiri, line 49: I don't think this is working, using a filter like `?repository_label=approved`, but the `pulp_labels` are in the format :arrow_down: 
```
pulp_labels: {
   pipeline: "approved"
}
```

jerabekjiri: okay, found out it can be used like `?repository_label=pipeline=approved`, my bad :facepalm:

jctanner: Yep, just make sure you encode that string.

pulp_ansible/app/models.py
jctanner: I think that was an accident. Let me re-add it.

pulp_ansible/app/tasks/collectionversion_index.py
jctanner:
```#!/usr/bin/env python

from semantic_version import Version

versions = ['1.0.0', '2.0.0', '3.0.0-rc.1']
versions = [Version(x) for x in versions]
print(max(versions))
```

Will print out the 3.0.0-rc.1 prerelease instead of 2.0.0

gerrod3: Hmm, probably need something like this then:

```python
release_versions = [x for x in versions if not x.prerelease]
# Account for scenario where there are only pre-releases
highest_version = release_versions[0] if release_versions else versions[0]
```

pulp_ansible/app/galaxy/v3/viewsets.py
jctanner: image

Performance is on par with other endpoints in the system as the CV count grows.  Uploads and syncs increase in duration, but I don't have baseline data to determine if that was always the case prior to this PR.

gerrod3 (changes requested):
Haven't gotten to the big test file yet (1000 lines :scream:), but lots of small optimizations and questions around the indexing task.

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,174 @@
...
+                for v in queryset.values_list("collection_version__version", flat=True)
+            ]
+            version_list = [str(v) for v in s.filter(full_version_list)]
+
+            return queryset.filter(collection_version__version__in=version_list)
+        except ValueError:
+            raise ValidationError(_("%s must be a valid semantic version range." % name))
+
+    def filter_by_repository_name(self, queryset, name, value):
+        repository_names = self.request.query_params.getlist(name)
```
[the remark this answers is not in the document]
jctanner: Given url params such as `limit=1000&repository_name=automation-hub-1&repository_name=automation-hub-2&repository_name=automation-hub-3&signed=True`, the filter function gets only the last entry for the "value".

```
pulp_1   | FBRN: name:repository_name value:automation-hub-3
pulp_1   | FBRN: get_list:['automation-hub-1', 'automation-hub-2', 'automation-hub-3']
```

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,210 @@
...
+        if CVIndex.objects.filter(Q(repository_version=repo_v) | Q(repository=repo)).exists():
+            return False
+    return True
+
+
+def delete_previous_repository_version_indexes(repository_version):
+    """Delete indexes for previous repository versions."""
+    repository = repository_version.repository
+    qs = RepositoryVersion.objects.filter(repository__pk=repository.pk)
+    for repo_version in qs:
```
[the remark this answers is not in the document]
jctanner: These are unused functions that we no longer need.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,210 @@
...
+    # iterate through each collection in the repository
+    for colkey in colset:
+        namespace, name = colkey
+
+        # get all the versions for this collection
+        related_cvs = [x for x in cvs if x.namespace == namespace and x.name == name]
+
+        # compute the highest version (allow pre-releases too?)
+        versions_strings = [x.version for x in related_cvs]
+        version_tuples_with_prereleases = [(x, Version(x)) for x in versions_strings]
```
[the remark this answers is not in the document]
jctanner: refactored into a simpler function.

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def filter_by_distribution_id(self, qs, name, value):
+        dist_ids = self.request.query_params.getlist(name)
+
+        query = Q()
+
+        for dist_id in dist_ids:
+            for distribution in models.AnsibleDistribution.objects.filter(pk=dist_id):
+                if distribution.repository_version:
+                    query = query | Q(repository_version__pk=distribution.repository_version.pk)
```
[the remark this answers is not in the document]
jctanner: It does. fixed.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,193 @@
...
+    namespaces = {x.name: x for x in namespaces}
+
+    # map out the namespace(s).name(s) for everything in the repo version
+    colset = {(x.namespace, x.name) for x in cvs}
+
+    # clean up any repo+namespace+name indexes that no longer live in the repo ...
+    for cvi in CVIndex.objects.filter(repository=repository):
+        if (cvi.collection_version.namespace, cvi.collection_version.name) not in colset:
+            cvi.delete()
+            continue
```
gerrod3, line 120:
```suggestion
    # map out the namespace(s).name(s) for everything in the repo version
    colset = set(cvs.values_list("namespace", "name").distinct())

    # clean up any CV indexes that no longer live in the repo ...
    CVIndex.objecst.filter(repository=repository).exclude(
        collection_version__pk__in=cvs.values_list("pk", flat=True)
    ).delete()
```
Let's let the DB do most of the work.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,193 @@
...
+        versions = [x.version for x in related_cvs]
+        (
+            CVIndex.objects.filter(
+                repository=repository,
+                collection_version__namespace=namespace,
+                collection_version__name=name,
+            )
+            .exclude(collection_version__version__in=versions)
+            .delete()
+        )
```
gerrod3, line 144: This can be removed if you use my other suggestion above.

pulp_ansible/app/urls.py
```diff
@@ -244,6 +245,11 @@
         views_v3.ClientConfigurationView.as_view(),
         name="client-configuration-viewset",
     ),
+    path(
+        "search/collection-versions/",
+        viewsets_v3.CollectionVersionSearchViewSet.as_view({"get": "list", "put": "update"}),
```
gerrod3, line 250: Do we want the rebuild index action to be a `PUT` request or a `POST` request? I'm more partial to `POST` since we are not updating just one individual index object.

gerrod3 (changes requested):
Are ContentMarks suppose to be a part of the x-repo search? I don't see them anywhere in index model/serializer.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,193 @@
...
+        if distribution.repository_version:
+            repository_version = distribution.repository_version
+
+        # sometimes distros point at a repository
+        elif distribution.repository is not None:
+            repository = distribution.repository
+            repository_version = distribution.repository.latest_version()
+            is_latest = True
+
+        # sometimes distros point at nothing?
```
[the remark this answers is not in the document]
jctanner: k, if we go down that route it should be in another PR.

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,162 @@
...
+                else:
+                    query = query | Q(repository__pk=distribution.repository_id)
+
+        return qs.filter(query)
+
+    def repository_version_filter(self, qs, name, value):
+        if value == "latest":
+            return qs.filter(repository_version=None)
+        else:
+            return qs.filter(repository_version__number=value)
```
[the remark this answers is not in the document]
jctanner: You are correct. I refactored to function into an inefficient but functional queryset for the latest version of each repo.

jctanner:
@gerrod3 contentmarks weren't in the original requirements, so i'm not sure. @newswangerd thoughts?

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,176 @@
...
+        for rpk, number in RepositoryVersion.objects.values_list("repository__pk", "number"):
+            rpk = str(rpk)
+            if rpk not in vmap:
+                vmap[rpk] = number
+            if number > vmap[rpk]:
+                vmap[rpk] = number
+
+        query = Q()
+        for rpk, highest_number in vmap.items():
+            query = query | Q(repository__pk=rpk, repository_version__number=highest_number)
```
gerrod3, line 135:
```suggestion
       query = Q()
       for rpk, number in RepositoryVersion.objects.filter(repository__pulp_type="ansible.ansible").order_by("repository", "-pulp_created").distinct("repository").values_list("repository__pk", "number"):
           query = query | Q(repository__pk=rpk, repository_version__number=number)
```
Sorry for the bad formatting, but I condensed the finding latest version logic into DB operations. If you think it's too complex you don't need to change it, but I would atleast add the `filter(repo__pulp_type)` so that we only loop over repo-versions for Ansible repos.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,201 @@
...
+        return str(stable_versions[0])
+    return str(versions[0])
+
+
+def update_index(distribution=None, repository=None, repository_version=None, is_latest=False):
+    """Rebuild index by distribtion|repository|repositoryversion."""
+
+    # if the distro points at a specific repo version, we should use that in the index
+    # otherwise the index value for repository version should be null
+    use_repository_version = False
```
gerrod3, line 43: I am not sure what is going on with all the commented out code below, but I am beginning to believe we should always specify a repository version when creating an index.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -0,0 +1,1201 @@
...
+    # repov=1 + cv1
+    # repov=2 + cv1,cv2
+    # repov=null + cv1,cv2,cv3
+
+    resp = galaxy_v3_default_search_api_client.list(limit=1000, repository=[repo_id])
+
+    assert resp.meta.count == 6
+    repo_versions = sorted([x.repository_version for x in resp.data])
+    assert repo_versions == ["1", "2", "2", "latest", "latest", "latest"]
+    # import epdb; epdb.st()
```
gerrod3, line 1201:
```suggestion
```

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -0,0 +1,1201 @@
...
+            "name": pulp_repo.name,
+            "base_path": pulp_repo.name,
+            "repository_version": repository_version_latest.pulp_href,
+        },
+    )
+
+    # make sure the CV was indexed
+    dist_id = distro.pulp_href.split("/")[-2]
+    resp = galaxy_v3_default_search_api_client.list(limit=1000, distribution=[dist_id])
+    assert resp.meta.count == 1
```
gerrod3, line 1017: Weird that the generated client doesn't return the response with a direct `count` property like how it does for the Pulp clients.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -0,0 +1,1201 @@
...
+        ansible_distro_api_client,
+        {
+            "name": pulp_repo.name + "_vN",
+            "base_path": pulp_repo.name + "_vN",
+            "repository": pulp_repo.pulp_href,
+        },
+    )
+
+    # per david, we should have 6 entries now
+    # v1,2,3 w/ rv=null, v1,2 with rv=1,2, v1 with rv=1
```
gerrod3, line 1190:
```suggestion
    # cv1,2,3 w/ rv=null, cv1,2 with rv=1,2, cv1 with rv=1
```
I think this is what you meant, the comment is confusing otherwise.

gerrod3 (changes requested):
I am pretty confused now about why we are no longer deleting old indices.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,201 @@
...
+    # optimization: -must- have an AnsibleRepository for the index model
+    if not isinstance(repository, AnsibleRepository):
+        repository = AnsibleRepository.objects.filter(pk=repository.pk).first()
+        if repository is None:
+            return
+
+    # optimization: we only want to index "distributed" CVs
+    if distribution is None and not has_distribution(repository, repository_version):
+        # clean indexes for this repository or repository version
+        # CVIndex.objects.filter(
```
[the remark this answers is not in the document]
jctanner: An index with (repository=foo, repository_version=1) and (repository=foo, repository_version=None) are two unique keys based on the idea that a distribution can point at a repository and repository version as you have pointed out in previous comments. If there's an index for the repo+repoversion, that index needs to continue to exist until the repository version is removed. That behavior is taken care of by the `on_delete=models.CASCADE`  kwarg for the repository_version in the model.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
jctanner:
```(Epdb) resp.count
*** AttributeError: 'PaginatedCollectionVersionSearchListResponseList' object has no attribute 'count'
(Epdb) resp.meta.count
1
```

pulp_ansible/app/tasks/collectionversion_index.py
jctanner: The commented out code was mostly oriented around the mistaken belief that repository versions would always be "latest". However, the expectations have changed after yours and @newswangerd's reviews so the code was wrong. The additional tests clarified and assert the current behavior.

We can't always set a repository version because we need to ensure that (repository, no version, cv1) and (repository, repository version 1, cv1) are two unique results that come back from the viewset. I think it would have been more obvious if we had kept the distribution field in the index model, but we currently have 10's of thousands of "synclist" distributions pointing to the same repository in CRC so we had to get rid of that column to prevent possibly millions of duplicate entries.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # map out the namespace(s).name(s) for everything in the repo version
+    colset = set(cvs.values_list("namespace", "name").distinct())
+
+    # clean up any repo+namespace+name indexes that no longer live in the repo ...
+    if not use_repository_version:
+        # CVIndex.objects.filter(repository=repository, repository_version=None).exclude(
+        #    collection_version__pk__in=cvs.values_list("pk", flat=True)
+        # ).delete()
+        pass
```
gerrod3, line 119: This needs to come back or the latest index results will possibly include CVs that are not present.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # iterate through each collection in the repository
+    for colkey in colset:
+        namespace, name = colkey
+
+        # get all the versions for this collection
+        related_cvs = cvs.filter(namespace=namespace, name=name).only(
+            "namespace", "name", "version"
+        )
```
gerrod3, line 128:
```suggestion
        related_cvs = cvs.filter(namespace=namespace, name=name).only("version")
```
These optimizations are getting pedantic, but `version` is the only field we actually need.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,179 @@
...
+            # create|update the index for this CV
+            CVIndex.objects.update_or_create(
+                repository=repository,
+                repository_version=repo_v,
+                collection_version=rcv,
+                defaults={
+                    "is_highest": rcv.version == highest_version,
+                    "is_signed": rcv_signatures > 0,
+                    "is_deprecated": is_deprecated,
+                    "namespace_metadata": namespaces.get(rcv.namespace, None),
```
gerrod3, line 154:
```suggestion
                    "namespace_metadata": namespaces.get(namespace, None),
```
All of the related_cvs will have the same namespace.

gerrod3 (changes requested):

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -0,0 +1,174 @@
...
+            return qs.filter(repository_version__number=value)
+
+        # Reduce queryset down to the "latest" repository version number(s)
+        rqs = (
+            RepositoryVersion.objects.filter(repository__pulp_type="ansible.ansible")
+            .order_by("repository", "-pulp_created")
+            .distinct("repository")
+            .values_list("repository__pk", "number")
+        )
+        query = Q()
```
gerrod3, line 131:
```suggestion
        query = Q(repository_version=None)
```
I think this part got dropped when I mentioned the problem of distros pointing to repo-vers that happened to be latest.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -0,0 +1,179 @@
...
+    versions_strings = [x.version for x in cv_objects]
+    versions = [Version(x) for x in versions_strings]
+    versions = sorted(versions, reverse=True)
+    stable_versions = [x for x in versions if not x.prerelease]
+    if stable_versions:
+        return str(stable_versions[0])
+    return str(versions[0])
+
+
+def update_index(distribution=None, repository=None, repository_version=None, is_latest=False):
```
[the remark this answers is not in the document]
jctanner: The issues brought up in this comment are addressed by new tests and related logic changes.

newswangerd (approved):
Need to uncomment the namespace metadata, but I don't want to block you by requiring a re-review, so approved

awcrosby (approved):

gerrod3 (approved):

Issue #1369, opened 2023-02-22
Duplicate 'Downloading Artifacts' progress report is from the downloading docs blob stage


gerrod3:

mdellweg:
Why is this an issue? If downloading one collection involved downloading 3 artifacts, that is exactly what we should report.

PR #1370, opened 2023-02-22
Rename duplicate Downloading Artifacts Progress Report

gerrod3:
fixes: #1369
Requires PR: https://github.com/pulp/pulpcore/pull/3597

### DIFF
--- a/CHANGES/1369.bugfix
+++ b/CHANGES/1369.bugfix
@@ -1,0 +1,1 @@
+Renamed duplicate 'Downloading Artifacts' sync progress report message to 'Downloading Docs Blob'.
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -42,6 +42,7 @@ from pulpcore.plugin.stages import (
     DeclarativeArtifact,
     DeclarativeContent,
     DeclarativeVersion,
+    GenericDownloader,
     QueryExistingArtifacts,
     QueryExistingContents,
     RemoteArtifactSaver,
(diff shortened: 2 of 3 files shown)


mdellweg:
Can we think of other solutions? I'd like the stage to only have a single download stage.
Do i see it correctly, that the docs_blob url is available at the time we create the declarative content? Can we make it a second declarative artifact that we extract again before saving? Or we introduce a DeclarativeBlob that is also downloaded in the artifact_download stage. Is the docs blob meant to be an attached artifact for the content unit?

gerrod3:
> Can we think of other solutions?

Yes, this is a quick fix for the problem noticed by David after looking into total count.

> I'd like the stage to only have a single download stage. Do i see it correctly, that the docs_blob url is available at the time we create the declarative content?

Yes, it's available on the collection version metadata. It's optional though, so it is not always present.

> Can we make it a second declarative artifact that we extract again before saving? Or we introduce a DeclarativeBlob that is also downloaded in the artifact_download stage. Is the docs blob meant to be an attached artifact for the content unit?

DocsBlob is a JSON field on the collection-version model. It can't be added as a second declarative artifact without a serious rework. We could however just override the first downloader stage and have it download the docs-blob with the artifact if the url is present.

Do you want me to look into merging this downloader stage into the first artifact downloader stage?

mdellweg:
> Do you want me to look into merging this downloader stage into the first artifact downloader stage?
Yes, I would prefer that.

I was thinking about a stage between ArtifactDownloader and ArtifactSaver that would find the "extra" d_artifact and turn it into the docs_blob.

Issue #1356, opened 2023-02-10
 duplicate operationId in generated openapi spec


jlsherrill:
**Version**
0.16.0

**Describe the bug**
 in the generated openapi spec, there seems to be a duplicated operationId:

```
        "{ansible_role_href}versions/": {
            "get": {
                "operationId": "versions_get",
                "description": "APIView for Role Versions.",
```

```
        "{ansible_collection_version_href}versions/": {
            "get": {
                "operationId": "versions_get",
                "description": "APIView for Collections by namespace/name.",
                "parameters": [

```
I don't think this is allowed and i'm seeing some go bindings generation generating bad code due to it.

jlsherrill:
related to https://github.com/pulp/pulp_python/issues/594

PR #1371, opened 2023-02-23
Fix duplicate operationIDs in Galaxy API

gerrod3:
fixes: #1356

### DIFF
--- a/pulp_ansible/app/galaxy/views.py
+++ b/pulp_ansible/app/galaxy/views.py
@@ -2,7 +2,7 @@ import re
 
 from django.conf import settings
 from django.shortcuts import get_object_or_404, HttpResponse
-from drf_spectacular.utils import extend_schema
+from drf_spectacular.utils import extend_schema, extend_schema_view
 from rest_framework import generics, pagination, response, views
 
 from pulpcore.plugin.models import PulpTemporaryFile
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/galaxy/views.py
```diff
@@ -189,6 +190,7 @@ def post(self, request, path):
         return OperationPostponedResponse(async_result, request)
 
 
+@extend_schema_view(get=extend_schema(operation_id="api_v2_collection_versions_list"))
```
mdellweg, line 193: I'm not too good glad that the liveapi is part in our open API, but I guess it's too late anyway.

mdellweg (approved):
Approving here, but can we think about a way to prefix all APIs in the galaxy part?

gerrod3:
> Approving here, but can we think about a way to prefix all APIs in the galaxy part?

What type of prefix would you want?

gerrod3:
@jlsherrill Do you need these backported to 0.16?

jlsherrill:
i don't really need a backport, but being included in a release sooner rather than later would be helpful.  How long till 0.17?

gerrod3:
> i don't really need a backport, but being included in a release sooner rather than later would be helpful. How long till 0.17?

It will be release when Repo cross search and rbac are merged. It is also requires pulpcore 3.23, so sometime in March.

mdellweg:
`galaxy_api_v1/2` would be compelling, no?
or just `galaxy_v1/2`. We know we are in an api definition at that point.

jlsherrill:
that time frame should be okay, no backport needed.

PR #1374, opened 2023-02-23
Fix excludes list on collection sync

newswangerd:
Attempting to fix this CI failure: https://github.com/ansible/galaxy_ng/actions/runs/4247017139/jobs/7384551387

fixes: #1381


mdellweg (commented):
Looks good to me. Too bad this was not caught by a test. If i understand correctly, this is fixing a bug. We should have a changelog.

We may need to backport this to 0.15 and 0.16.

PR #1376, opened 2023-02-24
Fix url parsing for collections with a name of "collections"

jctanner:
https://issues.redhat.com/browse/AAH-2158
https://github.com/pulp/pulp_ansible/issues/1388

An alternative fix is to change
```
path("content/<path:distro_base_path>/collections/", include(v3_collection_urls))
```
to
```
path("content/<str:distro_base_path>/collections/", include(v3_collection_urls))
```

However, that will not allow for base paths containing "/"'s

### DIFF
--- a/pulp_ansible/app/urls.py
+++ b/pulp_ansible/app/urls.py
@@ -174,28 +174,6 @@ v3_collection_urls = [
         name="collection-artifact-download",
     ),
     path(
-        "index/<str:namespace>/<str:name>/",
-        views_v3.CollectionViewSet.as_view(
-            {"get": "retrieve", "patch": "update", "delete": "destroy"}
-        ),
-        name="collections-detail",
-    ),
-    path(
-        "index/<str:namespace>/<str:name>/versions/",
-        views_v3.CollectionVersionViewSet.as_view({"get": "list"}),
-        name="collection-versions-list",
-    ),
-    path(
-        "index/<str:namespace>/<str:name>/versions/<str:version>/",
-        views_v3.CollectionVersionViewSet.as_view({"get": "retrieve", "delete": "destroy"}),
-        name="collection-versions-detail",
-    ),
-    path(
-        "index/<str:namespace>/<str:name>/versions/<str:version>/docs-blob/",
-        views_v3.CollectionVersionDocsViewSet.as_view({"get": "retrieve"}),
-        name="collection-versions-detail-docs",
-    ),
-    path(
         "all-collections/",
         views_v3.UnpaginatedCollectionViewSet.as_view({"get": "list"}),
         name="metadata-collection-list",
@@ -204,6 +182,31 @@ v3_collection_urls = [
         "all-versions/",
         views_v3.UnpaginatedCollectionVersionViewSet.as_view({"get": "list"}),
         name="metadata-collection-versions-list",
+    ),
+]
+
+v3_collection_detail_urls = [
+    path(
+        "",
+        views_v3.CollectionViewSet.as_view(
+            {"get": "retrieve", "patch": "update", "delete": "destroy"}
+        ),
+        name="collections-detail",
+    ),
+    path(
+        "versions/",
+        views_v3.CollectionVersionViewSet.as_view({"get": "list"}),
+        name="collection-versions-list",
+    ),
+    path(
+        "versions/<str:version>/",
+        views_v3.CollectionVersionViewSet.as_view({"get": "retrieve", "delete": "destroy"}),
+        name="collection-versions-detail",
+    ),
+    path(
+        "versions/<str:version>/docs-blob/",
+        views_v3.CollectionVersionDocsViewSet.as_view({"get": "retrieve"}),
+        name="collection-versions-detail-docs",
     ),
 ]
 
@@ -225,6 +228,10 @@ namespace_urls = [
 v3_plugin_urls = [
     # path:var captures /, so it has to have something at the end to make it work
     # correctly.
+    path(
+        "content/<path:distro_base_path>/collections/index/<str:namespace>/<str:name>/",
+        include(v3_collection_detail_urls),
+    ),
     path("content/<path:distro_base_path>/collections/", include(v3_collection_urls)),
     path("content/<path:distro_base_path>/namespaces/", include(namespace_urls)),
     path(
--- a/pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
+++ b/pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
@@ -1,0 +1,174 @@
+import hashlib
+import json
+import shutil
+import subprocess
+import os
+
+import pytest
+
+from orionutils.generator import build_collection
+from orionutils.generator import randstr
+
+from pulpcore.client.pulp_ansible.exceptions import ApiException
+
+
+def busted_collection_build(basedir=None, namespace=None, name=None, version=None):
+    """Make artifacts with namespaces and names that wouldn't normally be allowed."""
+
+    def file_checksum(fp):
+        with open(fp, "rb") as f:
+            checksum = hashlib.sha256(f.read()).hexdigest()
+        return checksum
+
+    # make a content dir
+    content_dir = os.path.join(basedir, "content")
+    os.makedirs(content_dir)
+
+    # make various files
+    for dname in ["docs", "plugins", "roles"]:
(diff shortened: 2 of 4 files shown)


pulp_ansible/app/urls.py
```diff
@@ -13,13 +14,28 @@
...
 
 GALAXY_API_ROOT = getattr(settings, "GALAXY_API_ROOT", "pulp_ansible/galaxy/<path:path>/api/")
 
 
+class BasePathConverter(PathConverter):
+    """Only allows valid basepaths in the URLs."""
+
+    @property
+    def regex(self):
+        basepaths = [x.base_path for x in AnsibleDistribution.objects.all()]
```
mdellweg, line 31: This is probably a very costly lookup. Not sure how we can solve this properly.

pulp_ansible/app/urls.py
```diff
@@ -225,8 +241,8 @@
 v3_plugin_urls = [
     # path:var captures /, so it has to have something at the end to make it work
     # correctly.
-    path("content/<path:distro_base_path>/collections/", include(v3_collection_urls)),
-    path("content/<path:distro_base_path>/namespaces/", include(namespace_urls)),
+    path("content/<basepath:distro_base_path>/collections/", include(v3_collection_urls)),
+    path("content/<basepath:distro_base_path>/namespaces/", include(namespace_urls)),
```
mdellweg, line 245: So this is an issue for collections named "namespaces" too. What about a namespace called "collections" or "namespaces"?

jctanner:
@mdellweg I think the latest commit fixes it in a cleaner way. By being more specific about the route the detail urls are included on top of, the path logic seems to work better.

pulp_ansible/app/urls.py
```diff
@@ -225,8 +228,12 @@
 v3_plugin_urls = [
     # path:var captures /, so it has to have something at the end to make it work
     # correctly.
-    path("content/<path:distro_base_path>/collections/", include(v3_collection_urls)),
-    path("content/<path:distro_base_path>/namespaces/", include(namespace_urls)),
+    path(
+        "content/<path:distro_base_path>/collections/index/<str:namespace>/<str:name>/",
+        include(v3_collection_detail_urls),
+    ),
```
mdellweg, line 234: So we give the regex more to munch on? I wonder if you could still craft a collection name to break this... But we don't allow "/" there, right?

jctanner: The "/" character should be an invalid one for a collection namespace or name because it wouldn't result in a valid python package name. I can add another test to validate that the import process fails on such a thing though.

mdellweg: I think, it's ok.

pulp_ansible/app/urls.py
```diff
@@ -225,8 +228,12 @@
...
+    path("content/<str:distro_base_path>/collections/", include(v3_collection_urls)),
+    path("content/<str:distro_base_path>/namespaces/", include(namespace_urls)),
```
mdellweg, line 236: Should these still exist? If yes, should they not still use the "path" matcher?

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,336 @@
+import json
+import logging
+import shutil
+import subprocess
+import os
+from urllib.parse import urljoin
+
+import pytest
```
mdellweg, line 8: Yay for adopting pytest style!

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,336 @@
...
+def pulp_client():
+    cfg = config.get_config()
+    client = api.Client(cfg)
+    headers = cfg.custom.get("headers", None)
+    if headers:
+        client.request_kwargs.setdefault("headers", {}).update(headers)
+
+    client.client_side_validation = False
+
+    return client
```
mdellweg, line 122: Is the fixture from conftest not serving the purpose here? I'd prefer not to duplicate it.

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,336 @@
...
+                            cv_info["version"],
+                        )
+                        cvs[key] = cv_info
+
+    return cvs
+
+
+def get_collection_versions(pulp_client):
+    cversions = pulp_client.get("/pulp/api/v3/content/ansible/collection_versions/")
+    return dict(((x["namespace"], x["name"], x["version"]), x) for x in cversions)
```
mdellweg, line 209: We are moving towards using fixtures that encapsulate bindings classes.
This one should do the trick:
https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/conftest.py#L91

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,336 @@
...
+        dist = dists[0]
+        created = False
+    else:
+        raise ValueError("Found too many Ansible Distributions at 'automation-hub'.")
+    yield dist
+    if created:
+        pulp_client.delete(dist["pulp_href"])
+
+
+@pytest.mark.pulp_on_localhost
```
mdellweg, line 258: What is the action that warrants this mark?

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,336 @@
...
+def test_collection_named_collections(pulp_client, tmp_path, ansible_repo_api_client, pulp_dist):
+    # https://issues.redhat.com/browse/AAH-2158
+    delete_orphans()
```
mdellweg, line 261: Do we need this?
Can we instead randomize the namespace? It may allow to mark the test as parallel.

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # validate we can get the direct path to the collection ...
+    resp = galaxy_v3_content_collection_index_api.read(
+        pulp_dist.base_path, namespace=spec["namespace"], name=spec["name"]
+    )
+    assert resp.namespace == spec["namespace"]
+    assert resp.name == spec["name"]
+
+
+def test_collection_named_with_slashes(
```
mdellweg, line 137:
```suggestion
@pytest.mark.parallel
def test_collection_named_with_slashes(
```

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,168 @@
...
+    with open(os.path.join(content_dir, "MANIFEST.json"), "w") as f:
+        f.write(json.dumps(mdata, indent=2))
+
+    # make the tarball
+    cmd = "tar czvf artifact.tar.gz *"
+    subprocess.run(cmd, shell=True, cwd=content_dir)
+    return os.path.join(content_dir, "artifact.tar.gz")
+
+
+def test_collection_named_collections(
```
mdellweg, line 98:
```suggestion
@pytest.mark.parallel
def test_collection_named_collections(
```
With the randomized name, we should be safe running these tests in parallel with others.

pulp_ansible/tests/functional/api/collection/v3/test_collection_naming_edgecases.py
```diff
@@ -0,0 +1,168 @@
...
+from orionutils.generator import randstr
+
+from pulp_smash.pulp3.bindings import monitor_task
+
+
+def busted_collection_build(basedir=None, namespace=None, name=None, version=None):
+    """Make artifacts with namespaces and names that wouldn't normally be allowed."""
+
+    def file_checksum(fp):
+        """Quick sha256 maker."""
```
mdellweg, line 16: Any reason to not using hashlib?

mdellweg (commented):
Thank you for refining the tests. This is so much easier to review now.

mdellweg (approved):

gerrod3 (approved):

mdellweg (approved):

PR #1385, opened 2023-03-01
hotfix: Add default value for signatures and marks

rochacbruno:
When syncing from a remote that has no marks or no signatures (older version) this ensures the field on sync defaults to empty list avoiding iteration error.

fixes: #1387

### DIFF
--- a/CHANGES/1387.bugfix
+++ b/CHANGES/1387.bugfix
@@ -1,0 +1,1 @@
+Fixed broken sync form servers without signatures or marks.
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -589,8 +589,8 @@ class CollectionSyncFirstStage(Stage):
         self.already_synced.add(cv_unique)
 
         info = metadata["metadata"]
-        signatures = metadata.get("signatures")
-        marks = metadata.get("marks")  # List[str]
+        signatures = metadata.get("signatures", [])
+        marks = metadata.get("marks", [])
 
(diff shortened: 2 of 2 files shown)


awcrosby (approved):

mdellweg (commented):
Is this fixing a bug in the already released signature sync?

rochacbruno:
@mdellweg this is fixing a problem introduced by the marks PR, it raised error on our QE sync test when the client expected a field named `marks` but the server was not serving it (outdated version)

mdellweg:
> @mdellweg this is fixing a problem introduced by the marks PR, it raised error on our QE sync test when the client expected a field named `marks` but the server was not serving it (outdated version)

Yes, but it changes a line with signatures too. And if that is a sneaky bugfix, we'd need a ticket for it. Basically I'm asking what the effect of the change in line 592 is.

rochacbruno:
@mdellweg I added a default value to signatures to follow the pattern and avoid the situation where a client syncs from a server that doesn't support signatures.

mdellweg:
OK, I read it properly in that case: This is a bugfix even for a released feature. Namely syncing from a server without sigs. We need a changelog.

CHANGES/1387.bugfix
```diff
@@ -0,0 +1 @@
+Add default value for signatures and marks on sync
```
mdellweg, line 1: How about this?
```suggestion
Fixed broken sync form servers without signatures or marks.
```

mdellweg (approved):

gerrod3 (approved):

PR #1389, opened 2023-03-06
Fixed pulpcore deprecations

gerrod3:
[noissue]


ipanova (approved):
I see main branch already has pulpcore version bumped to 3.23.

Issue #1301, opened 2022-11-30
Support for Sigstore signature generation and verification


mayaCostantini:
**Is your feature request related to a problem? Please describe.**
Implement [Sigstore](https://www.sigstore.dev/) signature generation and verification for Ansible collections in addition to the existing GPG signature workflow.

**Describe the solution you'd like**
Sigstore signing and verification for Ansible collections will provide the following advantages:

- **Key management**: Sigstore solves the issue of private key storage and revocation by providing “keyless” signatures for software artifacts, eliminating the need for the user to store private GPG keys which could eventually become compromised.

- **Identity-based signatures**: Sigstore generates free, short-lived signing certificates based on identities, delegating the authentication part to an OIDC provider (Google, GitHub, Microsoft…).

- An **immutable Transparency Log** (Rekor) for signature and provenance attestations that can be deployed as a private instance.

The signing and verification workflow for collections uploaded to pulp_ansible will be as follow:

- **Uploading signed collections**
A developer creates a collection, generates a `MANIFEST.in` file containing a list of collection files along with their sha256 checksums, and signs this file using the `pulp ansible` or [ansible-sign](https://github.com/ansible/ansible-sign) command lines supporting sigstore signatures. The developer can sign the collection using a key pair or via an [OIDC identity](https://docs.sigstore.dev/cosign/openid_signing) ("keyless" signing). It is possible to specify the Fulcio (certificate authority) and Rekor (transparency log) instances to use by passing the instance URLs as command line parameters. The two files necessary for signature verification are included in the resulting compressed artifact for upload.

- **Signing collections**
Applies the same signing workflow than above, where content units are signed using the `pulp ansible repository sign` command line. The signature and certificates are directly stored on the repository side.

- **Verifying signatures**
Verifying signatures requires two steps: 1) validating the integrity of the collection by verifying`MANIFEST.in` file checksums, 2) checking the signature authenticity using the signature and certificate files. As for the current workflow, verification could be done upon install via the `ansible-galaxy collection install` command or for already installed collections with `ansible-galaxy collection verify`.

cc @gerrod3

PR #1390, opened 2023-03-07
Implement Sigstore signing and verification of collections in Ansible repositories

mayaCostantini:
_Opening as draft to get feedback on a first implementation. Any questions and improvement ideas are welcome_

Related to #1301

This pull request implements the first steps towards an integration of Sigstore to Pulp / Galaxy.
Here is an overview of the features introduced in this pull request and of the next steps proposed to allow Pulp / Galaxy users to sign and verify Ansible content from their repository. The implementation proposed tries to follow a similar logic to the one already in place for signing with Signing Services.

### **Sigstore signing services**

Sigstore signing services are objects registered in Pulp's database that allow users to save a specific Sigstore configuration to sign artifacts. They can be created and removed from the `pulpcore-manager` command-line or when logged into Pulp as an administrator. The goal of a Sigstore signing service is to allow admins to save the following Sigstore signing and verification parameters:

- A name acting as a unique identifier for the service
- An environment to enable Pulp to request OIDC credentials from a cloud provider if the server is installed in a GCP or AWS environment
- The URL of the Rekor (transparency log) instance the content should be signed against (defaults to the Rekor public good instance)
- A PEM-encoded root public key for the Rekor instance.
- The URL of the Fulcio (certificate authority) instance that should generate X509 signing certificates (defaults to the Fulcio public good instance)
- The URL of the [TUF](https://dlorenc.medium.com/using-the-update-framework-in-sigstore-dc393cfe6b52) metadata repository instance to use.
- The URL of the OpenID Connect issuer to use for signing.
- The expected identity provider to find in the signing certificate to verify.
- A path to the OIDC client ID and client secret file on the server
- A PEM-encoded public key for the Fulcio Certificate Transparency log.
- The OIDC identity that should be present in the X509 signing certificate.
- An option to perform signature verification offline
- An option to write a single Sigstore bundle file to the collection.
- An option to set Keycloak as the OIDC issuer.
- An option to disable Sigstore's interactive browser flow.

When creating a Sigstore signing service from the command line, those options can either be passed as command line arguments or via a Sigstore signing service configuration file with fields as specified in the documentation.
See corresponding implementation in https://github.com/pulp/pulpcore/pull/3643

### **Workflow**

Upon creation of an Ansible repository, a previously created Sigstore signing service can be attached to it (as it is currently the case for GPG signing services).
When uploading a collection version sigstore signature object, the artifact signature will be automatically verified on upload, and the upload and association between the signature materials and corresponding collection version will fail if the signature could not be verified.

It is possible to sign collections in a repository with the `sigstore_sign` action or with the `pulp` command line.

### **Signing standard for Ansible collections**

The signing standards for Ansible collections and projects is documented in this pending [`ansible-sign PR`](https://github.com/ansible/ansible-sign/pull/33) that adds sigstore signing and verification capabilities to the `ansible-sign` CLI.
The signing steps for a collection (or project) are the following:
- Users specify a list of files to consider as the collection content in a `MANIFEST.in` file at the root
- `ansible-sign` generates a `sha256sum.txt` file under the `.ansible-sign/` directory containing all the checksums of the files specified in `MANIFEST.in`
- Sigstore signs `sha256sum.txt` and outputs signature materials under `.ansible-sign/`: a signature file `sha256sum.txt.sig`, a signing certificate `sha256sum.txt.crt` and eventually a Sigstore bundle `sha256sum.txt.sigstore` used for offline verification.

Those signature materials can be attached to a `CollectionVersion` and will be verified by the Sigstore signing service attached to the repository where the `CollectionVersionSigstoreSignature` gets uploaded.

### **Features and improvements to come**

 - [ ] Support for private instances of Rekor, Fulcio and OIDC provider:
     - [ ] More extensive testing with Keycloak as an identity provider and a a private Sigstore deployment.
     - [ ] Determine the most secure OAuth / OIDC flow in Keycloak that would allow users to sign content without any interaction (web browser prompt to log in with an identity provider)
     - [ ] Evaluate if OIDC client secrets can be stored securely on the server in the case of a non-interactive signing flow
 - [ ] Support for self-managed key pair signing
 - [ ] Tests
 - [ ] (To discuss) Support for ambient credentials detection in cloud environments
 - [ ] (To discuss) Associate right to sign with an identity / Keycloak service account with a Pulp account via RBAC


pulp_ansible/app/models.py
```diff
@@ -266,6 +298,312 @@ class Meta:
         unique_together = ("pubkey_fingerprint", "signed_collection")
 
 
+class SigstoreSigningService(Content):
```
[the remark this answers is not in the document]
mdellweg: This is most probably not a type of Content. Can you look into the signing services concept in Pulp?
I'm wondering how much of the vast amount of configuration can be handled on the shellscript level a signing service calls into. Also please sepearate the concept of signing from the concept of verifying.
Am i seeing this right that Sigstore creates x509 instead of gpg signatures? If that is correct, all the verification part should know of is whether a x509 cert is valid for some root cert.

I may be missing a point, but if think the signing service should be a x509-signing-service and using sigstore, openssl, a hardware cryptodevice or whatever cert authority is merely an implementation detail of the script called into.

pulp_ansible/app/models.py
```diff
@@ -266,6 +298,312 @@ class Meta:
...
+            identity=self.cert_identity,
+            issuer=self.expected_identity_provider,
+        )
+        return self.verifier.verify(materials=verification_materials, policy=policy)
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
+
+
+class CollectionVersionSigstoreSignature(Content):
```
mdellweg, line 555: Is there a useful way to have only one CollectionVerisionSignature Content type?

@rochacbruno I think this involves some questions around how the fact that a collection is signed is represented in galaxy.

pulp_ansible/app/tasks/signature.py
```diff
@@ -72,6 +91,71 @@ def verify_signature_upload(data):
     return data
 
 
+def verify_sigstore_signature_upload(data):
+    """The task code for verifying Sigstore signature upload."""
+    collection = data["signed_collection"]
+    repository = data.get("repository")
+    sigstore_signing_service = repository and repository.sigstore_signing_service
```
mdellweg, line 98: This is the point, where when you look at the existing verify code, there is no notion of a signing service at all. An installation that is not configured to sign anything still needs to be able to verify.

Also can we have the one existing verify task determine the type of signature and handle accordingly?

pulp_ansible/app/tasks/signature.py
```diff
@@ -91,13 +175,36 @@ def sign(repository_href, content_hrefs, signing_service_href):
     SigningDeclarativeVersion(first_stage, repository).create()
 
 
+def sigstore_sign(repository_href, content_hrefs, sigstore_signing_service_href):
```
mdellweg, line 178: Also here, why adding another singing task. The existing one should just create signatures based on the type of configured SigningService. In an ideal world there would not even be any distinction here. `singing_service.sign()` would just return the desired type of signature.
Or maybe a `sign_collection(cv, signing_service)` function for that matter.

mdellweg:
Thinking more about this (I am working off of the assumption that SigStore is no more than a certificate authority with a special api to get certificates) i have a challenge for you:
Find a way to write the feature such that 'git grep -i sigstore' comes out blank. I believe juggling x905 certificates is the proper level of abstraction here.

pulp_ansible/app/models.py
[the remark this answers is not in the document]
mayaCostantini: > Do you think the generic parts can be moved to pulpcore?

@ipanova The configuration part in SigstoreSigningService (model fields) could eventually be reused by other plugins as they are only relative to which Sigstore instances to connect to, which kind of verification to perform etc. The sigstore_sign and sigstore_verify methods could also be reused as AFAIK, the verification materials do not differ for different artifact types.

> I'm wondering how much of the vast amount of configuration can be handled on the shellscript level a signing service calls into.

@mdellweg The idea behind Sigstore signing services is to offer a native integration of Sigstore using the sigstore-python library directly into Pulp instead of calling the cosign command line from a shell script as it is the case for GPG signing as of now. As Sigstore signing and verification can get pretty complex and necessitates among other things to call the different Sigstore instance APIs and handle a larger number of exceptions than GPG, it is probably better to have some control over the process.

> Am i seeing this right that Sigstore creates x509 instead of gpg signatures? If that is correct, all the verification part should know of is whether a x509 cert is valid for some root cert.

Sigstore signing process generates two (plus one optional) outputs: a signature file and an X509 signing certificate (eventually a "Sigstore bundle" file used for offline verification). The certificate is short-lived (valid for 10min) and the verification process involves verifying the certificate signature by a certificate chain linking back to Sigstore CA (Fulcio).

> I may be missing a point, but if think the signing service should be a x509-signing-service and using sigstore, openssl, a hardware cryptodevice or whatever cert authority is merely an implementation detail of the script called into.

I am not quite sure what such an implementation would look like.
The CA used by Sigstore has [its own specification for issuing signing certificates](https://docs.sigstore.dev/fulcio/certificate-issuing-overview), and the client for verifying the certificate claims, which is not compatible with certificate signing with openssl or any other certificate authority as far as I can tell.

pulp_ansible/app/tasks/signature.py
mayaCostantini: > An installation that is not configured to sign anything still needs to be able to verify.

I agree this would be a better approach. Repositories could be configured with a separate verifying service, the only thing is that users should be aware that some of the parameters passed to the signing service used to sign collections (or to the Sigstore client used if signing was done locally) should be taken into account when creating the verifying service (for example, they should both use the same URLs for the Rekor and TUF instances and the identity and OIDC issuers used to sign should be the ones checked in the certificate by the verifying service).

> Also can we have the one existing verify task determine the type of signature and handle accordingly?

Yes, we could write both verification tasks in the same method and use a parameter to differentiate between which type of verification should be used.

mayaCostantini:
@ipanova @mdellweg Would you like to schedule a call to discuss about this feature and have a feedback session? I could also present you how Sigstore works and have a demo if you are interested.  Any Pulp team member would be welcome to attend of course (cc @goosemania).

mdellweg:
Now I'm even more confused. If the certificate is valid for no longer than 10 minutes, what is the purpose of generating it in the first place, let alone saving it in Pulp for later use?
All i see is a thing (as in black box) to which I say make me a signature for this artifact. What i do not see, is why such a signature (that only comes in a different format) should be handled vastly different than a gpg signature.
And being way more complicated is all the more reason to keep all the configuration and credentials for talking to the sigstore server away from Pulp's database. That does not mean that "the shell script" cannot be implemented in python.

mayaCostantini:
>  If the certificate is valid for no longer than 10 minutes, what is the purpose of generating it in the first place, let alone saving it in Pulp for later use?

The way Sigstore works is roughly the following:

**Signing an artifact**
- The client generates an ephemeral key pair that stays in memory during the whole signing process
- The signer is redirected to an identity provider (for example, Google, GitHub or Keycloak) and identifies through an OIDC flow
- The client gets backs a signed identity token attesting that the signer is the owner of the identity used to login (email)
- A Certificate Signing Request is built by the client and sent along to Fulcio (Sigstore's Certificate Authority) with the identity token and the ephemeral public key
- Fulcio verifies the identity token and binds the public key to the certificate with the signer's identity and the OIDC issuer URL. This signed short-lived (10min) certificate is sent back to the client
- The client verifies the certificate and signs off the artifact with the private key
- An entry is sent to be logged to Rekor (Sigstore's transparency log) with signature metadata
- The signature process outputs are a signature file and the ephemeral signing certificate

**Verifying an artifact**
- To verify an artifact signature, the client needs the artifact itself, the signature file and the certificate
- The client performs the following verifications:
    - The signature entry timestamp in Rekor is checked to show that the signing happened during the certificate period of validity
    - The OIDC identity and issuer used to sign the artifact (present in the certificate SAN) are the ones expected by the client
    - The signature can be verified with the public key in the certificate

mayaCostantini:
I understand that some steps in the signing and verifying task could be better shared between the different types of signing schemes. However I am not sure to understand how the Sigstore configuration should be stored if it is not in the database. Would that mean users would have to write and maintain a Python script on their server with the signing service configuration?

ipanova:
> @ipanova @mdellweg Would you like to schedule a call to discuss about this feature and have a feedback session? I could also present you how Sigstore works and have a demo if you are interested. Any Pulp team member would be welcome to attend of course (cc @goosemania).

Meeting would be good. This is a big PR, learning more details about Sigstore internals will also be beneficial. Please invite @mdellweg @gerrod3 @lubosmj

mayaCostantini:
Thanks @ipanova, I sent out a meeting invitation for next week.

pulp_ansible/app/models.py
rochacbruno: @mdellweg @mayaCostantini 

Right now galaxy shows a `signed` badge and allows filtering only signed collections. 

To determine that state it only counts the number of signatures in the listed repo for the specified collection.

There are plans to show more information about the signature in the future but none implemented yet

Issue #1391, opened 2023-03-09
semantic_version says pre-release versions can be higher than release versions


jctanner:
**Version**
main

**Describe the bug**
semantic_version says "1.0.1-rc1" is greater than "1.0.1", which is perhaps true per the spec but I think we collectively think otherwise.

**To Reproduce**
```from semantic_version import Version

versions = ['1.0.0', '2.0.0', '3.0.0-rc.1']
versions = [Version(x) for x in versions]
print(max(versions))
```

**Expected behavior**
Pre-release versions should only get marked as is_highest when no stable releases exist.

mdellweg:
I did some experiment, and i cannot confirm the original claim that
> "1.0.1-rc1" is greater than "1.0.1"

I can however see, that in case we have "2.0.0" and "3.0.0-rc.1" we do not want to consider the latter as the `highest_(stable_)version`. Can we not search the highest stable version and fall back to the highest of all if there was none?

```
In [1]: from semver import compare

In [2]: from semantic_version import Version

In [3]: compare("1.2.3-rc", "1.2.2")
Out[3]: 1

In [4]: compare("1.2.3-rc", "1.2.3")
Out[4]: -1

In [5]: Version("1.2.3-rc") > Version("1.2.2")
Out[5]: True

In [6]: Version("1.2.3-rc") > Version("1.2.3")
Out[6]: False

In [7]: Version("1.2.3-rc.1") > Version("1.2.3")
Out[7]: False

In [8]: Version("1.2.3-rc.1") > Version("1.2.2")
Out[8]: True
```

PR #1392, opened 2023-03-09
Do not use pre-release versions as is_highest unless no stable version exists.

jctanner:
fixes: #1391

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -430,10 +430,32 @@ def _update_highest_version(collection_version):
     equals False on the last highest version and True on this version.
     Otherwise does nothing.
     """
+
     last_highest = collection_version.collection.versions.filter(is_highest=True).first()
     if not last_highest:
-        collection_version.is_highest = True
+        # we only have one version, so mark it as the highest
+        if collection_version.collection.versions.count() == 1:
+            collection_version.is_highest = True
+            collection_version.save()
+            return None
+
+        # compute highest from the whole list ...
+        highest = None
+        for cv in collection_version.collection.versions.all():
+            sv = Version(cv.version)
+            if highest is None or (highest[0] < sv and not sv.prerelease):
+                highest = (sv, cv)
+        highest[1].is_highest = True
+        highest[1].save()
         return None
+
+    # don't use newer pre-releases if we have a stable version as is_highest
+    if (
+        Version(collection_version.version).prerelease
+        and not Version(last_highest.version).prerelease
+    ):
+        return None
+
     if Version(collection_version.version) > Version(last_highest.version):
         last_highest.is_highest = False
         collection_version.is_highest = True
--- a/CHANGES/1391.bugfix
+++ b/CHANGES/1391.bugfix
@@ -1,0 +1,1 @@
+Pre-release collection versions should only be higher than stable releases when none exist.
--- a/pulp_ansible/tests/unit/test_tasks.py
+++ b/pulp_ansible/tests/unit/test_tasks.py
@@ -7,6 +7,8 @@ from unittest import mock
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import TestCase
 
+from orionutils.generator import randstr
+
 from pulp_ansible.app.models import AnsibleDistribution
 from pulp_ansible.app.models import AnsibleRepository
 from pulp_ansible.app.models import Collection
@@ -14,8 +16,11 @@ from pulp_ansible.app.models import CollectionVersion
 from pulpcore.plugin.models import Artifact
 from pulpcore.plugin.models import ContentArtifact
 
-from pulp_ansible.app.tasks.collections import _rebuild_collection_version_meta
-from pulp_ansible.app.tasks.collections import rebuild_repository_collection_versions_metadata
+from pulp_ansible.app.tasks.collections import (
+    _rebuild_collection_version_meta,
+    rebuild_repository_collection_versions_metadata,
+    _update_highest_version,
+)
 
 
 def make_cv_tarball(namespace, name, version):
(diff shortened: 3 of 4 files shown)


newswangerd (approved):

pulp_ansible/app/tasks/collections.py
```diff
@@ -430,10 +430,32 @@ def _update_highest_version(collection_version):
...
+            collection_version.is_highest = True
+            collection_version.save()
+            return None
+
+        # compute highest from the whole list ...
+        highest = None
+        for cv in collection_version.collection.versions.all():
+            sv = Version(cv.version)
+            if highest is None or (highest[0] < sv and not sv.prerelease):
+                highest = (sv, cv)
```
gerrod3, line 447: I think this is wrong when all the versions are pre-releases. The first pre-release will be set to `highest`, but none of the following ones will be properly checked even if they are higher.

Sidenote, I hope this portion of the method won't get called often (hopefully never), as it could be quite a performance hit during a sync if we called this part for each CV.

gerrod3 (changes requested):

pulp_ansible/app/tasks/collections.py
jctanner: Yep, you were right. I fixed it with a helper function to make the comparison and added a unit test for the scenario.

pulp_ansible/app/tasks/collections.py
```diff
@@ -431,7 +431,17 @@ def _update_highest_version(collection_version):
     Otherwise does nothing.
     """
 
+    def is_higher(a, b):
+        """Is a > b ?"""
+        if a.prerelease and b.prerelease and a > b:
+            return True
+        if a.prerelease and not b.prerelease:
+            return False
```
gerrod3, line 439: Don't you also need to account for the scenario when `a` is not a prerelease, but `b` is?

pulp_ansible/app/tasks/collections.py
```diff
@@ -430,19 +430,54 @@ def _update_highest_version(collection_version):
...
     Otherwise does nothing.
     """
+
+    def is_higher(a, b):
+        """Is a > b ?"""
+        if a.prerelease and b.prerelease and a > b:
+            return True
+        if a.prerelease and not b.prerelease:
+            return False
+        return a > b
```
mdellweg, line 440:
```suggestion
        if a.prerelease == b.prerelease
            return a > b
        return b.prerelease
```
I think this should coverall cases.
Also, i don't see the function being used (yet).

jctanner: Your function doesn't pass this test:

```
    def test_prerelease_not_higher_than_stable(self):
        namespace = randstr()
        name = randstr()
        specs = [
            (namespace, name, "1.0.1"),
            (namespace, name, "2.0.0-rc1"),
        ]
    
        collection_versions = build_cvs_from_specs(specs)
        for cv in collection_versions:
            _update_highest_version(cv, save=True)
    
>       assert (
            CollectionVersion.objects.filter(namespace=namespace, name=name, version="1.0.1")
            .first()
            .is_highest
            is True
        )
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -447,25 +447,25 @@ def is_higher(a, b):
...
-        return None
+        return
 
     # don't use newer pre-releases if we have a stable version as is_highest
     if (
         Version(collection_version.version).prerelease
         and not Version(last_highest.version).prerelease
     ):
-        return None
+        return
```
mdellweg: Is this not a good place for another `is_new_highest`?

pulp_ansible/app/tasks/collections.py
```diff
@@ -430,19 +430,56 @@ def _update_highest_version(collection_version):
...
+    ):
+        return
+
     if Version(collection_version.version) > Version(last_highest.version):
         last_highest.is_highest = False
         collection_version.is_highest = True
         last_highest.save()
-        collection_version.save()
+        if save:
+            collection_version.save()
```
mdellweg, line 477: Was this for debugging reasons?

jctanner:
```
<jtanner>
think i found a bug in _update_highest_version
https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/tasks/collections.py#L434-L436
if you were to somehow make a couple CVs, that function hits the "if not last_highest" block. It sets is_highest, but doesn't save()

<gerrod> jtanner: `save` is called after that method is called. https://github.com/pulp/pulp_ansible/blob/8ec918ca4399539f599ad0514f7907cb66d36f28/pulp_ansible/app/tasks/collections.py#L347-L349

<jtanner> gerrod: k, but i think that doesn't allow for testing isolation.

<gerrod> hmm, true. You could just mimic the behavior in the tests and save after calling the method. If you find a way to refactor it without double saving that will be fine as well
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -421,7 +421,7 @@ def _get_backend_storage_url(artifact_file):
     return url
 
 
-def _update_highest_version(collection_version):
+def _update_highest_version(collection_version, save=False):
```
mdellweg, line 424: I fail to see the use for `save=False`. Don't we always want to update when we go to the lengths of running this function at all?

mdellweg (commented):
Just one last question before i think we are ready to merge.

pulp_ansible/app/tasks/collections.py
jctanner: I addressed this in a previous review comment. @gerrod3 wants to retain the behavior where the function caller does the saving when possible to reduce duplicate saving.

pulp_ansible/app/tasks/collections.py
```diff
@@ -430,18 +430,41 @@ def _update_highest_version(collection_version):
...
-        return None
-    if Version(collection_version.version) > Version(last_highest.version):
-        last_highest.is_highest = False
-        collection_version.is_highest = True
-        last_highest.save()
-        collection_version.save()
+        # we only have one version, so mark it as the highest
+        if collection_version.collection.versions.count() == 1:
+            collection_version.is_highest = True
+            collection_version.save()
```
gerrod3, line 446: This should also perform the `if save` check.

pulp_ansible/app/tasks/collections.py
```diff
@@ -430,18 +430,43 @@ def _update_highest_version(collection_version):
...
+        # compute highest from the whole list ...
+        highest = None
+        for cv in collection_version.collection.versions.all():
+            sv = Version(cv.version)
+            if highest is None or is_new_highest(sv, highest[0]):
+                highest = (sv, cv)
+
+        highest[1].is_highest = True
+        if save:
+            highest[1].save()
```
gerrod3, line 459:
```suggestion
        if highest[1] != collection_version or save:
            highest[1].save(update_fields=["is_highest"])
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -430,18 +430,43 @@ def _update_highest_version(collection_version):
...
+    if not is_new_highest(Version(collection_version.version), Version(last_highest.version)):
+        return
 
-    elif collection_version.is_highest and collection_version.version != last_highest.version:
-        collection_version.is_highest = False
+    last_highest.is_highest = False
+    collection_version.is_highest = True
+    last_highest.save()
+    if save:
         collection_version.save()
```
gerrod3, line 470:
```suggestion
    last_highest.is_highest = False
    last_highest.save(update_fields=["is_highest"])
    collection_version.is_highest = True
    if save:
        collection_version.save(update_fields=["is_highest"])
```

gerrod3 (changes requested):
Last changes required (I think). Makes it a bit more performant. https://docs.djangoproject.com/en/3.2/ref/models/instances/#specifying-which-fields-to-save

pulp_ansible/app/tasks/collections.py
mdellweg: Thank you, so we are actually preventing a superfluous save call.

mdellweg (approved):

gerrod3 (approved):

PR #1393, opened 2023-03-10
Add API endpoint for copying and moving content between repos.

newswangerd:
Allows for: `POST {repo_href}/copy_collection_version/` and `POST {repo_href}/copy_collection_version/` which allows for copying and moving collections between repos.

Payload:

```
{
  "collection_versions": ["href1", "href2"],       Collection version to move/copy,
  "destination_repositories": ["href1", "href2"],  List of repositories to copy/move the collection into
  "signing_service": "href"                        Optional. Sign the collections with the given signing service before copy/moving  
}
```

### DIFF
--- a/CHANGES/1394.feature
+++ b/CHANGES/1394.feature
@@ -1,0 +1,2 @@
+Create POST {repo_href}/copy_collection_version/ and POST {repo_href}/copy_collection_version/
+API endpoints which allow for copying and moving collections between repos.
--- a/CHANGES/1394.feature
+++ b/CHANGES/1394.feature
@@ -1,2 +1,3 @@
-Create POST {repo_href}/copy_collection_version/ and POST {repo_href}/copy_collection_version/
-API endpoints which allow for copying and moving collections between repos.
+Created POST ``{repo_href}/copy_collection_version/`` and POST
+``{repo_href}/copy_collection_version/`` API endpoints which
+allow for copying and moving collections between repos.
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -866,7 +866,7 @@ class AnsibleNamespaceMetadataSerializer(NoArtifactContentSerializer):
         )
 
 
-class CollectionVersionMarkSerializer(ModelSerializer):
+class CollectionVersionMarkSerializer(NoArtifactContentSerializer):
     """
     A serializer for mark models.
     """
@@ -883,7 +883,7 @@ class CollectionVersionMarkSerializer(ModelSerializer):
 
     class Meta:
         model = CollectionVersionMark
-        fields = ["marked_collection", "value"]
+        fields = ["pulp_created", "pulp_href", "marked_collection", "value"]
 
 
 class AnsibleRepositoryMarkSerializer(serializers.Serializer):
@@ -961,6 +961,38 @@ class CollectionImportDetailSerializer(CollectionImportListSerializer):
 
     class Meta(CollectionImportListSerializer.Meta):
         fields = CollectionImportListSerializer.Meta.fields + ("error", "messages")
+
+
+class CollectionVersionCopyMoveSerializer(serializers.Serializer):
+    """
+    Copy or move collections from a source repository into one or mor destinations.
+    """
+
+    collection_versions = DetailRelatedField(
+        required=True,
+        view_name_pattern=r"content(-.*/.*)-detail",
+        queryset=CollectionVersion.objects.all(),
+        help_text=_("A list of collection versions to move or copy."),
+        many=True,
+    )
+
+    destination_repositories = DetailRelatedField(
+        required=True,
+        view_name="repositories-ansible/ansible-detail",
+        queryset=AnsibleRepository.objects.all(),
+        help_text=_("List of repository HREFs to put content in."),
+        many=True,
+    )
+
+    signing_service = RelatedField(
+        required=False,
+        view_name="signing-services-detail",
+        queryset=SigningService.objects.all(),
+        help_text=_(
+            "HREF for a signing service. This will be used to sign the collection "
+            "before moving putting it in any new repositories."
+        ),
+    )
 
 
 class CopySerializer(serializers.Serializer):
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -965,12 +965,14 @@ class CollectionImportDetailSerializer(CollectionImportListSerializer):
 
 class CollectionVersionCopyMoveSerializer(serializers.Serializer):
     """
-    Copy or move collections from a source repository into one or mor destinations.
+    Copy or move collections from a source repository into one or more destinations.
+
+    This will carry associated content like Signatures and Marks along.
     """
 
     collection_versions = DetailRelatedField(
         required=True,
-        view_name_pattern=r"content(-.*/.*)-detail",
+        view_name=r"content-ansible/collection_versions-detail",
         queryset=CollectionVersion.objects.all(),
         help_text=_("A list of collection versions to move or copy."),
         many=True,
--- a/pulp_ansible/app/tasks/copy.py
+++ b/pulp_ansible/app/tasks/copy.py
@@ -140,14 +140,15 @@ def move_collection(cv_pk_list, src_repo_pk, dest_repo_list):
 
 
 def copy_or_move_and_sign(copy_or_move, signing_service_pk=None, **move_task_kwargs):
-    if signing_service_pk:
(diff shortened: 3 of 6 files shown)


CHANGES/1394.feature
```diff
@@ -0,0 +1,2 @@
+Create POST {repo_href}/copy_collection_version/ and POST {repo_href}/copy_collection_version/
```
mdellweg, line 1:
```suggestion
Created ``POST {repo_href}/copy_collection_version/`` and ``POST {repo_href}/copy_collection_version/``
```
We use past tense. Also please keep in mind that these will (currently) be rendered with RST.

Did you intend to name one of them differently?

pulp_ansible/app/serializers.py
```diff
@@ -963,6 +963,38 @@ class Meta(CollectionImportListSerializer.Meta):
         fields = CollectionImportListSerializer.Meta.fields + ("error", "messages")
 
 
+class CollectionVersionCopyMoveSerializer(serializers.Serializer):
+    """
+    Copy or move collections from a source repository into one or mor destinations.
```
mdellweg, line 968:
```suggestion
    Copy or move collections from a source repository into one or more destinations. This will carry associated content like Signatures and Marks along.
```

pulp_ansible/app/serializers.py
```diff
@@ -963,6 +963,38 @@ class Meta(CollectionImportListSerializer.Meta):
...
+    """
+
+    collection_versions = DetailRelatedField(
+        required=True,
+        view_name_pattern=r"content(-.*/.*)-detail",
```
mdellweg, line 973: We could probably use `view_name` here too.

pulp_ansible/app/tasks/copy.py
```diff
@@ -62,3 +73,80 @@ def process_entry(entry):
         base_version = dest_repo_version if dest_version_provided else None
         with dest_repo.new_version(base_version=base_version) as new_version:
             new_version.add_content(content_to_copy)
+
+
+def copy_collection(cv_pk_list, src_repo_pk, dest_repo_list):
+    """
+    Copy a collection and all of it's contents into a list of destination repositories
```
mdellweg, line 80:
```suggestion
    Copy a collection version and all of it's related contents into a list of destination repositories.
```

pulp_ansible/app/tasks/copy.py
```diff
@@ -62,3 +73,80 @@ def process_entry(entry):
...
+
+
+def move_collection(cv_pk_list, src_repo_pk, dest_repo_list):
+    """
+    Copy a collection and all of it's contents into a list of destination repositories
+    """
+    copy_collection(cv_pk_list, src_repo_pk, dest_repo_list)
+
+    # don't need to clean anything other than the collection version because everything
+    # else will get handled by AnsibleRepository.finalize_repo_version()
```
mdellweg, line 137:
```suggestion
    # No need to clean anything other than the collection version because everything
    # else will get handled by AnsibleRepository.finalize_repo_version().
```

pulp_ansible/app/tasks/copy.py
```diff
@@ -62,3 +73,80 @@ def process_entry(entry):
...
+            if deprecations_pks:
+                content.extend(deprecations_pks)
+
+    for pk in dest_repo_list:
+        add_and_remove(repository_pk=pk, add_content_units=content, remove_content_units=[])
+
+
+def move_collection(cv_pk_list, src_repo_pk, dest_repo_list):
+    """
+    Copy a collection and all of it's contents into a list of destination repositories
```
mdellweg, line 132:
```suggestion
    Copy a collection version and all of it's associated contents into a list of destination repositories.
```

pulp_ansible/app/tasks/copy.py
```diff
@@ -62,3 +73,80 @@ def process_entry(entry):
...
+    add_and_remove(repository_pk=src_repo_pk, add_content_units=[], remove_content_units=cv_pk_list)
+
+
+def copy_or_move_and_sign(copy_or_move, signing_service_pk=None, **move_task_kwargs):
+    if signing_service_pk:
+        sign(
+            signing_service_href=signing_service_pk,
+            repository_href=move_task_kwargs["src_repo_pk"],
+            content_hrefs=move_task_kwargs["cv_pk_list"],
+        )
```
mdellweg, line 147: This will sign the things in the original Repository, and not purely "on the fly", right?
This strikes me as a rather non-intuitive side-effect. I might be convinced otherwise.

pulp_ansible/app/viewsets.py
```diff
@@ -436,6 +437,57 @@ def sign(self, request, pk):
...
+        serializer.is_valid(raise_exception=True)
+        data = serializer.validated_data
+
+        dest_repos_pks = [x.pk for x in data["destination_repositories"]]
+
+        signing_service = data.get("signing_service", None)
+        if signing_service:
+            signing_service = signing_service.pk
+
+        reserved = data["destination_repositories"] + [repository]
```
mdellweg, line 452: Is the signing the reason to make the source repository exclusive?

pulp_ansible/app/viewsets.py
```diff
@@ -436,6 +437,57 @@ def sign(self, request, pk):
...
+        return OperationPostponedResponse(result, request)
+
+    @extend_schema(
+        description="Trigger an asynchronous task to copy collection versions.",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=True, methods=["post"], serializer_class=CollectionVersionCopyMoveSerializer)
+    def copy_collection_version(self, request, pk):
+        """
+        Copy a collection and all of its associated content from this repository.
```
mdellweg, line 474:
```suggestion
        Copy a collection version and all of its associated content from this repository.
```

pulp_ansible/app/viewsets.py
```diff
@@ -436,6 +437,57 @@ def sign(self, request, pk):
...
+        return self._handle_copy_or_move(request, "copy")
+
+    @extend_schema(
+        description="Trigger an asynchronous task to move collection versions.",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=True, methods=["post"], serializer_class=CollectionVersionCopyMoveSerializer)
+    def move_collection_version(self, request, pk):
+        """
+        Move a collection and all of its associated content from this repository.
```
mdellweg, line 486:
```suggestion
        Move a collection version and all of its associated content from this repository.
```

pulp_ansible/app/viewsets.py
newswangerd: Yes. We could check the serializer and modify that if no signing service is found

pulp_ansible/app/tasks/copy.py
newswangerd: I didn't want the overhead of calling sign on each dest repository. If there's a way to sign the collections without specifying a repo, and then add the signatures as a separate operation, I can do that.

pulp_ansible/app/serializers.py
newswangerd: @mdellweg what's the right value to use for viewname? I couldn't figure it out, so I finally gave up

mdellweg: That is a good question. I guess something like "content-ansible-collection_version-detail".

pulp_ansible/app/tasks/copy.py
mdellweg: And we def don't want different signatures on all dest repos. Would it be wise to see if there was a signature already? Or is the `sign` call taking care of that?
Yes, i'd prefer keeping the source repo untouched. You'd probably need to rewrite a `sign` method that instead of adding to a repo returns the resulting pk(s).

gerrod3: `sign` will grab the signatures already present in Pulp instead of resigning, so calling sign on each dest repo is fine. The first repo will perform the actual signing and everyone after will add the generated signatures. The only annoying thing will be all the extra progress report messages that will be present in the task report.

newswangerd: @gerrod3 @mdellweg I updated it to call sign on each dest repo. LMK if you want me to leave it like this, or change it.

mdellweg: This will create a whole lot of new repository versions, no?

mdellweg (approved):

pulp_ansible/app/tasks/copy.py
mdellweg: I don't see this applied.

PR #1395, opened 2023-03-20
Add namespace compatibility with galaxy_ng

newswangerd:
Make collection sync work with the quick and dirty galaxy ng namespace implementation.

Requires: https://github.com/ansible/galaxy_ng/pull/1642


pulp_ansible/app/migrations/0049_ansiblenamespacemetadata_avatar_url.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('ansible', '0048_collectionversionmark'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='ansiblenamespacemetadata',
+            name='avatar_url',
+            field=models.URLField(blank=True, max_length=256),
+        ),
```
gerrod3, line 17: We haven't released Namespace Metadata yet. I think we can just add this to the original migration. It should only break dev envs.

newswangerd: 👍

pulp_ansible/app/tasks/collections.py
```diff
@@ -674,21 +674,46 @@ async def _add_namespace(self, name):
...
-                        relative_path=f"{name}-avatar",
-                        deferred_download=False,
-                    )
-                ]
-                if url
-                else None
-            )
+            links = namespace.get("links", None)
+
+            # clean up the galaxy API for pulp
```
gerrod3, line 679: The special code for handling `avatar_url` and `avatar_sha256` doesn't really make sense to me. This method `_add_namespace` only gets called if the collection version metadata has `metadata_sha256` in it's `namespace` field (see line 658), so syncing against current galaxy_ng wouldn't trigger this. If you want namespace sync to work against a remote that doesn't have the pulp_ansible NamespaceMetadata code, then you are going to have to add some more hacks.

newswangerd: I added the `metadata_sha256` to the namespace API in galaxy (https://github.com/ansible/galaxy_ng/pull/1642/files#diff-7f24d0266c5878d82a98acb7d91612b64915a8209671edd13c57478283d119a0R53).

 Unfortunately we DO get `metadata_sha256` on the V3 collection apis because I migrated the existing ns metadata as part of https://github.com/pulp/pulp_ansible/pull/1395. I could remove the migration, but we'll only get NS metadata on the search api if there's a metadata object for it.

I thought about trying to provide the avatar_sha on galaxy_ng, but it's too complicated. I'd have to figure out how to download all of the images that are currently only provided as URLs. This approach seemed a lot more feasible, and as a bonus we get logo sync.

pulp_ansible/app/tasks/collections.py
```diff
@@ -674,21 +674,46 @@ async def _add_namespace(self, name):
...
+                await logo.download()
+
+                # If the remote doesn't support avatar_sha256, save the avatar_url to the
+                # database. This will cause the namespace metadata has to be calculated using
+                # the url instead of the sha, which will allow it to be computed correctly for
+                # syncs against galaxy_ng
+                if avatar_sha256:
+                    namespace.pop("avatar_url", None)
+                else:
+                    namespace["avatar_sha256"] = logo.artifact.sha256
```
gerrod3, line 711: Do we care if the upstream `avatar_url` changes the avatar it is pointing to? I'm pretty sure in that scenario we wouldn't properly update the namespace. The sync might even break.

newswangerd: @gerrod3 do you mean the image the avatar_url points to changes?

newswangerd: If that's what you mean, it won't download the new image if something else hasn't been changed on the namespace. I don't think it will break, because when the avatar_url is set on the ns model, that gets used for the metadata sha

gerrod3: Yes, in this case NamespaceMetadata would break on a resync because the metadata_sha256 wouldn't change (the url didn't change), but the avatar sha256 did. The NamespaceMetadata wouldn't be updated since the primary key didn't change, but since we downloaded the new avatar and put it into a declaritive artifact, the sync pipeline will update the artifact for the NamespaceMetadata. Then there will be a mismatch for the avatarsha256 and the actual avatar being stored/served. 

Two options (that I see): 

1. Don't download the logo when `avatar_url` is used. Have Pulp always redirect to original link. Pros - clients that fetch the avatar will always get the most up to date one. Cons - no longer storing avatar locally.
2. Include both `avatar_url` and `avatar_sha256` in `metadata_sha256` calculation. This would desync the metadatasha256 from the remote and Pulp, but Pulp would now properly create new AnsibleMetadata when the avatar image changes.

newswangerd: looks like you're right

pulp_ansible/app/tasks/collections.py
```diff
@@ -699,21 +711,38 @@ async def _add_namespace(self, name):
...
                         deferred_download=False,
                     )
-                ]
-                if url
-                else None
-            )
+
+                    await logo.download()
+                else:
+                    logo = None
```
gerrod3, line 742: This doesn't make sense. The pipeline already takes care of checking if the artifact exists based on the `avatar_sha256` and figuring out if it should be downloaded. Since you are not using the artifact immediately now, revert this back to the original code.

gerrod3 (changes requested):

gerrod3 (approved):
@newswangerd Looks good, leaving it to you to merge.

PR #1428, opened 2023-04-18
Implement Sigstore signing and verification of collections in Ansible repositories (refactored)

mayaCostantini:
Opening this PR as a refactored version of #1390

TODO (@mayaCostantini ):
- [x] Update documentation
- [ ] Tests
- [ ] CI
- [ ] Document discussion from #1390


pulp_ansible/app/models.py
```diff
@@ -275,6 +318,381 @@ class Meta:
         unique_together = ("pubkey_fingerprint", "signed_collection")
 
 
+
+class SigstoreSigningService(Content):
```
mdellweg, line 322: This should not be content.

pulp_ansible/app/models.py
```diff
@@ -275,6 +318,381 @@ class Meta:
...
+    PUBLIC_TUF_URL = "https://sigstore-tuf-root.storage.googleapis.com/"
+    PUBLIC_ISSUER_URL = "https://oauth2.sigstore.dev/auth"
+
+    name = models.CharField(db_index=True, unique=True, max_length=64)
+    rekor_url = models.TextField(default=PUBLIC_REKOR_URL)
+    rekor_root_pubkey = models.TextField(null=True)
+    fulcio_url = models.TextField(default=PUBLIC_FULCIO_URL)
+    tuf_url = models.TextField(default=PUBLIC_TUF_URL)
+    oidc_issuer = models.TextField(default=PUBLIC_ISSUER_URL)
+    credentials_file_path = models.TextField(null=True)
```
mayaCostantini, line 376: Important note: this field is just here temporarily to test the signing workflow by retrieving the OIDC credentials from a file stored in plaintext on the server. This is not the approach we will be going for; I will replace it once an alternative flow to get credentials has been validated.

pulp_ansible/app/models.py
mayaCostantini: What would be a better model to subclass?

mdellweg: SigningService or BaseModel

mdellweg:
Can you please elaborate on who is signing what (ansible collections, i guess), and what it means for a content to have a valid signature?

pulp_ansible/app/viewsets.py
```diff
@@ -790,6 +887,47 @@ def sign(self, request, pk):
...
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(
+        detail=True, methods=["post"], serializer_class=AnsibleRepositorySigstoreSignatureSerializer
+    )
+    def sigstore_sign(self, request, pk):
+        """
+        Dispatches a Sigstore signing task.
+        This endpoint is in tech preview and can change at any time in the future.
+        """
```
gerrod3, line 901: Would it be possible to move this into the existing `sign` endpoint? From what I've seen it's not that different call argument wised to what we have already. If you turn the `SigstoreSigningService` to be of type `SigningService` then you can use similar logic like we have in `sync` and use the type of the signing service to determine which task to call.

requirements.txt
```diff
@@ -6,3 +6,6 @@ Pillow>=7.0,<9.6
 pulpcore>=3.23.0,<3.25
 PyYAML>=5.4.1,<7.0
 semantic_version>=2.9,<2.11
+setuptools>=39.2,<66.2.0
+sigstore==1.1.1
+voluptuous==0.13.1
```
gerrod3, line 11: Is `voluptuous` still being used? Also what is `setuptools` being used for? I would like to keep that out the requirements.

gerrod3 (changes requested):

pulp_ansible/app/models.py
```diff
@@ -499,6 +907,19 @@ class AnsibleRepository(Repository, AutoAddObjPermsMixin):
     gpgkey = models.TextField(null=True)
     private = models.BooleanField(default=False)
 
+    sigstore_signing_service = models.ForeignKey(
+        SigstoreSigningService,
+        on_delete=models.SET_NULL,
+        related_name="ansible_repositories",
+        null=True,
+    )
```
gerrod3, line 915:
```suggestion
```
Unless you merge the two models together, the repository should only have a reference to the verifier.

pulp_ansible/app/models.py
```diff
@@ -499,6 +907,19 @@ class AnsibleRepository(Repository, AutoAddObjPermsMixin):
...
+    sigstore_verifying_service = models.ForeignKey(
+        SigstoreVerifyingService,
+        on_delete=models.SET_NULL,
+        related_name="ansible_repositories",
+        null=True,
+    )
```
gerrod3, line 921: I wonder if this should be a many-to-many relationship. An interesting feature of the `gpgkey` field is that the text can store multiple gpg-keys to verify against, and only one has to pass. Limiting this field to one verifying service heavily restricts the source of signatures that can be uploaded.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,295 @@
...
+    artifact_file = storage.open(artifact_name)
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        differ = DistlibManifestChecksumFileExistenceDiffer
+        checksum = ChecksumFile(tempdir, differ=differ)
+        with tarfile.open(fileobj=artifact_file, mode="r") as tar:
+            tar.extractall(path=tempdir)
+
+            try:
+                manifest = checksum.generate_gnu_style()
```
gerrod3, line 67: This seems potentially really slow. We already have the checksum file `MANIFEST.IN` in the tarball, can we not just extract it and pass it along instead of re-calculating the whole thing?

pulp_ansible/app/serializers.py
```diff
@@ -752,6 +906,64 @@ class Meta:
...
+        """Ensure `file` field is required."""
+        super().__init__(*args, **kwargs)
+        self.fields["file"].required = True
+
+    def validate(self, data):
+        """
+        Verify the signature is valid before creating it.
+        """
+        data = super().validate(data)
+
```
gerrod3, line 948: Looking at `verify_sigstore_signature_upload` below I think we should add a check here to make sure the `repository` field is set and if not raise a `serializer.ValidationError` so that we error quicker.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,295 @@
...
+        delimiter = "-----"
+        s = pubkey.split(delimiter)
+        res = delimiter + s[1] + delimiter + re.sub(" +", "\n", s[2]) + delimiter + s[3] + delimiter
+        return res
+
+    certificate = format(certificate)
+    sigstore_bundle = data.get("sigstore_bundle")
+    checksums = _generate_checksum_manifest(collection)
+
+    bundle = Bundle().from_json(sigstore_bundle) if sigstore_bundle != "null" else None
```
gerrod3, line 115: Weird that a non-existant `sigstore_bundle` gets passed in as 'null' instead of `None` or simply not being present in `data`. Seems like a problem with the serializer field.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,295 @@
...
+
+    with tempfile.NamedTemporaryFile(dir=".", delete=False, mode="w") as manifest_file:
+        manifest_file.write(checksums)
+        manifest_file.flush()
+    with open(manifest_file.name, "rb", buffering=0) as manifest_io:
```
gerrod3, line 120: Since checksums is already in-memory no need to re-write it to a file, use io.BytesIO instead: https://docs.python.org/3/library/io.html#binary-i-o

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,295 @@
...
+            "Failed to verify Sigstore signature for collection "
+            f"{collection}: {verification_result.reason}\n"
+            f"Exception: {verification_result.exception}"
+        )
+
+    log.info(f"Validated Sigstore signature for collection {collection}")
+
+    data["data"] = signature
+    data["sigstore_x509_certificate"] = certificate
+    data["sigstore_signing_service"] = sigstore_signing_service
```
gerrod3, line 146:
```suggestion
```
Uploaded signatures don't have a signing service.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,295 @@
...
+                        collection_version
+                    )
+                    await manifest_file.write(manifest_data)
+                async with aiofiles.open(manifest_file.name, mode="rb", buffering=0) as iofile:
+                    async with aiofiles.tempfile.NamedTemporaryFile(
+                        dir=".", mode="w", delete=False
+                    ) as manifest_content:
+                        content = await iofile.read()
+                        manifest_content.write(content)
+                    with open(manifest_content.name, "rb") as manifest_bytes:
```
gerrod3, line 245: Same two things as above with the verify code:
1. Can we just extract the `MANIFEST.IN` file instead of regenerating it?
2. Once we have the file in memory let's use io.BytesIO instead of rewriting it to a temp file.

pulp_ansible/app/tasks/signature.py
```diff
@@ -97,7 +97,7 @@ class SigningDeclarativeVersion(DeclarativeVersion):
     def pipeline_stages(self, new_version):
         """The stages for the signing process."""
         pipeline = [
-            self.first_stage,  # CollectionSigningFirstStage
+            self.first_stage,  # CollectionSigningFirstStage or CollectionSigstoreSigningFirstStage
```
gerrod3, line 100:
```suggestion
            self.first_stage,  # CollectionSigningFirstStage
```

pulp_ansible/app/sigstore/issuers/keycloak.py
```diff
@@ -0,0 +1,135 @@
...
+        # https://github.com/sigstore/sigstore-python/blob/v1.1.0/sigstore/oidc.py#L100
+        # This method works only on browser interaction and enables out-of-bond
+        # only if the former method fails.
+        if enable_interactive:
+            from sigstore._internal.oidc.oauth import _OAuthFlow
+
+            code: str
+            with _OAuthFlow(client_id, client_secret, self) as server:
+                # Launch web browser
+                if webbrowser.open(server.base_uri):
```
gerrod3, line 64: This code simply won't work for the signing task as 99% of the time the machine the task is running on is not accessible by the user, so the signing task will simply stall until it fails waiting for the user input. 

I think I have a good idea for how to get around this problem, but it's pretty in-depth and will take me a while to write down. I would also like to get feedback from my other team-members, so let's discuss it at our next sync.

gerrod3 (changes requested):
Sorry for the late review, lots of things still to discuss.

pulp_ansible/app/models.py
```diff
@@ -275,6 +313,368 @@ class Meta:
...
+            Enable Sigstore's interactive browser flow.
+            Defaults to False.
+    """
+
+    TYPE = "sigstore_signing_service"
+
+    PUBLIC_OIDC_ISSUERS = (
+        "https://accounts.google.com",
+        "https://oauth2.sigstore.dev/auth",
+        "http://dex-idp:8888/auth",
```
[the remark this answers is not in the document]
mayaCostantini: Yes, it is stated in the official [Fulcio config](https://github.com/sigstore/fulcio/blob/b428307d9b9b4186041538c92d64350004c45448/config/config.jsn#L14), probably for a locally deployed Dex server.

pulp_ansible/app/models.py
```diff
@@ -275,6 +313,368 @@ class Meta:
...
+    )
+    PUBLIC_REKOR_URL = "https://rekor.sigstore.dev"
+    PUBLIC_FULCIO_URL = "https://fulcio.sigstore.dev"
+    PUBLIC_TUF_URL = "https://sigstore-tuf-root.storage.googleapis.com/"
+    PUBLIC_ISSUER_URL = "https://oauth2.sigstore.dev/auth"
+
+    name = models.CharField(db_index=True, unique=True, max_length=64)
+    rekor_url = models.TextField(default=PUBLIC_REKOR_URL)
+    rekor_root_pubkey = models.TextField(null=True)
```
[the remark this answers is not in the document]
mayaCostantini: I didn't put a default here because the privileged way of retrieving Sigstore instances public keys should be via TUF, the public key fields for Rekor and the CTLog being here just in case users need to override key retrieval by TUF.

pulp_ansible/app/models.py
```diff
@@ -275,6 +313,368 @@ class Meta:
...
+        if self.ctfe_pubkey:
+            return bytes(self.ctfe_pubkey, "utf-8")
+        return self.trust_updater.get_ctfe_keys()
+
+    @property
+    def rekor_public_keys(self):
+        """Get the Rekor instance public key."""
+        if self.rekor_root_pubkey:
+            return bytes(self.rekor_root_pubkey, "utf-8")
+        return self.trust_updater.get_rekor_keys()
```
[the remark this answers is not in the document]
mayaCostantini: Right, fixed both the type (list of `bytes`) and checking if one of the TUF url or Rekor public key is set in https://github.com/pulp/pulp_ansible/pull/1428/commits/6edf4e9cafae04c20271525269b15a32a8f153a2 and https://github.com/pulp/pulp_ansible/pull/1428/commits/9d3bdb02450781466bb968995b7fc5edfe7f587b for models and `pulpcore-manager` commands.

pulp_ansible/app/models.py
```diff
@@ -275,6 +313,368 @@ class Meta:
...
+
+    @property
+    def rekor(self):
+        """Get a Rekor instance."""
+        if self.rekor_url == self.PUBLIC_REKOR_URL:
+            return RekorClient.production(self.trust_updater)
+
+        rekor_key = RekorKeyring(Keyring([self.rekor_public_keys]))
+        ctfe_key = CTKeyring(Keyring([self.ctfe_public_keys]))
+        return RekorClient(
```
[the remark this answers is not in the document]
mayaCostantini: See comment above

pulp_ansible/app/models.py
```diff
@@ -275,6 +313,368 @@ class Meta:
...
+
+        log.info(f"Using ephemeral certificate: {result.cert_pem}")
+        log.info(f"Transparency log entry created at index: {result.log_entry.log_index}")
+        signing_result["signature"] = result.b64_signature
+        signing_result["certificate"] = result.cert_pem
+        signing_result["bundle"] = result._to_bundle().to_json()
+
+        return signing_result
+
+    async def sigstore_asign(self, input_digest, private_key, cert):
```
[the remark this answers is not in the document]
mayaCostantini: Some operations can be executed asynchronously inside of `Signer.sign`, like signing artifacts and sending entries to Rekor. Actually, I am not sure if the `sign` method (non-async) above would be needed at all, so we can remove it if that's ok.

pulp_ansible/app/models.py
```diff
@@ -275,6 +313,368 @@ class Meta:
...
+            log_entry=log_entry,
+        )
+
+        signing_result["signature"] = result.b64_signature
+        signing_result["certificate"] = result.cert_pem
+        signing_result["bundle"] = result._to_bundle().to_json()
+
+        return signing_result
+
+    def save(self, *args, **kwargs):
```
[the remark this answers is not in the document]
mayaCostantini: Addressed in https://github.com/pulp/pulp_ansible/pull/1428/commits/39bffb737fc5ce900f7a4c69fbd7f285c63822d3

pulp_ansible/app/models.py
```diff
@@ -275,6 +313,368 @@ class Meta:
...
+            self.rekor_root_pubkey = validate_and_format_pem_public_key(self.rekor_root_pubkey)
+        if self.ctfe_pubkey:
+            self.ctfe_pubkey = validate_and_format_pem_public_key(self.ctfe_pubkey)
+        super(SigstoreSigningService, self).save(*args, **kwargs)
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
+
+
+class SigstoreVerifyingService(BaseModel):
```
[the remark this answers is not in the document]
mayaCostantini: I remember this was the approach I went with in my first version of this PR, but during a discussion we decided it would be better to separate signing and verifying capabilities for repositories. Is it still the case?

pulp_ansible/app/models.py
mayaCostantini: Yes, I agree this would be a better approach. Is there a specific way to define many-to-many relationships in Pulp or is a `ManyToManyField` enough?

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
mayaCostantini: That could be done, but initially I went with this approach to conserve compatibility with [`ansible-sign`](https://github.com/ansible/ansible-sign/blob/9e28a8ad65d89fceebc5d6a2a1ebb3e37f142512/src/ansible_sign/cli.py#L161) as it will also be used to sign Ansible artifacts with Sigstore in the future.

pulp_ansible/app/serializers.py
mayaCostantini: Addressed in https://github.com/pulp/pulp_ansible/pull/1428/commits/de1e8e1f031badfe015d5b421ba5e474631bf41c

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
mayaCostantini: Actually given the new standard adopted in sigstore-python (only the bundle is outputted by default as it already contains all the materials necessary for verifying the signature), I was thinking about removing the signature and certificate files from the `CollectionVersionSigstoreSignature` to simplify it. We can talk about it more in detail during our meeting if that's ok.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
mayaCostantini: Fixed in https://github.com/pulp/pulp_ansible/pull/1428/commits/df18426680d102d00a3859e749fa9b075d422f99

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
mayaCostantini: I think they actually do:

```
{
            "pulp_href": "/pulp/api/v3/content/ansible/content/68aeb271-691b-4910-baa8-bce859929b50/",
            "pulp_created": "2023-06-14T13:06:47.645521Z",
            "data": "MGYCMQD8J17K9G0aqKZTBPBJvL97P27fDVPp1HT/wHgxvF+h4HrxIwPYsYQG9E0aYKTzHmcCMQD3ixhAUsKdrUZTsQLEySOHVKrD61CvOeCkkk02Kgj6e+NlW99mJyCjLuoKb5v28ds=",
            "sigstore_x509_certificate": "-----BEGIN CERTIFICATE-----\nMIIDsTCCA1igAwIBAgIUUqFobSm73ExZkB6R2FBe8I/hxTkwCgYIKoZIzj0EAwIw\ngZcxCzAJBgNVBAYTAlVTMQswCQYDVQQIDAJNQTEPMA0GA1UEBwwGQm9zdG9uMRAw\nDgYDVQQKDAdSZWQgSGF0MRAwDgYDVQQLDAdSZWQgSGF0MRYwFAYDVQQDDA1vcGVu\nc2hpZnRhcHBzMS4wLAYJKoZIhvcNAQkBFh9vcGVuc2hpZnRhcHBzQG9wZW5zaGlm\ndGFwcHMuY29tMB4XDTIzMDYxNDEzMDY0NloXDTIzMDYxNDEzMTY0NlowADB2MBAG\nByqGSM49AgEGBSuBBAAiA2IABJk/8FHb0PA7kU5/zbbuQoFsAd1TJj+FWB1PyGtA\nVTBK4MvuXgcOKKQcI0GLeadac0HdvfZjsNitEhjobnH8+dDqvHsHOM9w+R2aZL2K\nB8dAeD9eiw5Jpa7Sov8noA94M6OCAfkwggH1MA4GA1UdDwEB/wQEAwIHgDATBgNV\nHSUEDDAKBggrBgEFBQcDAzAdBgNVHQ4EFgQUHk6ltOJgcKWkUGFpn545shIW8LIw\nHwYDVR0jBBgwFoAUwXoXJ20AjXcfuZ0Q+MfzGM5QU1cwLAYDVR0RAQH/BCIwIIEe\nbWF5YS5jb3N0YW50aW5pQHByb3Rvbm1haWwuY29tMGgGCisGAQQBg78wAQEEWmh0\ndHBzOi8va2V5Y2xvYWsta2V5Y2xvYWsuYXBwcy5vcGVuLXN2Yy1zdHMuazF3bC5w\nMS5vcGVuc2hpZnRhcHBzLmNvbS9hdXRoL3JlYWxtcy9zaWdzdG9yZTBqBgorBgEE\nAYO/MAEIBFwMWmh0dHBzOi8va2V5Y2xvYWsta2V5Y2xvYWsuYXBwcy5vcGVuLXN2\nYy1zdHMuazF3bC5wMS5vcGVuc2hpZnRhcHBzLmNvbS9hdXRoL3JlYWxtcy9zaWdz\ndG9yZTCBiQYKKwYBBAHWeQIEAgR7BHkAdwB1APlNklTuzJCFaZy/xxsoT1OgbYHg\nht07hNnZZSKIPp7GAAABiLoECfUAAAQDAEYwRAIgNuwoktic/YuF8VCSKnlhgZM4\nfxzvWtveNKTFoFaOqJ0CID6wdRqeoYJTNfgABjJyraiSRQGaa4GAf39pRNsOF3yg\nMAoGCCqGSM49BAMCA0cAMEQCIEgamVx2WHUSb7v562wGHA+4DHh+1B/6HawbIIox\n3XfgAiAG3+taNA4j8oJIiNWDNrkMbHl4eWy3YZrWDQN2SO8lTw==\n-----END CERTIFICATE-----\n",
            "sigstore_bundle": "{\"mediaType\": \"application/vnd.dev.sigstore.bundle+json;version=0.1\", \"verificationMaterial\": {\"x509CertificateChain\": {\"certificates\": [{\"rawBytes\": \"MIIDsTCCA1igAwIBAgIUUqFobSm73ExZkB6R2FBe8I/hxTkwCgYIKoZIzj0EAwIwgZcxCzAJBgNVBAYTAlVTMQswCQYDVQQIDAJNQTEPMA0GA1UEBwwGQm9zdG9uMRAwDgYDVQQKDAdSZWQgSGF0MRAwDgYDVQQLDAdSZWQgSGF0MRYwFAYDVQQDDA1vcGVuc2hpZnRhcHBzMS4wLAYJKoZIhvcNAQkBFh9vcGVuc2hpZnRhcHBzQG9wZW5zaGlmdGFwcHMuY29tMB4XDTIzMDYxNDEzMDY0NloXDTIzMDYxNDEzMTY0NlowADB2MBAGByqGSM49AgEGBSuBBAAiA2IABJk/8FHb0PA7kU5/zbbuQoFsAd1TJj+FWB1PyGtAVTBK4MvuXgcOKKQcI0GLeadac0HdvfZjsNitEhjobnH8+dDqvHsHOM9w+R2aZL2KB8dAeD9eiw5Jpa7Sov8noA94M6OCAfkwggH1MA4GA1UdDwEB/wQEAwIHgDATBgNVHSUEDDAKBggrBgEFBQcDAzAdBgNVHQ4EFgQUHk6ltOJgcKWkUGFpn545shIW8LIwHwYDVR0jBBgwFoAUwXoXJ20AjXcfuZ0Q+MfzGM5QU1cwLAYDVR0RAQH/BCIwIIEebWF5YS5jb3N0YW50aW5pQHByb3Rvbm1haWwuY29tMGgGCisGAQQBg78wAQEEWmh0dHBzOi8va2V5Y2xvYWsta2V5Y2xvYWsuYXBwcy5vcGVuLXN2Yy1zdHMuazF3bC5wMS5vcGVuc2hpZnRhcHBzLmNvbS9hdXRoL3JlYWxtcy9zaWdzdG9yZTBqBgorBgEEAYO/MAEIBFwMWmh0dHBzOi8va2V5Y2xvYWsta2V5Y2xvYWsuYXBwcy5vcGVuLXN2Yy1zdHMuazF3bC5wMS5vcGVuc2hpZnRhcHBzLmNvbS9hdXRoL3JlYWxtcy9zaWdzdG9yZTCBiQYKKwYBBAHWeQIEAgR7BHkAdwB1APlNklTuzJCFaZy/xxsoT1OgbYHght07hNnZZSKIPp7GAAABiLoECfUAAAQDAEYwRAIgNuwoktic/YuF8VCSKnlhgZM4fxzvWtveNKTFoFaOqJ0CID6wdRqeoYJTNfgABjJyraiSRQGaa4GAf39pRNsOF3ygMAoGCCqGSM49BAMCA0cAMEQCIEgamVx2WHUSb7v562wGHA+4DHh+1B/6HawbIIox3XfgAiAG3+taNA4j8oJIiNWDNrkMbHl4eWy3YZrWDQN2SO8lTw==\"}]}, \"tlogEntries\": [{\"logIndex\": \"349\", \"logId\": {\"keyId\": \"IFajaKdLy+sgq4+XSC5BH0KbPN8Kpcq7tYo5QMdCBpQ=\"}, \"kindVersion\": {\"kind\": \"hashedrekord\", \"version\": \"0.0.1\"}, \"integratedTime\": \"1686748007\", \"inclusionPromise\": {\"signedEntryTimestamp\": \"MEYCIQCVdYM7DhTo4ZQguWbXQD+R73UynZFtadNYjRATyBdLWwIhANPrpndVfmIf55c5Bn+dlCBGgO/IbEu8RWnxnK48jnd8\"}, \"inclusionProof\": {\"logIndex\": \"349\", \"rootHash\": \"pfsRcGU1+YTDFU2FtCoc7Ig8iv7nxypYzVn35KlIiSQ=\", \"treeSize\": \"350\", \"hashes\": [\"2Ln+aSDCopdrk9aY+DfGrwgKmkbWZuxb34NX4nCSQIg=\", \"JNyCbro9UJ9Yfhv55KCJ+vQQgv8029bretUa0ZAWFws=\", \"i+cLXIsbD4zB+OnN4JqNSuSajScXnV+UZ7B3H2o2azQ=\", \"54cKDlSFbb1IGAJLeE3wt6rWuhW4Mo7Dvu3y+UqeCwY=\", \"GUEiTpxawkRVsubS/3/DkEjy3qVFIqxwfafGliKkDTM=\", \"SfXoge+ONFcKFoRk3bbplo7ucOU9hE177NQdZxxvWZs=\"]}, \"canonicalizedBody\": \"eyJhcGlWZXJzaW9uIjoiMC4wLjEiLCJraW5kIjoiaGFzaGVkcmVrb3JkIiwic3BlYyI6eyJkYXRhIjp7Imhhc2giOnsiYWxnb3JpdGhtIjoic2hhMjU2IiwidmFsdWUiOiJlM2IwYzQ0Mjk4ZmMxYzE0OWFmYmY0Yzg5OTZmYjkyNDI3YWU0MWU0NjQ5YjkzNGNhNDk1OTkxYjc4NTJiODU1In19LCJzaWduYXR1cmUiOnsiY29udGVudCI6Ik1HWUNNUUQ4SjE3SzlHMGFxS1pUQlBCSnZMOTdQMjdmRFZQcDFIVC93SGd4dkYraDRIcnhJd1BZc1lRRzlFMGFZS1R6SG1jQ01RRDNpeGhBVXNLZHJVWlRzUUxFeVNPSFZLckQ2MUN2T2VDa2trMDJLZ2o2ZStObFc5OW1KeUNqTHVvS2I1djI4ZHM9IiwicHVibGljS2V5Ijp7ImNvbnRlbnQiOiJMUzB0TFMxQ1JVZEpUaUJEUlZKVVNVWkpRMEZVUlMwdExTMHRDazFKU1VSelZFTkRRVEZwWjBGM1NVSkJaMGxWVlhGR2IySlRiVGN6UlhoYWEwSTJVakpHUW1VNFNTOW9lRlJyZDBObldVbExiMXBKZW1vd1JVRjNTWGNLWjFwamVFTjZRVXBDWjA1V1FrRlpWRUZzVmxSTlVYTjNRMUZaUkZaUlVVbEVRVXBPVVZSRlVFMUJNRWRCTVZWRlFuZDNSMUZ0T1hwa1J6bDFUVkpCZHdwRVoxbEVWbEZSUzBSQlpGTmFWMUZuVTBkR01FMVNRWGRFWjFsRVZsRlJURVJCWkZOYVYxRm5VMGRHTUUxU1dYZEdRVmxFVmxGUlJFUkJNWFpqUjFaMUNtTXlhSEJhYmxKb1kwaENlazFUTkhkTVFWbEtTMjlhU1doMlkwNUJVV3RDUm1nNWRtTkhWblZqTW1od1dtNVNhR05JUW5wUlJ6bDNXbGMxZW1GSGJHMEtaRWRHZDJOSVRYVlpNamwwVFVJMFdFUlVTWHBOUkZsNFRrUkZlazFFV1RCT2JHOVlSRlJKZWsxRVdYaE9SRVY2VFZSWk1FNXNiM2RCUkVJeVRVSkJSd3BDZVhGSFUwMDBPVUZuUlVkQ1UzVkNRa0ZCYVVFeVNVRkNTbXN2T0VaSVlqQlFRVGRyVlRVdmVtSmlkVkZ2Um5OQlpERlVTbW9yUmxkQ01WQjVSM1JCQ2xaVVFrczBUWFoxV0dkalQwdExVV05KTUVkTVpXRmtZV013U0dSMlpscHFjMDVwZEVWb2FtOWlia2c0SzJSRWNYWkljMGhQVFRsM0sxSXlZVnBNTWtzS1FqaGtRV1ZFT1dWcGR6VktjR0UzVTI5Mk9HNXZRVGswVFRaUFEwRm1hM2RuWjBneFRVRTBSMEV4VldSRWQwVkNMM2RSUlVGM1NVaG5SRUZVUW1kT1ZncElVMVZGUkVSQlMwSm5aM0pDWjBWR1FsRmpSRUY2UVdSQ1owNVdTRkUwUlVablVWVklhelpzZEU5S1oyTkxWMnRWUjBad2JqVTBOWE5vU1ZjNFRFbDNDa2gzV1VSV1VqQnFRa0puZDBadlFWVjNXRzlZU2pJd1FXcFlZMloxV2pCUkswMW1la2ROTlZGVk1XTjNURUZaUkZaU01GSkJVVWd2UWtOSmQwbEpSV1VLWWxkR05WbFROV3BpTTA0d1dWYzFNR0ZYTlhCUlNFSjVZak5TZG1KdE1XaGhWM2QxV1RJNWRFMUhaMGREYVhOSFFWRlJRbWMzT0hkQlVVVkZWMjFvTUFwa1NFSjZUMms0ZG1FeVZqVlpNbmgyV1ZkemRHRXlWalZaTW5oMldWZHpkVmxZUW5kamVUVjJZMGRXZFV4WVRqSlplVEY2WkVoTmRXRjZSak5pUXpWM0NrMVROWFpqUjFaMVl6Sm9jRnB1VW1oalNFSjZURzFPZG1KVE9XaGtXRkp2VEROS2JGbFhlSFJqZVRsNllWZGtlbVJIT1hsYVZFSnhRbWR2Y2tKblJVVUtRVmxQTDAxQlJVbENSbmROVjIxb01HUklRbnBQYVRoMllUSldOVmt5ZUhaWlYzTjBZVEpXTlZreWVIWlpWM04xV1ZoQ2QyTjVOWFpqUjFaMVRGaE9NZ3BaZVRGNlpFaE5kV0Y2UmpOaVF6VjNUVk0xZG1OSFZuVmpNbWh3V201U2FHTklRbnBNYlU1MllsTTVhR1JZVW05TU0wcHNXVmQ0ZEdONU9YcGhWMlI2Q21SSE9YbGFWRU5DYVZGWlMwdDNXVUpDUVVoWFpWRkpSVUZuVWpkQ1NHdEJaSGRDTVVGUWJFNXJiRlIxZWtwRFJtRmFlUzk0ZUhOdlZERlBaMkpaU0djS2FIUXdOMmhPYmxwYVUwdEpVSEEzUjBGQlFVSnBURzlGUTJaVlFVRkJVVVJCUlZsM1VrRkpaMDUxZDI5cmRHbGpMMWwxUmpoV1ExTkxibXhvWjFwTk5BcG1lSHAyVjNSMlpVNUxWRVp2Um1GUGNVb3dRMGxFTm5ka1VuRmxiMWxLVkU1bVowRkNha3A1Y21GcFUxSlJSMkZoTkVkQlpqTTVjRkpPYzA5R00zbG5DazFCYjBkRFEzRkhVMDAwT1VKQlRVTkJNR05CVFVWUlEwbEZaMkZ0Vm5neVYwaFZVMkkzZGpVMk1uZEhTRUVyTkVSSWFDc3hRaTgyU0dGM1lrbEpiM2dLTTFobVowRnBRVWN6SzNSaFRrRTBhamh2U2tscFRsZEVUbkpyVFdKSWJEUmxWM2t6V1ZweVYwUlJUakpUVHpoc1ZIYzlQUW90TFMwdExVVk9SQ0JEUlZKVVNVWkpRMEZVUlMwdExTMHRDZz09In19fX0=\"}]}, \"messageSignature\": {\"messageDigest\": {\"algorithm\": \"SHA2_256\", \"digest\": \"47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=\"}, \"signature\": \"MGYCMQD8J17K9G0aqKZTBPBJvL97P27fDVPp1HT/wHgxvF+h4HrxIwPYsYQG9E0aYKTzHmcCMQD3ixhAUsKdrUZTsQLEySOHVKrD61CvOeCkkk02Kgj6e+NlW99mJyCjLuoKb5v28ds=\"}}",
            "signed_collection": "/pulp/api/v3/content/ansible/collection_versions/343fc505-e15b-48b9-a3f7-897532736336/",
            "sigstore_signing_service": "/pulp/api/v3/sigstore-signing-services/5f484e77-f5b0-4916-9680-e3bf45822804/"
},
```

The `sigstore_signing_service` field is a foreign key for `CollectionVersionSigstoreSignature`

pulp_ansible/app/sigstore/issuers/keycloak.py
mayaCostantini: Thanks for your input during the sync, I removed the interactive signing part in https://github.com/pulp/pulp_ansible/pull/1428/commits/ca80190904420239b6fbbc0992ba3df935da91f8

mayaCostantini:
Thanks again for the thorough review @gerrod3 
I tried to answer most of the comments you raised, please let me know of that works for you.

As discussed during the last sync, I will start a refactor based on sigstore-python v2 as the milestone seems not too far from being reached as of now: https://github.com/sigstore/sigstore-python/milestone/4

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
mdellweg: Why should the created signature have a reference to the signing service at all? What's the benefit of that link?
Also you can upload a signature into a system that has no signing service configured at all. So no, they don't have one.

gerrod3: The signatures having a reference to the signing service was originally a Galaxy request to know if they were locally created. Synced and uploaded signatures will not have an attached signing service, but I don't think they expose the upload signature endpoint in galaxy, so it actually determines if the signature was synced or locally created.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
gerrod3: Just use [TextIo](https://docs.python.org/3/library/io.html#text-i-o) instead.

pulp_ansible/app/models.py
gerrod3: `ManyToManyField` is enough IMO, don't think we need any extra fields in this case.

pulp_ansible/app/models.py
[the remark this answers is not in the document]
gerrod3: No it still is the case that they should be separate. I'm interested to see the changes to this model with the sigstore-bundle changes in the new V2.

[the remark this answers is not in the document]
mdellweg: Remember everyone can verify signatures all the time. But signing needs special setup. So there is no merging.
Also if we need something like a verifying service, the crypto system is inherently broken. The signature itself must contain all the information needed to verify it. And the public keys used for verifying are (as the name suggests) public. Please do not make verifying harder than it is.
The signing service is designed that way to allow using secret keys that pulp will never even touch. For verifying, this does not apply.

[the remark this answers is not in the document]
mayaCostantini: I understand your concern, but here you need information about the Rekor instance that was used to log the signature to retrieve an inclusion proof from the transparency log (which is the stronger kind of verification, even if offline verification is also possible), and so you need to keep a number of parameters related to signing available to verifiers, like the Rekor instance URL and public key for instance, hence the need for a "verifying service".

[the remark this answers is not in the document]
mdellweg: The only thing we need to do is tell pulp which entities to trust for a given repository. Again no secret bits involved. This is (at best) the equivalent of a public gpg key. Maybe it needs a separate db-table, because somehow sigstore defines complex as the new simple. But it is not going to reflect the modus operandi of a signing service. So no sevice and no setup via pulpcore-manager. If the user needs the admin to tell their repository which sigs to trust, this make for a very bad user experience.

pulp_ansible/app/models.py
```diff
@@ -291,6 +313,278 @@ class Meta:
...
+        signed_collection (models.ForeignKey):
+            A collection version this signature is relevant to.
+        sigstore_signing_service (models.ForeignKey):
+            The Sigstore Siging Service used for signing the collection version.
+    """
+
+    TYPE = "collection_sigstore_signatures"
+
+    sigstore_bundle = models.JSONField(null=True)
+    sigstore_bundle_sha256_hash = models.CharField(max_length=64, unique=True, db_index=True, null=True)
```
gerrod3, line 566:
```suggestion
    sigstore_bundle_sha256_hash = models.CharField(max_length=64, db_index=True, null=True)
```
`unique_together` is enough. Don't forget to remove the same code from the migration file.

pulp_ansible/app/serializers.py
```diff
@@ -128,6 +139,148 @@ class Meta:
         )
 
 
+class SigstoreSigningServiceSerializer(ModelSerializer):
```
gerrod3, line 142: This can only be created from the command line, correct? If so then I would change all the fields to `read_only=True`.

pulp_ansible/app/serializers.py
```diff
@@ -128,6 +139,148 @@ class Meta:
...
+            "fulcio_url",
+            "tuf_url",
+            "rekor_root_pubkey",
+            "oidc_issuer",
+            "oidc_client_secret",
+            "ctfe_pubkey",
+        )
+
+
+class SigstoreVerifyingServiceSerializer(ModelSerializer):
```
gerrod3, line 222: Same here as above.

pulp_ansible/app/serializers.py
```diff
@@ -648,11 +811,51 @@ class CollectionVersionSerializer(ContentChecksumSerializer, CollectionVersionUp
...
+        artifact = validated_data.get("artifact")
+        repository = validated_data.get("repository")
+        sigstore_verifying_services = repository.sigstore_verifying_service.all()
+
+        cv_instance = super().create(validated_data)
+
+        artifact_name = artifact.file.name
+        artifact_file = storage.open(artifact_name)
+
+        with tempfile.TemporaryDirectory() as tempdir:
```
gerrod3, line 824:
```suggestion
        with tempfile.TemporaryDirectory(dir=".") as tempdir:
```

pulp_ansible/app/serializers.py
```diff
@@ -648,11 +811,51 @@ class CollectionVersionSerializer(ContentChecksumSerializer, CollectionVersionUp
 
     creating = True
 
+    def create(self, validated_data):
+        artifact = validated_data.get("artifact")
+        repository = validated_data.get("repository")
```
gerrod3, line 816: This could technically return `None` which will result in an attribute error on the next line.

pulp_ansible/app/serializers.py
```diff
@@ -648,11 +811,51 @@ class CollectionVersionSerializer(ContentChecksumSerializer, CollectionVersionUp
...
+        artifact_file = storage.open(artifact_name)
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            with tarfile.open(fileobj=artifact_file, mode="r") as tar:
+                tar.extractall(path=tempdir)
+
+                for root, dirs, files in os.walk(tempdir):
+                    if ".ansible-sign" in dirs:
+                        manifest_path = Path(tempdir) / ".ansible-sign" / "sha256sum.txt"
+                        signature_path = Path(tempdir) / ".ansible-sign" / "sha256sum.txt.sigstore"
```
gerrod3, line 831: I take it this means that collection tarballs can have signatures inside now? Does this affect the digest of the collection's manifest file?

pulp_ansible/app/serializers.py
```diff
@@ -648,11 +811,51 @@ class CollectionVersionSerializer(ContentChecksumSerializer, CollectionVersionUp
...
+                        if os.path.exists(manifest_path) and os.path.exists(signature_path):
+                            sigstore_bundle = Bundle().from_json(signature_path.read_bytes())
+
+                            with manifest_path.open("rb", buffering=0) as io:
+
+                                for sigstore_verifying_service in sigstore_verifying_services:
+                                    verification_result = sigstore_verifying_service.sigstore_verify(manifest=io, sigstore_bundle=sigstore_bundle)
```
gerrod3, line 838: If none of the verifications pass, should the collection be rejected?

pulp_ansible/app/serializers.py
```diff
@@ -648,11 +811,51 @@ class CollectionVersionSerializer(ContentChecksumSerializer, CollectionVersionUp
...
 
+    def create(self, validated_data):
+        artifact = validated_data.get("artifact")
+        repository = validated_data.get("repository")
+        sigstore_verifying_services = repository.sigstore_verifying_service.all()
+
+        cv_instance = super().create(validated_data)
+
+        artifact_name = artifact.file.name
+        artifact_file = storage.open(artifact_name)
```
gerrod3, line 822: I really don't like that we create the CV instance from the artifact, doing a bunch of processing on the tarball, only to then again do more processing afterward to figure out of a signature is inside. It would be great if we could do this processing here: https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/tasks/upload.py#L17 and then maybe the final Signature creation in `finish_collection_upload` (although would need to find some way to also get the signature inside the repository as well...)

pulp_ansible/app/serializers.py
```diff
@@ -764,6 +967,57 @@ class Meta:
...
+                "'repository' field not set"
+            )
+        data = super().validate(data)
+
+        if "request" not in self.context:
+            # Validate is called twice, first on the viewset, and second on the create task
+            # data should be set up properly on the second time, when request isn't in context
+            data = verify_sigstore_signature_upload(data)
+
+        return data
```
gerrod3, line 1010: This way of performing secondary validation is outdated now. 
```suggestion
    def deferred_validate(self, data):
        data = super().deferred_validate(data)
        data = verify_sigstore_signature_upload(data)
        return data
```

pulp_ansible/app/serializers.py
```diff
@@ -764,6 +967,57 @@ class Meta:
...
+        self.fields["file"].required = True
+
+    def validate(self, data):
+        """
+        Verify the signature is valid before creating it.
+        """
+        if not "repository" in data:
+            raise serializers.ValidationError(
+                "'repository' field not set"
+            )
```
gerrod3, line 1002: I think you can remove this and just make the `repository` field required in the `init` like the `file` field.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
gerrod3: What is the status of `ansible-sign`? I see it below as a new dependency. Do we still need to have this code or can we call `ansible-sign`'s code instead?

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
gerrod3: What is the status of this now that V2 is out? Can this verification code be dramatically simplified? Can the verification service be removed entirely and be replaced with a text field like GPG?

pulp_ansible/app/viewsets.py
```diff
@@ -271,6 +279,54 @@ class Meta:
...
+        model = CollectionVersionSigstoreSignature
+        fields = {
+            "signed_collection": ["exact"],
+            "sigstore_bundle_sha256_hash": ["exact", "in"],
+            "sigstore_signing_service": ["exact"],
+        }
+
+
+class SigstoreSigningServiceViewSet(
+    NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin, RolesMixin
```
gerrod3, line 305:
```suggestion
    NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin
```

pulp_ansible/app/viewsets.py
```diff
@@ -271,6 +279,54 @@ class Meta:
...
+    """
+
+    endpoint_name = "sigstore-signing-services"
+    queryset = SigstoreSigningService.objects.all()
+    serializer_class = SigstoreSigningServiceSerializer
+    filterset_fields = ["name"]
+
+
+class SigstoreVerifyingServiceViewSet(
+    NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin, RolesMixin
```
gerrod3, line 318:
```suggestion
    NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin
```

requirements.txt
```diff
@@ -1,8 +1,12 @@
 async_lru>=1.0,<2.1
+ansible-sign==0.1.1
 galaxy_importer>=0.4.5,<0.5
 GitPython>=3.1.24,<3.2
 jsonschema>=4.9,<4.20
 Pillow>=7.0,<10.2
 pulpcore>=3.25.0,<3.40
 PyYAML>=5.4.1,<7.0
 semantic_version>=2.9,<2.11
+setuptools>=39.2,<66.2.0
```
gerrod3, line 10: What is `setuptools` needed for?

gerrod3 (changes requested):

pulp_ansible/app/serializers.py
mayaCostantini: Yes, `ansible-sign` will generate signatures under the mentioned path, which should be extracted and verified when a collection version gets uploaded to a repository. As far as I understand, this should not be computed as part of the manifest file so it shouldn't affect its digest.

pulp_ansible/app/serializers.py
mayaCostantini: I intentionally made it so that the collection can get uploaded even without any valid signature, as it seems to me like it should not be a criteria for admitting content. If you think this should be mandated on a repository configured with at least one Sigstore verifying service, let me know.

pulp_ansible/app/serializers.py
mayaCostantini: Yes, this indeed requires extra processing which I don't like either. However I couldn't figure out another place where to find a reference to the upload repository to upload the signature there as well. Do you think we could leave the signature creation process as it is now or do you see any alternative for the collection to be processed just once and still have eventual signatures uploaded to the right repo?

pulp_ansible/app/viewsets.py
mayaCostantini: Regarding `ansible-sign`, I am still waiting for the sigstore code to be merged there (see https://github.com/ansible/ansible-sign/pull/46), but the code should indeed be replaced when this is done. Let me get back to the `ansible-sign` maintainers and I will replace it.

mayaCostantini: The verification service would still be required as policies for verification must be specified with sigstore (i.e. signer's identity, OIDC issuer, etc).

requirements.txt
mayaCostantini: I am not sure anymore why I added it here, so removing it.

mayaCostantini:
Thanks a lot for the review @gerrod3 ! I pushed the changes you requested, let me know if anything else comes to your mind.

pulp_ansible/app/viewsets.py
mdellweg: I think, I said this before. There is no need for a verification service.
The design of the signing service is to protect secrets. Verification must always work without secrets. Otherwise your cryptosystem is broken. No secrets means: Everything lives in the database and is configurable via the api on the very Repository or Remote supposed to protect itself from improperly signed content.

mayaCostantini: There is no need for any secret here, the parameters specified are public and are simple text fields used by the sigstore client to verify the signature, exactly like you would use a public GPG key.

mayaCostantini: I think the term "verification service" might be misused here. It is actually more of a static configuration than a "service".

pulp_ansible/app/serializers.py
gerrod3: I think you can move this process code into `process_collection_artifact` and have the found signature bundle returned as a part of the collection info. Then on the create function (https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/serializers.py#L524-L529) you pop off the signature information and if it exists then you create the signatures, verifying with the repository's signing verifiers. This way we don't double process the artifact.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,238 @@
...
+        identity = self.sigstore_signing_service.issuer.identity_token(
+            "sigstore", client_secret,
+        )
+
+        signing_ctx = self.sigstore_signing_service.signing_context
+
+        with signing_ctx.signer(identity) as signer:
+            # Limits the number of subprocesses spawned/running at one time
+            async with self.semaphore:
+                async for collection_version in collection_versions:
```
gerrod3, line 194:
```suggestion
```

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,238 @@
...
+        new_content = self.content
+        ntotal = await sync_to_async(new_content.count)()
+        msg = _("Signing new CollectionVersions with Sigstore.")
+        async with ProgressReport(message=msg, code="sign.new.signature", total=ntotal) as p:
+            self.progress_report = p
+            await asyncio.create_task(
+                self.sigstore_sign_collection_versions(
+                    sync_to_async_iterable(new_content.iterator())
+                )
+            )
```
gerrod3, line 228:
```suggestion
            async for collection_version in sync_to_async_iterable(new_content.iterator()):
                tasks.append(asyncio.create_task(self.sigstore_sign_collection_version(collection_version)))
            await asyncio.gather(*tasks)
```

gerrod3 (changes requested):
Have some more big refactoring asks. Also, I am curious about how far along `ansible-sign` is and if we can use it in our verification code.

pulp_ansible/app/management/commands/add-sigstore-verifying-service.py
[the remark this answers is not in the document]
mayaCostantini: > it needs to be created/edited through the api by non-admin users.

I am not sure why the verifying service should be configurable by non-admin users, as its role is to filter collection versions that do not have valid signature according to a pre-defined policy, which is critical from a security point of view.
Regarding having the verifying service replaced by a JSON field, I have nothing against it except if this implies any non-admin user can configure it.

To answer your questions:

1. If users use a CLI such as `ansible-galaxy` to pull collections from a repository, they will use the command options to specify the verification policies (for example, OIDC issuer and signer's identity). The same is true for tooling from automated workflows (github actions, etc).
2. It would probably be useful, as private content in an organizations is likely to originate from the same few sources most of the time (however, this is just an assumption).
3. It would be more of a one and done or once-in-a-while action. As mentioned above, content signature sources are not so likely to change on a frequent basis.

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -0,0 +1,238 @@
...
+    async def sigstore_sign_collection_versions(self, collection_versions):
+        """Signs the collection versions with Sigstore."""
+        # Get an OIDC token to authenticate to Fulcio
+        client_secret = self.sigstore_signing_service.oidc_client_secret
+
+        identity = self.sigstore_signing_service.issuer.identity_token(
+            "sigstore", client_secret,
+        )
+
+        signing_ctx = self.sigstore_signing_service.signing_context
```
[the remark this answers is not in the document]
mayaCostantini: Addressed in https://github.com/pulp/pulp_ansible/pull/1428/commits/e0218f0d4d2173edb0872e1086ffe15c9cb80da6

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
mayaCostantini: Addressed by https://github.com/pulp/pulp_ansible/pull/1428/commits/6346d5d9301d210e28dbf4e9ac6f7dfd67e8486b

pulp_ansible/app/management/commands/add-sigstore-verifying-service.py
```diff
@@ -26,7 +26,7 @@
 
 DEFAULTS = {
     "rekor-url": "https://rekor.sigstore.dev",
-    "tuf-url": "https://sigstore-tuf-root.storage.googleapis.com/",
+    "tuf-url": "https://tuf-repo-cdn.sigstore.dev/",
```
gerrod3, line 29: Is this going to be the final url?

pulp_ansible/app/models.py
```diff
@@ -409,6 +409,8 @@ def issuer(self):
     @property
     def trust_updater(self):
         """Get a custom TrustUpdater instance depending on the TUF metadata repository provided."""
+        if self.tuf_url.endswith("/"):
+            self.tuf_url = self.tuf_url[-1]
```
gerrod3, line 413:
```suggestion
        if self.tuf_url.endswith("/"):
            self.tuf_url = self.tuf_url[:-1]
```
You could also remove the 'if' statement with `self.tuf_url = self.tuf_url.rstrip("/")`

pulp_ansible/app/models.py
```diff
@@ -477,42 +479,60 @@ class SigstoreVerifyingService(BaseModel):
...
     verify_offline = models.BooleanField(null=True, default=False)
 
+    @property
+    def get_rekor_url(self):
+        if self.rekor_url.endswith("/"):
+            self.rekor_url = self.rekor_url[:-1]
+        return self.rekor_url
+
+    @property
+    def get_tuf_url(self):
```
gerrod3, line 489: Is this really necessary? You didn't do the same on the SigningService model.

pulp_ansible/app/serializers.py
```diff
@@ -843,11 +847,18 @@ def create(self, validated_data):
...
                                             sigstore_signing_service=None,
                                         )
                                         cv_signature.save()
+                                    
+                                    io.seek(0)
+
+                        elif settings.ANSIBLE_SIGNATURE_REQUIRE_VERIFICATION:
+                            raise VerificationFailureException(
+                                "Failed to verify Sigstore signature against all verifying services configured on this repository with ANSIBLE_SIGNATURE_REQUIRE_VERIFICATION enabled."
+                            )
```
gerrod3, line 861: Not sure this elif statement is on the right indentation. I was expecting it to be a part of the for loop if we don't break out of it  (which I think you need to add since we don't want to save the signature twice if we have a verifier).

pulp_ansible/app/sigstore/tasks/sigstore_signature.py
```diff
@@ -221,11 +205,9 @@ async def run(self):
...
         async with ProgressReport(message=msg, code="sign.new.signature", total=ntotal) as p:
             self.progress_report = p
-            await asyncio.create_task(
-                self.sigstore_sign_collection_versions(
-                    sync_to_async_iterable(new_content.iterator())
-                )
-            )
+            async for collection_version in sync_to_async_iterable(new_content.iterator()):
+                tasks.append(asyncio.create_task(self.sigstore_sign_collection_version(collection_version)))
+                await asyncio.gather(*tasks)
```
gerrod3, line 210:
```suggestion
            await asyncio.gather(*tasks)
```
This needs to be outside the for loop else we would still be doing the tasks serially.

gerrod3 (changes requested):

pulp_ansible/app/management/commands/add-sigstore-verifying-service.py
mayaCostantini: Yes, this is the one currently used for the public Sigstore instance

PR #1417, opened 2023-04-04
Use yaml.safe_load/dump in tests

mdellweg:
[noissue]

PR #1435, opened 2023-05-04
Reduce fetched fields in cv list endpoint to prevent OOMKILL. (#1430)

jctanner:
fixes: #1433

Signed-off-by: James Tanner <tanner.jc@gmail.com>
(cherry picked from commit a52dc98b7056987b127ebdb1331c5bb058d7540b)

PR #1436, opened 2023-05-04
Use yaml.safe_load/dump in tests (#1417)

jctanner:
[noissue]

(cherry picked from commit 382895abb8b6cb6ce1c83c17a3f5b34583bb15a4)


mdellweg (approved):

jctanner:
Well, this PR is now stuck on the issue with ansible core that is fixed in https://github.com/pulp/pulp_ansible/pull/1435

mdellweg:
so we need both in the same PR?

mdellweg:
OTOH, it is failing on test_install_signed_collection Is this really related to the OOM?

mdellweg:
Anyway, i'm going to force merge this now and we can see through fixing from there.

jctanner:
Installing signed collections is what broke in the newest ansible-core releases, which is why 1435 is trying to pin core to an older version.

PR #1438, opened 2023-05-05
[rebased] backport cv perf

jctanner:
fixes: #1433

Signed-off-by: James Tanner <tanner.jc@gmail.com>
(cherry picked from commit a52dc98b7056987b127ebdb1331c5bb058d7540b)

PR #1440, opened 2023-05-09
Fix sphinx intersphinx_mapping config. (#1439)

jctanner:
New value comes from https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#configuration

[noissue]

(cherry picked from commit ee877c0a92f2ba9e3a69dd2ad515afd9e3a4d3fc)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -173,4 +173,4 @@ texinfo_documents = [
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/': None}
+intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}


jctanner:
https://github.com/pulp/pulp_ansible/pull/1439

gerrod3:
> test failures are due to https://github.com/pulp/pulp_ansible/pull/1438/files#diff-93394e9ecb9acc4a48dffa8515bf0d6c45d2170dd7fd2dcc76d1bd722d94fa0fR1

So the issue is circular? I can always force merge it if we know that to be the case.

jctanner:
I guess it's "circular" in a sense. The ansible-core pin should have been done in a separate PR and then backported directly. However, with the sphinx issue on top of that, we're just in a bad state right now.

mdellweg (approved):

PR #1476, opened 2023-06-02
Optimize v3 collection API querysets.

newswangerd:
### Changes
This PR updates the v3 collection API querysets so that they are all executed as either one queryset or one queryset plus a set of preteches.

Content is queried using a new, optimized function that queries the Content table directly instead of querying RerpositoryContent and then performing a `pk__in` filter on the Content table, reducing the number and size of queries.

The CollectionVersion model has been updated with individual fields for the version number components so that CollectionVersions can be efficiently sorted by their semver version using .order_by() on the queryset.

The collection version list view and collection version detail view responses are now cached by repository version pk.

#### Summary
~8x improvements in number of simultaneous clients and 24x improvement in average client response time

#### Tests
Tested using X number of ansible-galaxy clients simultaneously pulling `ansible.posix`

Current pulp ansible main branch: 128 simultaneous downloads before any client timeouts
```
-------------------------------------------------------------------------------------------
running with 128 total clients against http://localhost:5001/pulp_ansible/galaxy/test/api/
-------------------------------------------------------------------------------------------
failures: []
failure percent: 0.0
avg duration: 41.084137023437506
```

This pr: 1024 simultaneous downloads before any clients timeout
```
-------------------------------------------------------------------------------------------
running with 128 total clients against http://localhost:5001/pulp_ansible/galaxy/test/api/
-------------------------------------------------------------------------------------------
failures: []
failure percent: 0.0
avg duration: 1.6661786093750004

--------------------------------------------------------------------------------------------
running with 1024 total clients against http://localhost:5001/pulp_ansible/galaxy/test/api/
--------------------------------------------------------------------------------------------
failures: []
failure percent: 0.0
avg duration: 5.763926267578129
```

Fixes: https://issues.redhat.com/browse/AAH-2346


pulp_ansible/app/utils.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db.models import OuterRef, Exists
+
+
+# TODO: if this turns out to be a more efficient way to query repo version
+# content, then this should be contributed to pulpcore
+def filter_content_for_repo_version(qs, repo_version):
+    """
+    A more efficient way to query content in a repository version.
+
+    Returns a the given queryset, but filtered to only include content
```
jctanner, line 11: Returns a ... ?

pulp_ansible/app/utils.py
```diff
@@ -0,0 +1,55 @@
...
+    511,           0.007927591800689698, 0.010698480606079102, 1.349524657052642
+    1023,          0.01072232961654663,  0.014668509960174561, 1.3680338587556766
+    2047,          0.016402604579925536, 0.024496512413024904, 1.4934525973396422
+    4095,          0.025542356967926026, 0.0399853777885437,   1.56545372217427
+    8191,          0.046061162948608396, 0.07391655683517456,  1.6047479504077031
+    16383,         0.08123507976531982,  0.15503086805343627,  1.9084226728318017
+    32767,         0.2031115198135376,   0.35589320182800294,  1.7522058923822907
+    """
+
+    return qs.filter(pk__in=repo_version._content_relationships().values_list("content_id"))
```
gerrod3, line 55: Here are the different SQL statements for the old and new way:
OLD:
```
SELECT 
  "core_content"."pulp_id", 
  "core_content"."pulp_created", 
  "core_content"."pulp_last_updated", 
  "core_content"."pulp_type", 
  "core_content"."upstream_id", 
  "core_content"."timestamp_of_interest", 
  "core_content"."pulp_domain_id" 
FROM 
  "core_content" 
  INNER JOIN "core_repositorycontent" ON (
    "core_content"."pulp_id" = "core_repositorycontent"."content_id"
  ) 
WHERE 
  "core_repositorycontent"."pulp_id" IN (
    SELECT 
      U0."pulp_id" 
    FROM 
      "core_repositorycontent" U0 
      INNER JOIN "core_repositoryversion" U2 ON (
        U0."version_added_id" = U2."pulp_id"
      ) 
      LEFT OUTER JOIN "core_repositoryversion" U3 ON (
        U0."version_removed_id" = U3."pulp_id"
      ) 
    WHERE 
      (
        U0."repository_id" = 018806d4 - 1f79 - 7400 - bada - d72b8c775712 
        AND U2."number" <= 1 
        AND NOT (
          U3."number" <= 1 
          AND U3."number" IS NOT NULL
        )
      )
  )
```
NEW:
```
SELECT 
  "core_content"."pulp_id", 
  "core_content"."pulp_created", 
  "core_content"."pulp_last_updated", 
  "core_content"."pulp_type", 
  "core_content"."upstream_id", 
  "core_content"."timestamp_of_interest", 
  "core_content"."pulp_domain_id" 
FROM 
  "core_content" 
WHERE 
  "core_content"."pulp_id" IN (
    SELECT 
      U0."content_id" 
    FROM 
      "core_repositorycontent" U0 
      INNER JOIN "core_repositoryversion" U2 ON (
        U0."version_added_id" = U2."pulp_id"
      ) 
      LEFT OUTER JOIN "core_repositoryversion" U3 ON (
        U0."version_removed_id" = U3."pulp_id"
      ) 
    WHERE 
      (
        U0."repository_id" = 018806d4 - 1f79 - 7400 - bada - d72b8c775712 
        AND U2."number" <= 1 
        AND NOT (
          U3."number" <= 1 
          AND U3."number" IS NOT NULL
        )
      )
  )
```
The speed up seems to be coming from losing an inner join that isn't needed. Interestingly, on my current work optimizing the content-app I have found that postgres does not handle subqueries with RepositoryContent very well, which are still present in this new optimized query. [0] I wonder if there is a way to make the subquery more performant or remove it altogether. 

[0] https://github.com/pulp/pulpcore/pull/3803#discussion_r1187380670

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -256,108 +296,48 @@ def get_queryset(self):
...
-        )
 
-        return collections.distinct("versions__collection_id").only(
-            "pulp_created", "name", "namespace"
+        deprecated_qs = filter_content_for_repo_version(
+            AnsibleCollectionDeprecated.objects, repo_version
+        ).filter(namespace=OuterRef("namespace"), name=OuterRef("name"))
+
+        latest_cv_version_qs = (
+            filter_content_for_repo_version(CollectionVersion.objects, repo_version)
```
gerrod3, line 305: Have you try using `distinct("namespace", "name")` on this query after the `order_by` to see if it speeds it up?

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -868,23 +850,56 @@ def list(self, request, *args, **kwargs):
         """
         Returns paginated CollectionVersions list.
         """
+
+        # there are too many potential permutations of query params to be able
+        # to cache all of them, so this will only cache queries that stick to
+        # limit/offset params, which the ansible-galaxy client uses
```
gerrod3, line 856: What are all the endpoints that the ansible-client uses?

pulp_ansible/app/utils.py
newswangerd: @gerrod3 i found that the RepositoryContent table completely collapses when you get to 65k content. Basic queries like qs[0:100] take 3+ seconds to execute. It seems to do with the joins on version added and version removed. I did a test by adding the version numbers directly to the RepositoryContent table and it lowered the query time to 100ms, which is still pretty bad, but not unmanageable.

pulp_ansible/app/galaxy/v3/views.py
newswangerd: - `index/<namespace>/<name>/versions/`
- `index/<namespace>/<name>/versions/<version>/`
- `artifacts/collections/<path>/<filename>/`

pulp_ansible/app/utils.py
newswangerd: Tanner and I are thinking about moving these viewsets to using the collection version index instead of the RepositoryContent table until pulpcore can improve the schema or the performance of the queries on this table.

gerrod3: That sounds like a good idea.

gerrod3: I wonder if we could do the two joins manually if it would speed it up. i.e.
```
versions = repo.versions.values_list("pk", "number")
versions_present = [pk for pk, number in versions if number <= repo_ver.number]
rc = RepositoryContent.objects.filter(repository=repo.pk, version_added__in=versions_present).exclude(version_removed__in=versions_present)
return qs.filter(pk__in=rc.values_list("content_id"))

Comment (newswangerd) [reply pulp_ansible/app/utils.py:55]: I really don't like these queries that can potentially insert thousands of PKs into a sql query. I'll give this a shot though

Comment (newswangerd) [reply pulp_ansible/app/utils.py:55]: @gerrod3 you're a genius.

```
def filter_content_for_repo_version(qs, repo_version):
    repo_version_qs = RepositoryVersion.objects.filter(repository=repo_version.repository, number__lte=repo_version.number).only("pk")
    content_rel = RepositoryContent.objects.filter(version_added__in=repo_version_qs).exclude(version_removed__in=repo_version_qs)

    return qs.filter(pk__in=content_rel.values_list("content_id"))
```

Results:

```
Content Count, New (sec), Old (sec), Ratio
1, 0.007971327304840087, 0.005735735893249511, 0.7195459016927891
3, 0.008062942028045654, 0.006848299503326416, 0.8493549227447873
7, 0.006672074794769287, 0.005816881656646729, 0.8718250072986284
15, 0.007210805416107178, 0.0057263350486755375, 0.7941325161658507
31, 0.007328469753265381, 0.008088014125823974, 1.1036429702422061
63, 0.007654345035552979, 0.009096157550811768, 1.1883652368114899
127, 0.009392633438110351, 0.012355432510375977, 1.315438592572361
255, 0.00953948736190796, 0.01563152313232422, 1.6386124892562164
511, 0.014935557842254638, 0.0175215744972229, 1.173144966012055
1023, 0.011724605560302734, 0.017754955291748045, 1.5143328447536792
2047, 0.010672361850738525, 0.02244783639907837, 2.103361628197135
4095, 0.012695586681365967, 0.03511731147766113, 2.766103872080587
8191, 0.019123435020446777, 0.06781854391098023, 3.5463578503793194
16383, 0.024063756465911867, 0.11325654029846191, 4.706519551878709
```

I suspect this will start to break down when you get into the thousands of repo vesions, but initial testing looks good

Comment (newswangerd) [reply pulp_ansible/app/utils.py:55]: Interestingly enough, the old qs didn't start choking at 65k as it did in my original test. I think I had more items in the RepositoryContent table outside of the repo version I was testing before.

Comment (newswangerd) [reply pulp_ansible/app/utils.py:55]: I tried running a second test after the first where I added one content per repo version and the results were pretty attrocious:

```
Content Count, RV Num, New (sec), Old (sec), Ratio
1, 1, 0.056594743728637695, 0.0060496687889099125, 0.10689453455107173
3, 3, 0.054788048267364504, 0.005242955684661865, 0.0956952446832264
7, 7, 0.05434585809707641, 0.005223925113677978, 0.09612370282840385
15, 15, 0.05550434350967407, 0.005522563457489014, 0.09949786103724413
31, 31, 0.05482580661773682, 0.0061423659324646, 0.11203421000791021
63, 63, 0.056079366207122804, 0.006893675327301025, 0.12292712620609894
127, 127, 0.057129118442535404, 0.008924922943115234, 0.1562237119428434
255, 255, 0.05769376516342163, 0.009042255878448487, 0.1567284758211856
511, 511, 0.18193699836730956, 0.012168281078338623, 0.06688183924949825
```

Interestingly, after I wiped the database and started over, it got better. This new apprach seems to be sensitive to the number of total items in the RepositoryContent table for some reason.

```
Content Count, RV Num, New (sec), Old (sec), Ratio
1, 2, 0.006590936183929443, 0.005165419578552246, 0.7837156110154718
3, 4, 0.005729966163635254, 0.005085377693176269, 0.8875057108452383
7, 8, 0.005913944244384766, 0.005303664207458496, 0.8968065961214083
15, 16, 0.006420111656188965, 0.0054044342041015625, 0.8417975408405408
31, 32, 0.006573936939239502, 0.005987346172332764, 0.9107702473679954
63, 64, 0.007027056217193604, 0.006422135829925537, 0.9139155332516105
127, 128, 0.008137497901916504, 0.0076317286491394045, 0.9378470804080966
255, 256, 0.008519001007080078, 0.008491809368133546, 0.9968081188247385
511, 512, 0.010964252948760987, 0.01227822780609131, 1.11984171319933
1023, 1024, 0.011482336521148682, 0.01439176082611084, 1.2533826020168848
2047, 2048, 0.012757422924041749, 0.023115384578704833, 1.811916459643522
```

Comment (newswangerd) [reply pulp_ansible/app/utils.py:55]: @gerrod3 After a bunch of testing, i think that the slowest part of the process is `qs.count()`. I also haven't really nailed down what's going on with these queries. In some tests, loading the repo versions first improved query performance by 30x. In other cases it has been 20% slower. Setting the number of retained versions to 1 doesn't seem to help much when you get up into the hundreds of thousands. I think we should be fine with the original filter I had in my PR along with perhaps some caching or maybe using a smarter queryset for the pagination counts.

File: pulp_ansible/app/galaxy/v3/views.py (hunk 3 above)
Comment (newswangerd) [reply pulp_ansible/app/galaxy/v3/views.py:286]: I don't think that would matter. It's already filtering by `.filter(collection=OuterRef("pk"))`

File: pulp_ansible/app/galaxy/v3/views.py [hunk 5]
```diff
@@ -257,108 +277,64 @@ def get_queryset(self):
...
+            .select_related("version_added")
+            .select_related("version_removed")
+            .filter(content__ansible_collectionversion__collection_id=OuterRef("pk"))
+            .annotate(
+                last_updated=Greatest(
+                    "version_added__pulp_created", "version_removed__pulp_created"
+                )
+            )
+            .order_by("-last_updated")
+            .only("last_updated")
```
gerrod3, line 313: This seems overly complicated. RepositoryContent should only get updated when the content is removed (version_removed gets set) so we probably don't need to perform this `select_related`, just check the `pulp_last_updated` on the RepositoryContent.
```suggestion
            RepositoryContent.objects.filter(
                repository_id=repo_version.repository_id,
                version_added__number__lte=repo_version.number,
            )
            .select_related("content__ansible_collectionversion")
            .filter(content__ansible_collectionversion__collection_id=OuterRef("pk"))
            .order_by("-pulp_last_updated")
            .only("pulp_last_updated")
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -878,27 +856,82 @@ def list(self, request, *args, **kwargs):
...
+            "collection",
+        )
 
         queryset = self.filter_queryset(queryset)
 
-        # This is -very- slow when the collection has many versions.
-        # queryset = sorted(
-        #    queryset, key=lambda obj: semantic_version.Version(obj.version), reverse=True
-        # )
-
```
gerrod3, line 892: Do we not want to bring back sorting by version now that version has been broken up? Right now I think the returned results are default sorted by `pulp_created` which will typically be the same as sorted versions.

gerrod3 (changes requested):
Looking good. Have some questions and suggestions before final approval.

pulp_ansible/app/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+
+    repo_version_qs = RepositoryVersion.objects.filter(
+        repository=repo_version.repository, number__lte=repo_version.number
+    ).values_list("pk")
+
+    f = Q(version_added__in=repo_version_qs) & Q(
+        Q(version_removed=None) | ~Q(version_removed__in=repo_version_qs)
+    )
+    content_rel = RepositoryContent.objects.filter(f)
```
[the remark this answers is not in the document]
newswangerd: In my testing this was slightly more efficient. The key here is `Q(version_removed=None) | ~Q(version_removed__in=repo_version_qs)`. If the version remove is none, then it prevents the lookup in the `repo_version_qs` in postgres from happening. When you use `.exclude()` it will always perform the `version_removed__in` lookup

pulp_ansible/app/galaxy/v3/views.py
newswangerd: I tried using `pulp_last_updated` and it doesn't work. That field doesn't seem to get updated when version removed is added to the RepositoryContent entry

pulp_ansible/app/galaxy/v3/views.py
newswangerd: sorting by versions will still likely be slower, so I'm not sure it's worth bringing this back since it seems to be working fine.

gerrod3 (changes requested):
Final changes needed.

pulp_ansible/app/utils.py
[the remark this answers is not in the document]
gerrod3: I think you should add `Q(repository=repo_version.repository_id)`  to the filter.

PR #1477, opened 2023-06-02
Expose collection download count to the api

bmclaughlin:
Expose collection download counts (aggregate across collection versions) to the api. 
https://issues.redhat.com/browse/AAH-2236
[noissue]


pulp_ansible/app/galaxy/v3/serializers.py
```diff
@@ -96,6 +98,15 @@ def get_highest_version(self, obj):
...
         return {"href": href, "version": version}
 
+    def get_download_count(self, obj):
+        """Get the download count of the collection"""
+        dlc_model = models.CollectionDownloadCount
+        if dlc_model.objects.filter(namespace=obj.namespace, name=obj.name).count() == 0:
+            return 0
+        return (
+            dlc_model.objects.filter(namespace=obj.namespace, name=obj.name).first().download_count
+        )
```
mdellweg, line 108: Lets use a reusable queryset...
```suggestion
        qs = models.CollectionDownloadCount.objects.filter(namespace=obj.namespace, name=obj.name)
        if qs.count() == 0:
            return 0
        return qs.first().download_count
```

bmclaughlin: Wow, how'd I miss that.  Thanks!

pulp_ansible/tests/functional/cli/test_collection_download_count.py
```diff
@@ -0,0 +1,125 @@
+"""Tests that Collections hosted by Pulp can be installed by ansible-galaxy."""
+from os import path
+import subprocess
+import pytest
+from urllib.parse import urljoin
+
+from pulpcore.client.pulp_ansible import AnsibleRepositorySyncURL
+from pulp_smash import api, config
+from pulp_smash.pulp3.bindings import monitor_task
```
mdellweg, line 9: Please try not to use pulp_smash anymore.

`monitor_task` is available a s a pytest fixture these days.
And instead of the pulp_client fixture, a simple call to requests may do the trick, no?

template_config.yml
```diff
@@ -56,6 +56,7 @@ pulp_settings:
   allowed_import_paths: /tmp
   ansible_api_hostname: https://pulp:443
   ansible_collect_download_log: true
+  ansible_collect_download_count: true
```
mdellweg, line 59: It would be good to have these settings only on one of the test scenarios.
(Their tests would need to be optionally skipped though.)

pulp_ansible/app/models.py
```diff
@@ -372,6 +372,19 @@ class DownloadLog(BaseModel):
     )
 
 
+class CollectionDownloadCount(BaseModel):
```
mdellweg, line 375: Is this orthogonal to the CollectionDownloadLog?

bmclaughlin: I looked into using `CollectionDownloadLog` to base the data off of, but there were numerous complications that were eliminated with a separate `CollectionDownloadCount`.

The complications included counting objects in a log that is very similar to an access log (high activity on community Galaxy exacerbates this). That lead to caching strategies (always fun) and time stamp based download count updates via a `TaskSchedule` as well as log maintenance.

awcrosby: Also on a high traffic system (multiple million downloads weekly), `CollectionDownloadLog` may need to have records deleted occasionally due to db growth. So yes they are orthogonal.

awcrosby (commented):
@bmclaughlin overall this seems like a good approach to meet the need to display counts in the api and ui

template_config.yml
bmclaughlin: I'm not exactly sure how to accomplish this, is there an existing example I could use for inspiration?  Assuming this is regarding both `ansible_collect_download_log` and `ansible_collection_download_count`?

pulp_ansible/tests/functional/cli/test_collection_download_count.py
```diff
@@ -0,0 +1,111 @@
...
+        assert not path.exists(directory), "Directory {} already exists".format(directory)
+        subprocess.run(cmd, cwd=temp_dir)
+        assert path.exists(directory), "Could not find directory {}".format(directory)
+
+        collection_detail = galaxy_v3_collections_api_client.read(
+            path=install_scenario_distribution.base_path,
+            namespace=collection_namespace,
+            name=collection_name,
+        )
+        assert collection_detail.download_count == "3"
```
mdellweg, line 91: Why do the install commands not add to the count?
I'm clearly missing something.

template_config.yml
mdellweg: You would move these two lines to the specific scenario settings below (`pulp_settings_azure`, ...). I think we can have all possible 4 combinations. ;)
For the tests, there is a `pulp_settings` fixture and you can conditionally `pytest.skip("")` a test.

pulp_ansible/tests/functional/cli/test_collection_download_count.py
```diff
@@ -0,0 +1,111 @@
...
+):
+    """Test that the collection download counting functions."""
+    with pulp_admin_user:
+        collection_namespace = ANSIBLE_DEMO_COLLECTION.split(".")[0]
+        collection_name = ANSIBLE_DEMO_COLLECTION.split(".")[1]
+        collection_detail = galaxy_v3_collections_api_client.read(
+            path=install_scenario_distribution.base_path,
+            namespace=collection_namespace,
+            name=collection_name,
+        )
```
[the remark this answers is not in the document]
bmclaughlin: This is reading the collection detail page to retrieve the current download count, the `subprocess.run(cmd, cwd=temp_dir)` lines are downloading the collection.

pulp_ansible/tests/functional/cli/test_collection_download_count.py
```diff
@@ -0,0 +1,111 @@
...
+
+        collection_detail = galaxy_v3_collections_api_client.read(
+            path=install_scenario_distribution.base_path,
+            namespace=collection_namespace,
+            name=collection_name,
+        )
```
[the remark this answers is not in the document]
bmclaughlin: The demo collection currently only has one collection version, can another version be generated?  I do not have the permissions to do so.  Downloading different versions of the same collection does increment the count in manual testing.

pulp_ansible/tests/functional/cli/test_collection_download_count.py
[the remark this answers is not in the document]
mdellweg: Why are we doing it twice and why is the counter at 2 at this point?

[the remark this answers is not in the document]
bmclaughlin: The counter is 2 at this point due to both tests in [tests.functional.cli.test_collection_install](https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/cli/test_collection_install.py) install the demo collection prior to this test being run and add 2 entries to the `CollectionDownloadCount` table. 

I envisioned the first check showing 0 (establishing the baseline), the second check showing 1 (against one collection version) and the third check showing 2 (against a different collection version) to show progression in the test.

pulp_ansible/tests/functional/cli/test_collection_download_count.py
bmclaughlin: The `subprocess.run(cmd, cwd=temp_dir)` calls are what increment the count, I check the result in the api to test the changes to the collection serializer.

template_config.yml
```diff
@@ -55,14 +55,20 @@ pulp_settings:
...
+pulp_settings_gcp:
+  ansible_collect_download_log: true
+  ansible_collect_download_count: true
+pulp_settings_s3:
+  ansible_collect_download_log: true
+  ansible_collect_download_count: true
+pulp_settings_stream:
+  ansible_collect_download_log: true
+  ansible_collect_download_count: true
```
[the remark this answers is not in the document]
mdellweg: stream is really not a good test these days...

pulp_ansible/app/models.py
```diff
@@ -372,6 +372,19 @@ class DownloadLog(BaseModel):
...
+    """
+    Aggregate count of downloads per collection
+    """
+
+    namespace = models.CharField(max_length=64, editable=False)
+    name = models.CharField(max_length=64, editable=False)
+    download_count = models.IntegerField(default=0)
```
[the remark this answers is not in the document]
mdellweg: Would this help to make it more future proof?

pulp_ansible/tests/functional/cli/test_collection_download_count.py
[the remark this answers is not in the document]
mdellweg: That sounds dangerous. We cannot assume that the order of test execution is stable.

pulp_ansible/tests/functional/cli/test_collection_download_count.py
```diff
@@ -0,0 +1,109 @@
...
```
mdellweg, line 62:
```suggestion
```
This call really looks redundant to me. If you say, the dl-count is interfered with by another test, let's not check that it goes from 2 to 3, but that it increases by exactly 1.

pulp_ansible/tests/functional/cli/test_collection_download_count.py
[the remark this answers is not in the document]
bmclaughlin: Makes sense.  I'll re-write that portion to get the base number and base assertions on that number being incremented as expected.

pulp_ansible/app/models.py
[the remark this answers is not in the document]
bmclaughlin: Definitely, I just missed this yesterday.  Coming shortly!

pulp_ansible/tests/functional/cli/test_collection_download_count.py
bmclaughlin: I'm going to re-write this to be clearer.

mdellweg (approved):

mdellweg (approved):

PR #1484, opened 2023-06-20
Gracefully skip download counter increment if database is read-only.

jctanner:
https://issues.redhat.com/browse/AAH-2416
https://issues.redhat.com/browse/AAH-2333

https://github.com/pulp/pulp_ansible/issues/1491

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -12,6 +12,7 @@ from django_filters import filters
 from django.views.generic.base import RedirectView
 from django.conf import settings
 from django.core.cache import cache
+from django.db.utils import InternalError as DatabaseInternalError
 
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
 from jinja2 import Template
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -669,9 +670,16 @@ def _get_org_id(request):
 
     def count_download(filename):
         ns, name, _ = filename.split("-", maxsplit=2)
-        collection, _ = CollectionDownloadCount.objects.get_or_create(namespace=ns, name=name)
-        collection.download_count = F("download_count") + 1
-        collection.save()
+        try:
+            collection, _ = CollectionDownloadCount.objects.get_or_create(namespace=ns, name=name)
+            collection.download_count = F("download_count") + 1
+            collection.save()
```
mdellweg, line 676:
```suggestion
            collection, created = CollectionDownloadCount.objects.get_or_create(namespace=ns, name=name, defaults={"download_count": 1})
            if not created:
                collection.download_count = F("download_count") + 1
                collection.save()
```

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -669,9 +670,16 @@ def _get_org_id(request):
...
+        except DatabaseInternalError as e:
+            # handle a read-only replica scenario
+            msg = str(e)
+            if "read-only" in msg:
+                return
+            raise (e)
```
mdellweg, line 682:
```suggestion
            raise
```

mdellweg (commented):
Would the same care be needed for the CollectionDownloadLog?

jctanner:
@mdellweg I didn't actually see that cause an error in testing ... perhaps because it's not on by default for the functional tests?

newswangerd (approved):

pulp_ansible/app/galaxy/v3/views.py
jctanner: Although this suggestion does reduce the overall db calls, the model's default value is zero and this new code causes tests to fail because the first download doesn't increment the counter.

```
>           assert (download_count + 1) == get_current_download_count(                                                                      
                api_client=galaxy_v3_collections_api_client,          
                path=install_scenario_distribution.base_path,         
                namespace=collection_namespace,                       
                name=collection_name,                                 
            )                                                         
E           AssertionError: assert (0 + 1) == 0 
```

jctanner:
I added and tested the same try except for the download log as well.

pulp_ansible/app/galaxy/v3/views.py
mdellweg: But we could change that default to 1. I think this was implied. Sorry.

mdellweg: We do have it on in some scenarios:
https://github.com/pulp/pulp_ansible/blob/main/template_config.yml#L60

And on close inspection, the suggestion overwrites the default inline (`defaults={"download_count": 1}`) with the `get_or_create` that's why i think it should pass the tests.

mdellweg (approved):
I would still like to understand, why the get_or_create with a default to 1 fails for you. I believe there are two scenarios that actually test this codepath.
But also I'm not going to block this.

jctanner:
@mdellweg ah, i didn't fully grok what you wanted. It's added back now and tests pass.

mdellweg (approved):

PR #1492, opened 2023-06-30
use most recent namespacemetadata in x-repo search indexing

jctanner:
https://github.com/pulp/pulp_ansible/issues/1494

### DIFF
--- a/pulp_ansible/app/tasks/collectionversion_index.py
+++ b/pulp_ansible/app/tasks/collectionversion_index.py
@@ -39,7 +39,6 @@ def get_highest_version_string_from_cv_objects(cv_objects):
 
 def compute_repository_changes(repository_version):
     """Use the previous version to make a list of namespace(s).name(s) changed."""
-
     # Figure out what the previous repo version is
     repository = repository_version.repository
     previous_number = repository_version.number - 1
@@ -56,6 +55,7 @@ def compute_repository_changes(repository_version):
     cv_type = CollectionVersion.get_pulp_type()
     deprecation_type = AnsibleCollectionDeprecated.get_pulp_type()
     signature_type = CollectionVersionSignature.get_pulp_type()
+    metadata_type = AnsibleNamespaceMetadata.get_pulp_type()
 
     for func in [repository_version.added, repository_version.removed]:
         for modified in func(base_version=previous_version):
@@ -70,6 +70,9 @@ def compute_repository_changes(repository_version):
                 changed_collections.add(
                     (signature.signed_collection.namespace, signature.signed_collection.name)
                 )
+            elif modified.pulp_type == metadata_type:
+                metadata = modified.ansible_ansiblenamespacemetadata
+                changed_collections.add((metadata.name, "*"))
 
     return changed_collections
 
@@ -134,6 +137,8 @@ def update_index(distribution=None, repository=None, repository_version=None, is
 
     # What has changed between this version and the last?
     changed_collections = compute_repository_changes(repository_version)
+    if not changed_collections:
+        return
 
     # get all CVs in this repository version
     cvs_pks = repository_version.content.filter(pulp_type="ansible.collection_version").values_list(
@@ -154,9 +159,14 @@ def update_index(distribution=None, repository=None, repository_version=None, is
     deprecations = AnsibleCollectionDeprecated.objects.filter(pk__in=deprecations)
     deprecations_set = {(x.namespace, x.name) for x in deprecations}
 
-    # find all namespaces in the repo version
-    namespaces = repository_version.get_content(content_qs=AnsibleNamespaceMetadata.objects).all()
-    namespaces = {x.name: x for x in namespaces}
+    # find all the most recent namespace metadata in the repo version
+    namespaces = {}
+    for ns in repository_version.get_content(content_qs=AnsibleNamespaceMetadata.objects).all():
+        if ns.name not in namespaces:
+            namespaces[ns.name] = ns
+            continue
+        if namespaces[ns.name].timestamp_of_interest < ns.timestamp_of_interest:
+            namespaces[ns.name] = ns
 
     # map out the namespace(s).name(s) for everything in the repo version
     colset = set(cvs.values_list("namespace", "name").distinct())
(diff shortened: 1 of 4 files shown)


pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -1434,3 +1434,92 @@ def test_cross_repo_search_index_on_distribution_with_repository_and_deprecation
...
+    # define col name
+    collection_name = randstr()
+
+    # make namespace metadata
+    kwargs = {"path": distro.base_path, "distro_base_path": distro.base_path}
+    task = galaxy_v3_plugin_namespaces_api_client.create(
+        name=namespace_name, description="hello", company="Testing Co.", **kwargs
+    )
+    result = monitor_task(task.task)
+    namespace_href = result.created_resources[1]
```
mdellweg, line 1481: I'm still unsure whether that order is stable.
(Reminds me that i want tasks to be able to actually return something.)

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -1434,3 +1434,92 @@ def test_cross_repo_search_index_on_distribution_with_repository_and_deprecation
...
+    assert cv.namespace_metadata.pulp_href == namespace_href
+    assert cv.namespace_metadata.avatar_url is None
+    # assert cv.namespace_metadata.avatar_sha256 is None
+    assert cv.namespace_metadata.company == "Testing Co."
+    assert cv.namespace_metadata.description == "hello"
+    assert cv.namespace_metadata.name == namespace_name
+
+    # update the namespace metadata with an avatar
+    avatar_path = random_image_factory()
+    kwargs2 = {"path": distro.base_path, "distro_base_path": distro.base_path}
```
mdellweg, line 1503: This is the same as kwargs above, no?
Can we call it a little more useful and reuse it? dist_kwargs, maybe?

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -1434,3 +1434,92 @@ def test_cross_repo_search_index_on_distribution_with_repository_and_deprecation
...
+
+    # make sure the CV was re-indexed
+    resp2 = galaxy_v3_default_search_api_client.list(limit=1000, distribution=[dist_id])
+    assert resp2.meta.count == 1, resp2
+
+    # did it get all the NEW metadata?
+    cv2 = resp2.data[0]
+    assert cv2.namespace_metadata.pulp_href == namespace2_href
+    assert cv2.namespace_metadata.avatar_url is not None
+    # assert cv2.namespace_metadata.avatar_sha256 is not None
```
mdellweg, line 1522: Can we not assert on that? Wouldn't we even know the sha256 of the image we generated above?

mdellweg (approved):
I'm not enough of an expert for these namespaces, but the test looks well written.
Please wait for a second review.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
jctanner: It another bug that I found last night. I'm not sure if it's an issue in the serializer or what, but I wasn't sure if it's something we can easily solve in this PR and still have a backportable patch. We're most interested in the avatar_url for the UI elements and the sha256 is less concerning.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
jctanner: I tried to use gen_object_with_cleanup, but something about that api client or maybe the async task is incompatible.

mdellweg: Hah, no a task with two returned entitites is not compatible with that fixture.

mdellweg: You can still use the add_to_cleanup fixture though.

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -154,9 +157,14 @@ def update_index(distribution=None, repository=None, repository_version=None, is
     deprecations = AnsibleCollectionDeprecated.objects.filter(pk__in=deprecations)
     deprecations_set = {(x.namespace, x.name) for x in deprecations}
 
-    # find all namespaces in the repo version
-    namespaces = repository_version.get_content(content_qs=AnsibleNamespaceMetadata.objects).all()
-    namespaces = {x.name: x for x in namespaces}
+    # find all the most recent namespace metadata in the repo version
+    namespaces = {}
+    for ns in AnsibleNamespaceMetadata.objects.all():
```
awcrosby, line 162: this looks like it gets all `AnsibleNamespaceMetadata` objects, not just the ones scoped to the repository_version

jctanner: fixed

pulp_ansible/app/tasks/collectionversion_index.py
```diff
@@ -175,7 +183,11 @@ def update_index(distribution=None, repository=None, repository_version=None, is
     for colkey in colset:
         namespace, name = colkey
 
-        if changed_collections is not None and (namespace, name) not in changed_collections:
+        if (
+            changed_collections is not None
+            and (namespace, name) not in changed_collections
+            and (namespace, "*") not in changed_collections
+        ):
```
awcrosby, line 190: I'm not sure what circumstance would make `changed_collections` be `None`, but  (prior to this PR and also with this PR) this logic seems to do this:

if `changed_collections` is `None`, it will not `continue`, so it will attempt the rest of the create|update index... is that desired?

jctanner: That is correct. However changed_collections reflects changes to CVs, deprecatation, signatures and now namespacemetadata.

awcrosby:
```suggestion
        if changed_collections is None:
            continue
        if (
            (namespace, name) not in changed_collections
            and (namespace, "*") not in changed_collections
        ):
```

IIUC, and this is logically correct, this is a suggestion to be explicty that there is no indexing when no changed collections, optional edit

awcrosby: probably better to just remove check for `changed_collections is not None`

rochacbruno (approved):

Issue #1506, opened 2023-07-04
Performance issue on `pulp/api/v3/repositories/ansible/ansible/`


jerabekjiri:
Issue: [AAH-2367](https://issues.redhat.com/browse/AAH-2367)

The AAH issue mentions slowness on the CRC stage and production environment. However, the slowness becomes more apparent in any environment when there are multiple repositories and a large number of tasks in the database.

PR #1507, opened 2023-07-04
Enhance `/repositories/ansible/ansible/` endpoint performance

jerabekjiri:
fixes: #1506

Issue: [AAH-2367](https://issues.redhat.com/browse/AAH-2367)

Performance testing:

Number of repositories: 493

current performance: 
```
Time
CPU: 14230.32ms (15279.12ms)
SQL: 1663 queries in 1773.93ms
```
Screenshot from 2023-08-23 13-17-44

updated solution:
```
Time
CPU: 2972.00ms (3486.00ms)
SQL: 6 queries in 525.13ms
```
Screenshot from 2023-08-23 13-02-38

### DIFF
--- a/pulp_ansible/app/utils.py
+++ b/pulp_ansible/app/utils.py
@@ -29,8 +29,8 @@ def filter_content_for_repo_version(qs, repo_version):
     return qs.filter(pk__in=content_rel.values_list("content_id"))
 
 
-def get_queryset_annotated_with_task_sync_task(qs):
-    tasks = (
+def get_queryset_annotated_with_last_sync_task(qs):
+    last_task = (
         Task.objects.filter(
             name__contains="sync",
             reserved_resources_record__icontains=Cast(OuterRef("pk"), output_field=CharField()),
@@ -47,4 +47,4 @@ def get_queryset_annotated_with_task_sync_task(qs):
         .order_by("-pulp_created")[:1]
     )
 
-    return qs.annotate(last_sync_task=Subquery(tasks))
+    return qs.annotate(last_sync_task=Subquery(last_task))
--- a/pulp_ansible/app/utils.py
+++ b/pulp_ansible/app/utils.py
@@ -1,5 +1,6 @@
-from django.db.models import Q
-from pulpcore.plugin.models import RepositoryVersion, RepositoryContent
+from django.db.models import Q, fields, OuterRef, Subquery, F
+from pulpcore.plugin.models import RepositoryVersion, RepositoryContent, Task
+from django.db.models.functions import JSONObject, Cast
 
 
 def filter_content_for_repo_version(qs, repo_version):
@@ -26,3 +27,25 @@ def filter_content_for_repo_version(qs, repo_version):
     content_rel = RepositoryContent.objects.filter(f)
 
     return qs.filter(pk__in=content_rel.values_list("content_id"))
+
+
+def get_queryset_annotated_with_task_sync_task(qs):
+    tasks = (
+        Task.objects.filter(
+            name__contains="sync", reserved_resources_record__icontains=OuterRef("outer_pk")
+        )
+        .annotate(
+            task=JSONObject(
+                pk=F("pk"),
+                state=F("state"),
+                pulp_created=F("pulp_created"),
+                finished_at=F("finished_at"),
+                error=F("error"),
+            )
+        )
+        .values("task")[:1]
+    )
+
+    return qs.annotate(outer_pk=Cast("pk", output_field=fields.CharField())).annotate(
+        last_sync_task=Subquery(tasks)
+    )
--- a/pulp_ansible/app/utils.py
+++ b/pulp_ansible/app/utils.py
@@ -43,6 +43,7 @@ def get_queryset_annotated_with_task_sync_task(qs):
                 error=F("error"),
             )
         )
+        .order_by("-pulp_created")
         .values("task")[:1]
     )
 
--- a/pulp_ansible/app/utils.py
+++ b/pulp_ansible/app/utils.py
@@ -1,6 +1,6 @@
-from django.db.models import Q, fields, OuterRef, Subquery, F
+from django.db.models import Q, OuterRef, Subquery, CharField
 from pulpcore.plugin.models import RepositoryVersion, RepositoryContent, Task
-from django.db.models.functions import JSONObject, Cast
+from django.db.models.functions import Cast, JSONObject
 
 
 def filter_content_for_repo_version(qs, repo_version):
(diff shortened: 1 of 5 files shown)


mdellweg:
Have we considered, not showing that field in the list serializer?
Also do you regularly purge old tasks? The tasking system performance may benefit from it too.

jerabekjiri:
@mdellweg 
 >Have we considered, not showing that field in the list serializer?
 
 we need it for state, finished_at and error
 
 
Screenshot from 2023-07-07 11-41-16

>Also do you regularly purge old tasks? The tasking system performance may benefit from it too.
I'm not sure about that I'll ask around

jerabekjiri:
@mdellweg 
>Also do you regularly purge old tasks? The tasking system performance may benefit from it too.

From Slack conversation: *We may have purged tasks once or twice a long time ago as a one-off thing during an upgrade, but in general no*, so periodically remove stale tasks would most likely help performance

mdellweg:
I'm unsure what to say about this. First it is duplicating an information already in the db. These tend to get out of sync and be hard to debug. But that is probably not why we should not do this.
The bigger issue is (both with the current situation as well as with this new approach) that the sync task may be deleted/purged at some point. Also what if the last action on the repository was not a sync, but adding/removing content by any other means? I think the best place for this information to live is the info field of the repository version.
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/repository.py#L668
If it contains only information about how the repository version was created (without directly referring to the task), it will never get outdated (versions are immutable...).

jerabekjiri:
@mdellweg thanks for the insight. First, I'd like to test the impact of deleting old tasks in the CRC stage and If this won't be sufficient, adding it to RepositoryVersion sounds like a good idea.

jerabekjiri:
current solution (@property def last_sync_task):
Screenshot from 2023-07-21 11-16-30


annotated `last_sync_task` to queryset:
Screenshot from 2023-07-21 11-23-13

pulp_ansible/app/utils.py
```diff
@@ -26,3 +27,26 @@ def filter_content_for_repo_version(qs, repo_version):
...
+        )
+        .annotate(
+            task=JSONObject(
+                pk=F("pk"),
+                state=F("state"),
+                pulp_created=F("pulp_created"),
+                finished_at=F("finished_at"),
+                error=F("error"),
+            )
+        )
```
gerrod3, line 45: Why is this needed? Can you just use `only()`?

pulp_ansible/app/utils.py
```diff
@@ -26,3 +27,26 @@ def filter_content_for_repo_version(qs, repo_version):
     content_rel = RepositoryContent.objects.filter(f)
 
     return qs.filter(pk__in=content_rel.values_list("content_id"))
+
+
+def get_queryset_annotated_with_task_sync_task(qs):
+    tasks = (
+        Task.objects.filter(
+            name__contains="sync", reserved_resources_record__icontains=OuterRef("outer_pk")
```
gerrod3, line 35: Could you try putting the `OuterRef` inside a `Cast` so that we don't have to annotate the parent queryset to perform the cast on the pk?

pulp_ansible/app/utils.py
jerabekjiri: I need the `JSONObject` so I can annotate more than one field, made it a slightly better with the latest solution.

pulp_ansible/app/utils.py
jerabekjiri: Yep, I got confused while experimenting with various solutions. Fixed, but still trying to figure out on how to implement prefetch or select_related.

jerabekjiri:
Updated title and description to match the code. Adding images showing performance testing:

before:
Screenshot from 2023-08-23 13-17-44

after:
Screenshot from 2023-08-23 13-02-38

pulp_ansible/app/utils.py
```diff
@@ -26,3 +27,24 @@ def filter_content_for_repo_version(qs, repo_version):
...
+            reserved_resources_record__icontains=Cast(OuterRef("pk"), output_field=CharField()),
+        )
+        .values(
+            json=JSONObject(
+                pk="pk",
+                state="state",
+                pulp_created="pulp_created",
+                finished_at="finished_at",
+                error="error",
+            )
```
gerrod3, line 45: Can you remove this JSONObject and just have the values be the fields?

pulp_ansible/app/viewsets.py
```diff
@@ -987,6 +995,13 @@ class CollectionRemoteViewSet(RemoteViewSet, RolesMixin):
     serializer_class = CollectionRemoteSerializer
     filterset_class = CollectionRemoteFilter
 
+    def get_queryset(self):
+        qs = super().get_queryset()
+        if getattr(self, "action", "") == "list":
+            return get_queryset_annotated_with_task_sync_task(qs)
+
+        return super().get_queryset()
```
gerrod3, line 1003: Can you rewrite this one to be the same as the RepositoryViewset?

gerrod3 (changes requested):
Can you explain how the number of queries is still growing when you have 2000 repositories? I would expect the number of queries would still be 6.

jerabekjiri:
>Can you explain how the number of queries is still growing when you have 2000 repositories? I would expect the number of queries would still be 6.

you're right, I had the Invalid token error in the environment after rebuilding and didn't notice it affected the queryset. 

created some more repositories:
```
6 repositories:
before: 14 queries in 11.76ms
after: 6 queries in 9.52ms
--------------------------------------------
306 repositories:
before: 214 queries in 248.86ms
after:  6 queries in 107.99ms
--------------------------------------------
606 repositories:
before: 406 queries 473.57ms
after: 6 queries in 47.90ms
--------------------------------------------
906 repositories:
before: 400 queries in 580.72ms
after: 6 queries in 46.41ms
```

pulp_ansible/app/utils.py
jerabekjiri: Well, without using `JSONObject`, I get this
```
Exception Value: | subquery must return only one column LINE 1: ...yversion"."complete") AS "latest_version_number", (SELECT U0...                                                              ^
```

pulp_ansible/app/utils.py
```diff
@@ -26,3 +27,24 @@ def filter_content_for_repo_version(qs, repo_version):
     content_rel = RepositoryContent.objects.filter(f)
 
     return qs.filter(pk__in=content_rel.values_list("content_id"))
+
+
+def get_queryset_annotated_with_last_sync_task(qs):
+    last_task = (
+        Task.objects.filter(
+            name__contains="sync",
+            reserved_resources_record__icontains=Cast(OuterRef("pk"), output_field=CharField()),
```
mdellweg, line 36: This assumes that the uuids are truly _universally_ unique. Well...
Can you leave a comment about that?

How bad would it be if the task wasn't found, a wrong one displayed?

pulp_ansible/app/viewsets.py
```diff
@@ -554,6 +555,13 @@ class AnsibleRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin, R
     queryset = AnsibleRepository.objects.all()
     serializer_class = AnsibleRepositorySerializer
 
+    def get_queryset(self):
+        qs = super().get_queryset()
+        if getattr(self, "action", "") == "list":
```
mdellweg, line 560: When doing the read (single object), do we still want to see the task?

jerabekjiri: oops... good catch :+1: . Yes, UI uses it in detail as well.

pulp_ansible/app/utils.py
jerabekjiri: if the task wasn't found, null is returned. And with `pk`, `name="sync"` and `order_by`, I think there's a little chance that the wrong task would be displayed.

gerrod3 (approved):

Issue #1516, opened 2023-07-17
param `order_by=version` doesn't sort correctly


jerabekjiri:
AAH issue: [AAH-2581](https://issues.redhat.com/browse/AAH-2581)

**Version**
```
"server_version":"4.8.0dev"
"galaxy_ng_version":"4.8.0dev"
galaxy_ng_commit":"origin/master:a314e11"
"galaxy_importer_version":"0.4.10"
"pulp_core_version":"3.23.4"
"pulp_ansible_version":"0.17.2"
"pulp_container_version":"2.14.5"
```

**Describe the bug**
The order_by=-version param isn't working as expected on x-repo search, so the ui is getting the whole list and 1.7.3 shows up first and it's just marking it "latest".

```
pulp=> SELECT version FROM ansible_collectionversion ORDER BY version DESC;
 version
---------
 1.7.3
 1.21.2
 1.21.1
 1.21.0
 1.20.0
 1.19.3
 1.16.2
 1.15.1
 1.12.1
 1.0.1
 1.0.0
(11 rows)
```

**To Reproduce**
Go to the rhel_system roles collection main page: https://console.redhat.com/ansible/automation-hub/repo/published/redhat/rhel_system_roles/

Click on the collection version drop down, and the collection showing "(latest)" is the v1.7.3 version, not the actual latest, v1.21.1. 

The list comes from the URL: https://console.redhat.com/api/automation-hub/v3/plugin/ansible/search/collection-versions/?repository_name=published&namespace=redhat&name=rhel_system_roles&order_by=-version&offset=0&limit=10

**Expected behavior**
Semantic versions should be sorted in the correct order.

PR #1517, opened 2023-07-17
Fix wrong version ordering in x-repo search

jerabekjiri:
fix: #1516

AAH issue: [AAH-2581](https://issues.redhat.com/browse/AAH-2581)

### DIFF
--- a/CHANGES/1516.bugfix
+++ b/CHANGES/1516.bugfix
@@ -1,1 +1,1 @@
-Fixed version ordering in CollectionVersionSearchFilter
+Fixed ordering by version in the ``/ansible/search/collection-versions/`` endpoint.
--- a/CHANGES/1516.bugfix
+++ b/CHANGES/1516.bugfix
@@ -1,0 +1,1 @@
+Fixed version ordering in CollectionVersionSearchFilter
--- a/pulp_ansible/app/galaxy/v3/filters.py
+++ b/pulp_ansible/app/galaxy/v3/filters.py
@@ -1,11 +1,10 @@
 from django.contrib.postgres.search import SearchQuery
 from django.db.models import fields as db_fields
 from django.db.models import Q
-from django.db.models.expressions import F, Func
+from django.db.models.expressions import F, Func, RawSQL
 from django_filters import (
     filters,
     FilterSet,
-    OrderingFilter,
 )
 import semantic_version
 from rest_framework.exceptions import ValidationError
@@ -15,6 +14,16 @@ from pulpcore.plugin.viewsets import LabelFilter
 
 from pulpcore.plugin.models import RepositoryVersion
 from pulp_ansible.app import models
+
+
+class SemanticVersionOrderingFilter(filters.OrderingFilter):
+    def filter(self, qs, value):
+        if value is not None and any(v in ["version", "-version"] for v in value):
+            return qs.annotate(
+                version_compare=RawSQL("string_to_array(version, '.')::int[]", [])
+            ).order_by(f"{value[0]}_compare")
+
+        return super().filter(qs, value)
 
 
 class CollectionVersionSearchFilter(FilterSet):
@@ -49,7 +58,7 @@ class CollectionVersionSearchFilter(FilterSet):
     keywords = filters.CharFilter(field_name="q", method="filter_by_q")
     repository_label = LabelFilter(label_field_name="repository__pulp_labels")
 
-    order_by = OrderingFilter(
+    order_by = SemanticVersionOrderingFilter(
         choices=(
             ("pulp_created", "by CV created"),
             ("-pulp_created", "by CV created (descending)"),
@@ -64,7 +73,6 @@ class CollectionVersionSearchFilter(FilterSet):
             "collection_version__pulp_created": "pulp_created",
             "collection_version__namespace": "namespace",
             "collection_version__name": "name",
-            "collection_version__version": "version",
         },
     )
 
--- a/pulp_ansible/app/galaxy/v3/filters.py
+++ b/pulp_ansible/app/galaxy/v3/filters.py
@@ -1,7 +1,7 @@
 from django.contrib.postgres.search import SearchQuery
 from django.db.models import fields as db_fields
 from django.db.models import Q
-from django.db.models.expressions import F, Func, RawSQL
+from django.db.models.expressions import F, Func
 from django_filters import (
     filters,
     FilterSet,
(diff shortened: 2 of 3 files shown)


CHANGES/1516.bugfix
```diff
@@ -0,0 +1 @@
+Fixed version ordering in CollectionVersionSearchFilter
```
mdellweg, line 1:
```suggestion
Fixed version ordering in ``CollectionVersionSearchFilter``.
```
But maybe you want to reformulate this in user terms. Which filter on which endpoint was broken? Users do not know about our Python classes.

jerabekjiri:
@mdellweg would it be better to add the `SemanticVersionOrderingFilter` to pulpcore instead of pulp_ansible?

mdellweg:
> @mdellweg would it be better to add the `SemanticVersionOrderingFilter` to pulpcore instead of pulp_ansible?

No, let's keep it here for. I think this is not the last iteration of it. Looks a bit naive to me. Are any letters allowed in ansible collection versions? How does it react to "1.0.0-beta"?
I would love to see some domain specific PostgresQL version-fields however, eventually.

jerabekjiri:
@mdellweg 
> Are any letters allowed in ansible collection versions? How does it react to "1.0.0-beta"?

good point. I checked and yes, they are allowed. We can theoretically have versions like 1.0.0-beta or 1.0.0-rc. So my solution doesn't really solve it, I can remove letters with regex for example, but then again sorting will be wrong.

jerabekjiri:
As you mentioned, the best solution might be to add the version field directly in PostgreSQL https://github.com/theory/pg-semver/ https://pgxn.org/dist/semver/doc/semver.html

jerabekjiri:
Updated sorting, now the sorting works correctly even with the prerelease part.
```
/search/collection-versions/?order_by=-version:
"1.22.2",
"1.22.2-beta",
"1.22.1",
"1.22.1-rc",
"1.22.1-beta",
"1.22.1-alpha",
"1.7.3",
"1.0.0"
```

jctanner:
Can you enable postgres query logging and show us the raw sql this annotation is adding to the query?

jerabekjiri:
hmm... for some reason the annotated field `prerelease` show as 9
```
 ORDER BY "ansible_collectionversion"."version_major" ASC,
          "ansible_collectionversion"."version_minor" ASC,
          "ansible_collectionversion"."version_patch" ASC,
          9 ASC
 ```
 
 https://gist.github.com/jerabekjiri/8eb71d6e0764f8761e5443df53b6440e

**Review by jctanner — CHANGES_REQUESTED** [review]
Seems fine, but it needs a functional test.

File: pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py [hunk 2]
```diff
@@ -1523,3 +1523,48 @@ def test_cross_repo_search_index_with_updated_namespace_metadata(
...
+    ]
+
+    for version in versions:
+        build_and_upload_collection(ansible_repo=pulp_repo, config={"version": version})
+
+    resp = galaxy_v3_default_search_api_client.list(limit=1000, order_by=["version"])
+
+    built_collection_versions = [col.collection_version.version for col in resp.data]
+
+    versions.sort(key=Version)
```
mdellweg, line 1570: Is the versions array not yet ordered correctly?

jerabekjiri: It is, though reversed. But I wanted it to go through the Version sort just to make sure or in case the versions array would be changed.

jerabekjiri: Not sure I get that right, but updated the test with the correctly sorted `versions` array assert.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -1562,11 +1562,26 @@ def test_cross_repo_search_semantic_version_ordering(
...
+        "1.22.1",
+        "1.22.1-rc",
+        "1.22.1-pre",
+        "1.22.1-dev",
+        "1.22.1-beta",
+        "1.22.1-alpha",
+        "1.1.0",
+        "1.0.1",
+        "1.0.0",
+    ]
```
mdellweg, line 1585: Sorry, i didn't mean to duplicate the written list of versions.

```suggestion
    assert versions == sorted(versions, key=Version, reverse==True)
```

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -17,6 +16,26 @@
...
+            return qs.annotate(
+                prerelease=Case(
+                    When(collection_version__version_prerelease="", then=Value(None)),
+                    default="collection_version__version_prerelease",
+                ),
+            ).order_by(
+                f"{order}collection_version__version_major",
+                f"{order}collection_version__version_minor",
+                f"{order}collection_version__version_patch",
+                f"{order}prerelease",
```
mdellweg, line 33: Looks like these happen to be alright in alphabetical order.

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -64,7 +83,6 @@ class CollectionVersionSearchFilter(FilterSet):
             "collection_version__pulp_created": "pulp_created",
             "collection_version__namespace": "namespace",
             "collection_version__name": "name",
-            "collection_version__version": "version",
```
mdellweg, line 67: What is this change doing?

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -1523,3 +1523,51 @@ def test_cross_repo_search_index_with_updated_namespace_metadata(
...
+    resp = galaxy_v3_default_search_api_client.list(
+        limit=1000, order_by=["-version"], repository_name=[pulp_repo.name]
+    )
+
+    built_collection_versions = [col.collection_version.version for col in resp.data]
+
+    # Make sure versions are correctly sorted according to Semantic Versioning.
+    assert versions == sorted(versions, key=Version, reverse=True)
+
+    assert versions == built_collection_versions
```
mdellweg, line 1573: Nice test!

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
jerabekjiri: Sorry, I got a little confused before. Thanks for providing a solution :) :+1:

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
jerabekjiri: Sorry, I got a little confused before. Thanks for providing a solution :) :+1:

pulp_ansible/app/galaxy/v3/filters.py
jerabekjiri: Well, since we have a custom `SemanticVersionOrderingFilter`, we don't want this wrong sorting by collection version.

mdellweg (approved):

PR #1523, opened 2023-07-19
Reject collection process

jerabekjiri:
[noissue]

I'll add an issue if this will seem like a good idea to add to pulp_ansible.

AAH issue: [AAH-2599](https://issues.redhat.com/browse/AAH-2599)

questions:
- do we allow multiple rejected repositories?
- should this work as a bulk action, rejecting multiple collections at the same time?

cc @newswangerd about questions

### DIFF
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -953,6 +953,22 @@ class AnsibleRepositoryMarkSerializer(serializers.Serializer):
         return value
 
 
+class AnsibleRepositoryRejectSerializer(serializers.Serializer):
+    collection_version = DetailRelatedField(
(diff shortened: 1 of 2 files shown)


MilanPospisil:
The framework treats bulk actions like multiple parallel operations, so you dont need to implement it. But you must ensure that db will not break with some bad parallel conflicting operations. However, this is requirement anyway even without bulk actions in UI.

jerabekjiri:
Closing. @MilanPospisil I verified and you can use `/pulp/api/v3/repositories/ansible/ansible/{pulp_id}/move_collection_version/`. This works well and doesn't fail on moving collection to rejected repo if a copy is already there.

Issue #1543, opened 2023-08-02
Sync should not fail if namespace's avatar fails to download


gerrod3:
Allow the sync pipeline to skip bad downloads of namespace avatars to make sync more resilient. 

Related:
https://github.com/pulp/pulp_ansible/issues/1415

newswangerd:
Jira: 
- https://issues.redhat.com/browse/AAH-2619
- https://issues.redhat.com/browse/AAP-14090

PR #1544, opened 2023-08-02
Fix sync failure on unreachable namespace avatar

gerrod3:
fixes: #1543

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -10,7 +10,7 @@ from urllib.parse import urljoin
 from uuid import uuid4
 
 import yaml
-from aiohttp.client_exceptions import ClientResponseError
+from aiohttp.client_exceptions import ClientError, ClientResponseError
 from asgiref.sync import sync_to_async
 from async_lru import alru_cache
 from django.conf import settings
@@ -496,7 +496,7 @@ class AnsibleDeclarativeVersion(DeclarativeVersion):
             ArtifactSaver(),
             QueryExistingContents(),
             DocsBlobDownloader(),
-            CollectionContentSaver(new_version),
+            AnsibleContentSaver(new_version),
             RemoteArtifactSaver(),
             ResolveContentFutures(),
         ]
@@ -726,8 +726,8 @@ class CollectionSyncFirstStage(Stage):
 
             da = (
                 [
-                    DeclarativeArtifact(
-                        Artifact(sha256=namespace["avatar_sha256"]),
+                    DeclarativeFailsafeArtifact(
+                        Artifact(sha256=namespace.get("avatar_sha256")),
                         url=url,
                         remote=self.remote,
                         relative_path=f"{name}-avatar",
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -1074,6 +1074,25 @@ async def run(self):
...
+
+    async def download(self):
+        """Allow download to fail, but not stop the sync."""
+        artifact_copy = self.artifact
+        try:
+            return await super().download()
+        except ClientError as e:
+            # Reset DA so that future stages can properly handle it
+            self.artifact = artifact_copy
+            self.deferred_download = True
```
mdellweg, line 1090: It reminds me a lot of https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/tasks/synchronizing.py#L192 with https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/tasks/synchronizing.py#L513 .

Can you elaborate more on the strategy? Do we handle the namespace as if no avatar existed, or do we turn the artifact into on-demand?

pulp_ansible/app/tasks/collections.py
```diff
@@ -1149,6 +1168,15 @@ def _pre_save(self, batch):
...
                 d_content.content.namespace = namespace
+                if d_content.d_artifacts:
+                    da = d_content.d_artifacts[0]
+                    # Check to see if avatar failed to download, update metadata if so
+                    if da.deferred_download:
+                        d_content.d_artifacts = None
+                        d_content.content.avatar_sha256 = None
+                    # Check to see if upstream didn't have avatar_sha256 set
+                    elif d_content.content.avatar_sha256 is None:
+                        d_content.content.avatar_sha256 = da.artifact.sha256
```
mdellweg, line 1179: The name of this class does not match anymore.

pulp_ansible/app/tasks/collections.py
```diff
@@ -1074,6 +1074,25 @@ async def run(self):
             pb.total = pb.done
 
 
+class DeclarativeNamespaceAvatar(DeclarativeArtifact):
```
mdellweg, line 1077: Maybe give it a more functional name (like the DeclarativeFailsafeArtifact).

pulp_ansible/app/tasks/collections.py
gerrod3: It is very similar. The current strategy I am using is to handle the namespace as if no avatar existed. In the custom content saver stage I modify the namespace's metadata before save if the avatar failed to download.

pulp_ansible/app/tasks/collections.py
mdellweg: It's called `CollectionContentSaver`, yet it manipulates all types of content that come flying by (not your fault, but we could take the chance and make it a `AnsibleContentSaver`.).

pulp_ansible/app/tasks/collections.py
gerrod3: Probably, we could design it to accept a list of exceptions during download that would be failsafe.

mdellweg (approved):

mdellweg:
AAP-14090

Issue #1547, opened 2023-08-07
Collection sync update_highest_version calculation can potentially fail on re-sync


gerrod3:
**Version**
Since this commit: https://github.com/pulp/pulp_ansible/commit/8ad72bab7c318b2ddebc98474fc658535c34a46b

**Describe the bug**
When syncing collections the collection version with the highest version is specially marked with the `is_highest=True` field. There is an uniqueness constraint on this field to only allow one collection version from a collection to be marked as the highest at a time. Due to the design of the sync pipeline there is a special scenario where the current update logic tries to save two collection versions as the highest resulting in the constraint being triggered and the sync failing. Here's how:

1. Sync a collection and its versions. Let's say collection A and versions 1, 2 & 3.
2. CV A-3 will be marked `is_highest=True`
3. Collection A has a new version released upstream, 4.
4. Re-sync collection A
5. Sync task creates potential CVs A-1, A-2, A-3, & A-4 and sends them down the sync pipeline. A-4 gets sent before the others.
6. Pipeline discovers that A-1, A-2, & A-3 already exist and replaces them with their current records, these are now in-memory objects.
7. A-4 reaches the save stage and gets saved. `update_highest_version` properly sets A-4 as the new highest and updates the record of A-3 to no longer be highest.
8. A-3 reaches the save stage, but now its in-memory representation is desynced with its DB representation. The custom save stage then tries to save the instance again (even if nothing has changed) and thus triggers the constraint since the model will try to save with `is_highest=True`.

**To Reproduce**
Sync a repository, wait for some new collection-versions to be uploaded, and then sync again. I think the error might occur only sporadically due to the async nature of the sync pipeline. 

Note this issue is a regression and has been fixed before in https://github.com/pulp/pulp_ansible/pull/481.
I am going to apply the same fix, but a refactoring of the `CollectionSaver` stage might be needed to better account for assumptions made on when to save/update collections.

PR #1548, opened 2023-08-07
Fix is_highest logic on resyncs

gerrod3:
fixes: #1547

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -428,7 +428,9 @@ def _update_highest_version(collection_version, save=False):
     If this version is the first version in collection, is_highest is set to True.
     If this version is greater than the highest version in collection, set is_highest
     equals False on the last highest version and True on this version.
-    Otherwise does nothing.
+    Otherwise does nothing. The collection version is updated to the database if `save=True`,
+    otherwise only the `is_highest` field is updated on the instance. This is an optimization
+    to prevent double saving when this method is called during syncs and uploads.
     """
 
     def is_new_highest(new, old):
@@ -447,7 +449,7 @@ def _update_highest_version(collection_version, save=False):
                 collection_version.save(update_fields=["is_highest"])
             return
 
-        # compute highest from the whole list ...
+        # previous highest must have been removed, re-compute highest from the whole list ...
         highest = None
         for cv in collection_version.collection.versions.all():
             sv = Version(cv.version)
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -461,6 +461,11 @@ def is_new_highest(new, old):
 
     # exit if the new CV is not higher
     if not is_new_highest(Version(collection_version.version), Version(last_highest.version)):
+        # Ensure that this collection_version doesn't have is_highest improperly set
+        if collection_version.is_highest:
+            collection_version.is_highest = False
+            if save:
```
mdellweg, line 467: What is this save trying to tell us?

mdellweg (commented):
Content in Pulp is meant to be immutable.
`is_highest` does not even look right to me when a collection version is in two different repositories.

pulp_ansible/app/tasks/collections.py
gerrod3: What do you mean? I explained the bug in the issue. This `if save` is to just be consistent with how the function can be called.

gerrod3: That previous condition is for an odd scenario where none of the collection versions for a collection had `is_highest` set. The reason why `save` is conditional is for an optimization since this function is called to update a collection version before it is saved in the `CollectionSaver` stage or the `import_collection` task, there is no point in saving the collection twice in a row.

mdellweg (approved):
I'm really not happy, we ignore that content must be immutable.
What happens if we remove the uniqueness constraint on the `is_latest` and specifically allow it to be a bit fuzzy at times? A mutable field within a uniqueness constraint seems to be additionally painful. Also what is true for a collection in one repository version does not necessarily be as true in another. Could we move that information elsewhere?

gerrod3:
> I'm really not happy, we ignore that content must be immutable. What happens if we remove the uniqueness constraint on the `is_latest` and specifically allow it to be a bit fuzzy at times? A mutable field within a uniqueness constraint seems to be additionally painful. Also what is true for a collection in one repository version does not necessarily be as true in another. Could we move that information elsewhere?

This is how it has been for years... Yes we should refactor it to better be align with Pulp philosophy, not sure how though. For AAH it has been fine to have this field calculated on the model since they only cared about one repository, but with the introduction of repo management it might be time to look at this field again. I think we should file an issue for this and put the design discussion there.

mdellweg (approved):

mdellweg:
Manual cherry-pick:
https://github.com/pulp/pulp_ansible/pull/1984

Issue #1555, opened 2023-08-16
Katello: Pulp-To-Pulp sync ends in 403 error


ianballou:
**Version**
Katello versions:

```
  "versions": [
    {
      "component": "core",
      "version": "3.28.5",
      "package": "pulpcore",
      "domain_compatible": true
    },
    {
      "component": "rpm",
      "version": "3.22.3",
      "package": "pulp-rpm",
      "domain_compatible": true
    },
    {
      "component": "python",
      "version": "3.10.0",
      "package": "pulp-python",
      "domain_compatible": false
    },
    {
      "component": "ostree",
      "version": "2.1.1",
      "package": "pulp-ostree",
      "domain_compatible": false
    },
    {
      "component": "file",
      "version": "1.14.3",
      "package": "pulp-file",
      "domain_compatible": true
    },
    {
      "component": "deb",
      "version": "2.21.1",
      "package": "pulp_deb",
      "domain_compatible": false
    },
    {
      "component": "container",
      "version": "2.15.2",
      "package": "pulp-container",
      "domain_compatible": false
    },
    {
      "component": "certguard",
      "version": "1.6.5",
      "package": "pulp-certguard",
      "domain_compatible": true
    },
    {
      "component": "ansible",
      "version": "0.18.0",
      "package": "pulp-ansible",
      "domain_compatible": false
    }
  ],

```

Smart proxy 1 versions:

```
{
  "versions": [
    {
      "component": "core",
      "version": "3.28.10",
      "package": "pulpcore",
      "domain_compatible": true
    },
    {
      "component": "rpm",
      "version": "3.22.3",
      "package": "pulp-rpm",
      "domain_compatible": true
    },
    {
      "component": "python",
      "version": "3.10.0",
      "package": "pulp-python",
      "domain_compatible": false
    },
    {
      "component": "ostree",
      "version": "2.1.1",
      "package": "pulp-ostree",
      "domain_compatible": false
    },
    {
      "component": "file",
      "version": "1.14.3",
      "package": "pulp-file",
      "domain_compatible": true
    },
    {
      "component": "deb",
      "version": "2.21.1",
      "package": "pulp_deb",
      "domain_compatible": false
    },
    {
      "component": "container",
      "version": "2.15.2",
      "package": "pulp-container",
      "domain_compatible": false
    },
    {
      "component": "certguard",
      "version": "1.6.5",
      "package": "pulp-certguard",
      "domain_compatible": true
    },
    {
[...]

PR #1561, opened 2023-08-23
Add new settings for customizing Galaxy auth/authZ

gerrod3:
@newswangerd Katello needs the ability to turn off the Galaxy RBAC in order to sync since they don't use any RBAC at all. Would these new settings be acceptable with you?

@ianballou

fixes: #1555


ianballou:
Also cc @sjha4

mdellweg:
I do not see, how this change changes anything on it's own, but still we seem to get a 403 in the cli tests.
Can you make any sense of that?

gerrod3:
> I do not see, how this change changes anything on it's own, but still we seem to get a 403 in the cli tests. Can you make any sense of that?

If I am reading the error log correctly, it is failing to download the role from github. It is not even Pulp giving a 403...

mdellweg (approved):

PR #1570, opened 2023-09-14
Fix downloading testing content in docs build

mdellweg:
[noissue]


mdellweg:
The collection that was missing was reuploaded.

Issue #1608, opened 2023-10-04
Parsing CollectionVersion Metadata sync progress report does not report accurate total when using signed_only and remote has no signatures.


gerrod3:
**Version**
0.17+

**Describe the bug**
When using `signed_only=True` and a remote has no signatures, the progress report for the first stage `Parsing CollectionVersion Metadata` will show a total count of 0 at the end even if the remote is not empty. This is because the pipeline sees that the downloaded metadata for the CV does not have a signature and thus returns early without incrementing the done counter. At the end of the sync the total count is updated to the erroneous done count

PR #1609, opened 2023-10-04
Fix syncing total count when using signed_only=True

gerrod3:
fixes: #1608
https://issues.redhat.com/browse/AAH-2692

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -585,17 +585,23 @@ class CollectionSyncFirstStage(Stage):
         )
         cv_unique = attrgetter("namespace", "name", "version")(collection_version)
         fullname, version = f"{cv_unique[0]}.{cv_unique[1]}", cv_unique[2]
-        if fullname in self.exclude_info and Version(version) in self.exclude_info[fullname]:
-            return
         if cv_unique in self.already_synced:
             return
+
+        # Mark the collection version as being processed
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -576,11 +576,15 @@ async def _add_collection_version(self, api_version, collection_version_url, met
         )
         cv_unique = attrgetter("namespace", "name", "version")(collection_version)
         fullname, version = f"{cv_unique[0]}.{cv_unique[1]}", cv_unique[2]
-        if fullname in self.exclude_info and Version(version) in self.exclude_info[fullname]:
-            return
         if cv_unique in self.already_synced:
             return
+
+        # Mark the collection version as being processed
         self.already_synced.add(cv_unique)
```
mdellweg, line 583: What is this? Apart from eating a lot of memory?

mdellweg (approved):

pulp_ansible/app/tasks/collections.py
mdellweg: Do we want to skip this when doing a full mirror sync anyway?
What if we do a bloom filter thing?

(Just a few thoughts...)

mdellweg: Sorry, this was asking on a tangent. Would you mind making it a separate change?

mdellweg (approved):

PR #1610, opened 2023-10-05
Fix syncing tests around v2 api

gerrod3:
I think it is time for us to resume work on creating a local fixture to sync off of. Also, what should we do with the v2 tests now? I don't expect old-galaxy to stay around forever.


pulp_ansible/tests/functional/api/collection/v2/test_upload.py
```diff
@@ -15,6 +15,7 @@
 from pulp_ansible.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
 
 
+# TODO: Check if this url will stay around
```
mdellweg, line 18: At least we have some control around this one.
We could base all of our tests on squeezer...

pulp_ansible/tests/functional/api/collection/v2/test_sync.py
```diff
@@ -14,13 +14,14 @@
 from pulp_smash.pulp3.bindings import monitor_task, delete_orphans
 
 
+# TODO: Rewrite or delete?
```
mdellweg, line 17: So you are saying we might consider removing the v2 tests completely instead of rewriting them in pytest style?

mdellweg (approved):

pulp_ansible/tests/functional/api/collection/v2/test_upload.py
gerrod3: Potentially, there are some tests that are designed for oddball collections that squeezer wouldn't be a good fit for.

pulp_ansible/tests/functional/api/collection/v2/test_sync.py
gerrod3: Pretty much, considering the new galaxy doesn't support the v2 api anymore, it might not be worth it to keep the tests around if v2 isn't used anywhere.

Issue #1612, opened 2023-10-11
Add `count` attribute to `TagSerializer`


jerabekjiri:
**Is your feature request related to a problem? Please describe.**

Old galaxy (https://old-galaxy.ansible.com/search) has a list of popular tags. Since we're reusing `TagSerializer` it would be useful to add this to `TagSerializer` in pulp_ansible.
![Screenshot from 2023-10-10 13-29-31](https://github.com/pulp/pulp_ansible/assets/19647757/4c690fa5-cf51-4d91-b15b-60f07346cb08)

**Describe the solution you'd like**
Update `TagSerializer` with
```
    count = serializers.SerializerMethodField()

    def get_count(self, obj):
        return obj.ansible_collectionversion.count()

    class Meta:
        model = Tag
        fields = ["name", "count"] 
```

**Describe alternatives you've considered**
If this doesn't work out, we can add it to galaxy_ng codebase.

mdellweg:
sounds feasible...

PR #1621, opened 2023-10-16
Add `count` to `TagSerializer`

jerabekjiri:
fixes: #1612

### DIFF
--- a/pulp_ansible/app/serializers.py
+++ b/pulp_ansible/app/serializers.py
@@ -385,9 +385,14 @@ class TagSerializer(serializers.ModelSerializer):
     A serializer for the Tag model.
     """
 
+    count = serializers.SerializerMethodField()
+
+    def get_count(self, obj):
+        return obj.ansible_collectionversion.count()
+
     class Meta:
         model = Tag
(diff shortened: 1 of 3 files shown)


mdellweg (approved):

gerrod3:
I would suggest testing this change first on a large system. You are adding N db queries to each tag list call with this change.

mdellweg:
> I would suggest testing this change first on a large system. You are adding N db queries to each tag list call with this change.

Right. Can we use some annotation logic?

pulp_ansible/app/serializers.py
```diff
@@ -385,9 +385,14 @@ class TagSerializer(serializers.ModelSerializer):
     A serializer for the Tag model.
     """
 
+    count = serializers.SerializerMethodField()
+
+    def get_count(self, obj):
+        return obj.count
```
mdellweg, line 391: I think it knows how to do that now... (Please try removing it).

You need to change the SerializerMethodField for something else...

jerabekjiri: changed to integer field

Issue #1623, opened 2023-10-20
`is_highest` update query can still fail since Django doesn't order the query for Postgres


gerrod3:
**Version**
<=0.20.1

**Describe the bug**
https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/tasks/collections.py#L432-L434
This `order_by` has no effect on the following `update` query allowing a possibility of hitting a unique constraint violation if the `new_is_highest` collection-version is not last in the natural ordering of the query. This behavior is noted in the django docs (https://docs.djangoproject.com/en/4.2/ref/models/querysets/#ordered-queryset).

PR #1624, opened 2023-10-20
Fix update_highest_version when versions were uploaded out-of-order

gerrod3:
fixes: #1623


jctanner (approved):

jctanner:
@gerrod3 is there more to do before this is merged and released?

PR #1636, opened 2023-11-03
Remove pulp-smash test dependency

mdellweg:
[noissue]


gerrod3 (approved):
Hurray!!! :partying_face:

PR #1701, opened 2023-12-08
Remove api v2

mdellweg:
This will become a larger piece of work...


mdellweg:
If something breaks, I'll blame it on insufficient test coverage. ;)

sivel:
FWIW, the ansible core team is testing against the v2 API within pulp-ansible, and if this is removed, it also removes our ability to update our CI to test against newer pulp/galaxy-ng versions, until such a time as we decide to remove all of the v2 code from core, which is as of yet unscheduled.

sivel:
> Can you tell me more about this? 

We use the `quay.io/pulp/galaxy` image within our CI.  For v2 testing, we are hitting `http://{pulp_host}/pulp_ansible/galaxy/`.  

> So you are not testing that pulp_ansible can sync from v2, but that clients can download from v2, right?

That is correct

>  Is this testing against a blank pulp_ansible, or behind galaxy?

The testing is happening within the galaxy container, but we interact with pulp_ansible separately.  Basically using 1 container to access both API versions.

> Would it be possible to select an old version of pulp_ansible?

Theoretically/technically, yes.  Although it would require a fair amount of re-working in our tests.  We would have to then spin up and manage 2 different containers to do it, which the tests aren't currently capable of really handling. The tests are rather complex, I'd say it would take 4 weeks or so to rewrite the tests to handle this.

gerrod3 (commented):
Do I need to wait on anything in ansible land before I can approve this?

mdellweg:
> Do I need to wait on anything in ansible land before I can approve this?

I believe we have everyone on board by now.

gerrod3 (approved):

PR #1437, opened 2023-05-05
Namespace compatibility changes

newswangerd:
Added:
- `/v3/plugin/ansible/search/namespace-metadata/` -> provides cross repo search functionality for namespace metadata
- `/pulp/api/v3/pulp_ansible/namespaces/` -> API for creating, updating permissions and viewing pulp ansible global namespace

Changed:
- `/v3/namespaces/` -> make the API backwards compatible with galaxy_ng

Jira ticket: https://issues.redhat.com/browse/AAH-2293 (github issue comming soon)

Todo:
- [x] Add filtersets on new APIs
- [x] Tests
- [ ] RBAC
- [ ] v3 compatiblity
    - [ ] error messages
    - [x] filters
    - [ ] allow saving an image with an avatar URL
    - [x] save namespace metadata before launching task to add it to repository

PR #1713, opened 2024-01-04
Namespace compatibility changes

rochacbruno:
Trying to rework #1437 

https://issues.redhat.com/browse/AAH-2293

### DIFF
--- a/pulp_ansible/app/custom_fields.py
+++ b/pulp_ansible/app/custom_fields.py
@@ -1,0 +1,209 @@
+import json
+
+from collections import OrderedDict
+
+from django.utils.translation import gettext_lazy as _
+from django.core.exceptions import BadRequest
+
+from drf_spectacular.utils import extend_schema_field
+
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+
+from pulpcore.plugin.models.role import Role
+
+from pulpcore.plugin.util import (
+    assign_role,
+    remove_role,
+    get_groups_with_perms_attached_roles,
+    get_perms_for_model,
+)
+from pulpcore.plugin.models import Group
+
+
+class RelatedFieldsBaseSerializer(serializers.Serializer):
+    """
+    Serializer only returns fields specified in 'include_related' query param.
+
+    This allows for fields that require more database queries to be optionally
+    included in API responses, which lowers the load on the backend. This is
+    intended as a way to include extra data in list views.
+
+    Usage:
+
+    This functions the same as DRF's base `serializers.Serializer` class with the
+    exception that it will only return fields specified in the `?include_related=`
+    query parameter.
+
+    Example:
+
+    MySerializer(RelatedFieldsBaseSerializer):
+        foo = CharField()
+        bar = CharField()
+
+    MySerializer will return:
+
+    {"foo": None} when called with `?include_related=foo` and {"foo": None, "bar" None}
+    when called with `?include_related=foo&include_related=bar`.
+    """
+
+    def __init__(self, *args, **kwargs):
+        # This should only be included as a sub serializer and shouldn't be used for
+        # updating objects, so set read_only to true
+        kwargs["read_only"] = True
+        return super().__init__(*args, **kwargs)
+
+    def to_representation(self, instance):
+        result = OrderedDict()
+        fields = self._readable_fields
+        request = self.context.get("request", None)
+        if request:
+            # TODO: Figure out how to make `include_related` show up in the open API spec
+            include_fields = request.GET.getlist("include_related")
+
+            if len(include_fields) > 0:
+                for field in fields:
+                    if field.field_name in include_fields:
+                        result[field.field_name] = field.to_representation(instance)
+        else:
+            # When there is no request present, it usually means that the serializer is
+            # being inspected by DRF spectacular to generate the open API spec. In that
+            # case, this should act like a normal serializer.
+            return super().to_representation(instance)
+
+        return result
+
+
+class MyPermissionsField(serializers.Serializer):
+    def to_representation(self, original_obj):
+        request = self.context.get("request", None)
+        if request is None:
+            return []
+        user = request.user
+
+        if original_obj._meta.proxy:
+            obj = original_obj._meta.concrete_model.objects.get(pk=original_obj.pk)
+        else:
+            obj = original_obj
+
+        my_perms = []
+        for perm in get_perms_for_model(type(obj)).all():
+            codename = "{}.{}".format(perm.content_type.app_label, perm.codename)
+            if user.has_perm(codename) or user.has_perm(codename, obj):
+                my_perms.append(codename)
+
+        return my_perms
+
+
+def set_object_group_roles(obj, new_groups):
+    current_groups = get_groups_with_perms_attached_roles(obj, include_model_permissions=False)
+
+    # remove the old roles
+    for group in current_groups:
+        if group in new_groups:
+            new_groups.pop(group)
+            continue
+        for role in current_groups[group]:
+            remove_role(role, group, obj)
+
+    # add the new ones
+    for group in new_groups:
+        for role in new_groups[group]:
+            try:
+                assign_role(role, group, obj)
+            except BadRequest:
+                raise ValidationError(
+                    detail={
+                        "groups": _(
+                            "Role {role} does not exist or does not "
+                            "have any permissions related to this object."
+                        ).format(role=role)
+                    }
+                )
+
+
+class GroupFieldType(serializers.Serializer):
+    """
+    Group open api type
+    """
+
+    name = serializers.CharField(max_length=256, allow_blank=False)
+    id = serializers.CharField(max_length=256, allow_blank=True, required=False)
+    object_roles = serializers.ListField(child=serializers.CharField())
+
+
+@extend_schema_field(GroupFieldType(many=True))
+class GroupPermissionField(serializers.Field):
+    def get_value(self, dictionary):
+        data = dictionary.get(self.field_name, [])
+
+        # The open api client sends data as a form request rather than json
+        # because of the avatar URL. It converts a list like
+        # [{"foo": "bar"}, {"bar": "foo"}] into "{'foo': 'bar'}, {'bar': 'foo'}"
+        # This is a best effort attempt to capture that data and convert it into
+        # valid JSON and then transform it into a list that the API can understand
+        if isinstance(data, str):
+            try:
+                data = f"[{data}]".replace("'", '"')
+                return json.loads(data)
+            except json.decoder.JSONDecodeError:
+                raise ValidationError(detail={"links": "Must be valid JSON"})
+
+        return super().get_value(dictionary)
+
+    def _validate_group(self, group_data):
+        if "object_roles" not in group_data:
+            raise ValidationError(detail={"groups": _("object_roles field is required")})
+
+        if "id" not in group_data and "name" not in group_data:
+            raise ValidationError(detail={"groups": _("id or name field is required")})
+
+        roles = group_data["object_roles"]
+
+        if not isinstance(roles, list):
+            raise ValidationError(detail={"groups": _("object_roles must be a list of strings")})
+
+        # validate that the permissions exist
+        for role in roles:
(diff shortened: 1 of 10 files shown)


pulp_ansible/app/custom_fields.py
```diff
@@ -0,0 +1,209 @@
...
+from pulpcore.plugin.util import (
+    assign_role,
+    remove_role,
+    get_groups_with_perms_attached_roles,
+    get_perms_for_model,
+)
+from pulpcore.plugin.models import Group
+
+
+class RelatedFieldsBaseSerializer(serializers.Serializer):
```
mdellweg, line 24: Is this a reinvention of drf_queryfields?

pulp_ansible/app/custom_fields.py
```diff
@@ -0,0 +1,209 @@
...
+            # When there is no request present, it usually means that the serializer is
+            # being inspected by DRF spectacular to generate the open API spec. In that
+            # case, this should act like a normal serializer.
+            return super().to_representation(instance)
+
+        return result
+
+
+class MyPermissionsField(serializers.Serializer):
+    def to_representation(self, original_obj):
```
mdellweg, line 78: This has tremendous resemblance with the `my_permissions` action in our `pulpcore.app.viewsets.base.RolesMixin`.

```python
    @extend_schema(
        summary="List user permissions",
        description="List permissions available to the current user on this object.",
        responses={
            200: inline_serializer(
                name="MyPermissionsSerializer", fields={"permissions": ListField(child=CharField())}
            )
        },
    )
    @action(detail=True, methods=["get"])
    def my_permissions(self, request, pk=None):
        obj = self.get_object()
        app_label = obj._meta.app_label
        permissions = [
            ".".join((app_label, codename)) for codename in request.user.get_all_permissions(obj)
        ]
        return Response({"permissions": permissions})
```

pulp_ansible/app/custom_fields.py
rochacbruno: I am not aware about drf_queryfields, but this is the pattern we use in other places in galaxy API

rochacbruno:
Hi @mdellweg do you think any changes are required on this PR?

pulp_ansible/app/viewsets.py
```diff
@@ -376,6 +391,16 @@ class Meta:
             "metadata_sha256": ["exact", "in"],
         }
 
+    def no_op(self, queryset, name, value):
+        return queryset
+
+    def has_permissions(self, queryset, name, value):
+        perms = self.request.query_params.getlist(name)
```
mdellweg, line 398: How is that different from `value`?

pulp_ansible/app/viewsets.py
```diff
@@ -368,6 +377,12 @@ class AnsibleNamespaceFilter(ContentFilter):
     A filter for namespaces.
     """
 
+    # this is a hack to keep pulp from throwing a 400 when an unknwon
+    # query param is included, since filtering is handled by the related
+    # field serializer
+    include_related = filters.CharFilter(method="no_op")
+    my_permissions = filters.CharFilter(method="has_permissions")
+
```
mdellweg, line 385: This part I don't get. These are pulp api views.
Also we don't break down permissions to the content. I believe we should have the `RolesMixin` in the `AnsibleGlobalNamespaceViewSet` instead.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -799,50 +798,34 @@ def _create(self, request, data, context=None):
...
-                    artifact = Artifact.objects.get(sha256=artifact.sha256)
-                    artifact.touch()
-                except (Artifact.DoesNotExist, DatabaseError):
-                    # the artifact has since been removed from when we first attempted to save it
-                    artifact.save()
-            context["artifact"] = artifact.pk
-
-        # Dispatch general_create task
-        app_label = AnsibleNamespaceMetadata._meta.app_label
+        c = serializer.save()
```
mdellweg, line 801: `c` is not such a good variable name.

pulp_ansible/app/custom_fields.py
mdellweg: Can you at least consider using

```python
        permissions = [
            ".".join((app_label, codename)) for codename in request.user.get_all_permissions(obj)
        ]
```

pulp_ansible/app/custom_fields.py
mdellweg: So this is a bit of code copied from galaxy_ng?
Let me rephrase: Why is it called `include_related`? All I see is, it can select fields on the serializer to be included or not. And that is exactly what drf_queryfields (a pulp dependency for about forever) does.

pulp_ansible/app/serializers.py
```diff
@@ -808,11 +818,38 @@ def to_representation(self, value):
...
+
+
+class AnsibleNamespaceMetadataSerializer(NoArtifactContentSerializer, NamespaceValidationMixin):
     """
     A serializer for Namespaces.
     """
 
+    # Task on the serializer will always return none. It's just here to prevent the bindings
+    # from breaking.
+    task = serializers.SerializerMethodField()
```
mdellweg, line 850: Why would the bindings break?
We generate them for each version individually anyway.
I'd rather not do such things.

mdellweg (commented):
I left some suggestions/questions.
Still not quite confident to understand every aspect of this large PR.

rochacbruno:
@newswangerd can you help me answer @mdellweg questions as you are the original author of this code?

pulp_ansible/app/custom_fields.py
rochacbruno: @mdellweg I talked to David and here is the explanation:

include_related controls what is included in the specific field that is based on RelatedFieldsSerializer, this is not to control the regular fields from the serializer, but to control which fields are included nested in a specialized field, the name is `include_related` because it is always data coming from related models.

So instead of always fetching the related models we do that only when is explicitly specified in the query param.

pulp_ansible/app/custom_fields.py
rochacbruno: It is implemented this way to be included in the ListView, while the one from RolesMixin is for DEtailView only

pulp_ansible/app/viewsets.py
rochacbruno: Looks like it needs to be taken using getlist() so it returns a Python list type.

pulp_ansible/app/viewsets.py
rochacbruno: Those fields must be declared, otherwise we are not able to pass those fields as query params.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -1458,8 +1458,7 @@ def test_cross_repo_search_index_with_updated_namespace_metadata(
     task = galaxy_v3_plugin_namespaces_api_client.create(
         name=namespace_name, description="hello", company="Testing Co.", **distro_kwargs
     )
-    result = monitor_task(task.task)
-    namespace_href = [x for x in result.created_resources if "namespaces" in x][0]
+    namespace_href = task.pulp_href
```
mdellweg, line 1461: Is this api no longer using tasks? Then we should rename `task` to `namespace`.

pulp_ansible/tests/functional/api/collection/v3/test_collection_version_search.py
```diff
@@ -1489,8 +1488,7 @@ def test_cross_repo_search_index_with_updated_namespace_metadata(
         avatar=avatar_path,
         **distro_kwargs,
     )
-    result2 = monitor_task(task2.task)
-    namespace2_href = [x for x in result2.created_resources if "namespaces" in x][0]
+    namespace2_href = task2.pulp_href
```
mdellweg, line 1491: Here too.

pulp_ansible/tests/functional/api/collection/v3/test_namespace.py
```diff
@@ -41,6 +44,336 @@ def _random_image():
...
+        "2cool4school",
+        "x",
+        "xy",
+        "my-namespace-name",
+        "my.namespace.name",
+        "hello!",
+    ]
+
+    for name in invalid_names:
+        try:
```
mdellweg, line 69: This should use `with pytest.raises`

pulp_ansible/tests/functional/api/collection/v3/test_namespace.py
```diff
@@ -41,6 +44,336 @@ def _random_image():
...
+):
+    """
+    Test that the legacy groups field works.
+    """
+
+    group_1 = random_string()
+    group_2 = random_string()
+
+    groups_api_client.create({"name": group_1})
+    groups_api_client.create({"name": group_2})
```
mdellweg, line 102: can we use gen_object_with_cleanup?

pulp_ansible/tests/functional/api/collection/v3/test_namespace.py
```diff
@@ -41,6 +44,336 @@ def _random_image():
...
+
+    repo, distro = ansible_repo_and_distro_factory()
+    kwargs = {"path": distro.base_path, "distro_base_path": distro.base_path}
+
+    # Create namespace with no groups
+    name = random_string()
+    result = galaxy_v3_plugin_namespaces_api_client.create(name=name, **kwargs)
+    monitor_task(result.task)
```
mdellweg, line 110: I'm confused now. Is this a created with a task, or no?

mdellweg: Also we probably want it cleaned up.

mdellweg (commented):
Still not a full review... Sorry.

pulp_ansible/tests/functional/api/collection/v3/test_namespace.py
rochacbruno: pytest.raises allows only a single exception and this test is checking for ApiException and also ApiValueError.

I am not sure why (I have not written the test) but looks like the API can raise one of those errors, I guess it will depend on the input.

I dont see any advantage of using pytst.raises in this case.

mdellweg: Hmm, I'll trust you on that.
But for the record, i believe, pytest.raises is more readable and it marks a test failed, not errored.
Do we really not know which one Exception we are to expect here?

pulp_ansible/app/custom_fields.py
```diff
@@ -0,0 +1,209 @@
...
+        request = self.context.get("request", None)
+        if request is None:
+            return []
+        user = request.user
+
+        if original_obj._meta.proxy:
+            obj = original_obj._meta.concrete_model.objects.get(pk=original_obj.pk)
+        else:
+            obj = original_obj
```
mdellweg, line 87: I'm pretty sure, in Pulp we use permissions on the proxy model is that different in Galaxy?

(I think the Group is the only proxy model we use, and we needed that for additional permissions).

pulp_ansible/app/custom_fields.py
```diff
@@ -0,0 +1,209 @@
...
+                            "Role {role} does not exist or does not "
+                            "have any permissions related to this object."
+                        ).format(role=role)
+                    }
+                )
+
+
+class GroupFieldType(serializers.Serializer):
+    """
+    Group open api type
```
mdellweg, line 127: This is more a nested serializer than an openapi type.

pulp_ansible/app/custom_fields.py
```diff
@@ -0,0 +1,209 @@
...
+        data = dictionary.get(self.field_name, [])
+
+        # The open api client sends data as a form request rather than json
+        # because of the avatar URL. It converts a list like
+        # [{"foo": "bar"}, {"bar": "foo"}] into "{'foo': 'bar'}, {'bar': 'foo'}"
+        # This is a best effort attempt to capture that data and convert it into
+        # valid JSON and then transform it into a list that the API can understand
+        if isinstance(data, str):
+            try:
+                data = f"[{data}]".replace("'", '"')
```
mdellweg, line 147:
```suggestion
                data = f"[{data.replace("'", '"')}]"
```

Issue #1754, opened 2024-02-15
Convert docs to new stack


pedro-psb:
The general workflow for this task can be found [here](https://hackmd.io/@pbrochad/SyfGeQc86#Content-Migration).

PR #1766, opened 2024-02-22
Initial conversion to new style docs

gerrod3:
[noissue]

### DIFF
--- a/staging_docs/admin/reference/settings.md
+++ b/staging_docs/admin/reference/settings.md
@@ -1,0 +1,71 @@
+# Settings
+
+`pulp_ansible` provides a few settings to control various features. These settings are settable
+with dynaconf in various as a regular Pulp setting. See the [pulpcore Setting](https://docs.pulpproject.org/en/3.0/nightly/installation/configuration.html#configuration).
+
+## ANSIBLE_API_HOSTNAME
+
+> The origin, e.g. "<http://example.com>" that will instruct the client how to find the Pulp API
+> service. This URL is formed in various Galaxy APIs (V1, V2, V3) responses.
+>
+> This defaults to http on your fqdn, which is usable with the Ansible installer's default Nginx
+> configuration. So if `example.com` is your fqdn, this would default to "<http://example.com>".
+
+## ANSIBLE_CONTENT_HOSTNAME
+
+> The origin, e.g. "<http://example.com>" that will instruct the client how to find the Pulp content
+> app. This URL is formed in various Galaxy APIs (V1, V2, V3) responses.
+>
+> This defaults to [CONTENT_ORIGIN](https://docs.pulpproject.org/pulpcore/settings.html?#content-origin).
+> By default it includes the `pulp/content` subpath. So if `https://example.com` is your
+> CONTENT_ORIGIN, this would default to "<https://example.com/pulp/content>".
+
+## ANSIBLE_SIGNATURE_REQUIRE_VERIFICATION
+
+> By default, `pulp_ansible` rejects uploaded signatures if they cannot be verified against a
+> public key specified on the repository. Setting this to false will allow accepting signatures
+> if no key was specified. A repository with a configured key will always reject invalid
+> signatures.
+
+## ANSIBLE_SIGNING_TASK_LIMITER
+
+> This number determines the amount of concurrent signing processes that can be spawned at one time
+> during a signature task. Increasing this number will generally increase the speed of the task, but
+> will also consume more resources of the worker. Defaults to 10 concurrent processes.
+
+## GALAXY_API_ROOT
+
+> By default the Galaxy V1, V2, and V3 APIs are rooted at
+> "/pulp_ansible/galaxy/\<path:path>/api/", but this is configurable. Specifying `GALAXY_API_ROOT`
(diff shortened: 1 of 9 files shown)


staging_docs/index.md
pedro-psb, line 1: I'll recommend moving to `staging_docs/user/index.md`.

staging_docs/user/guides/copy.md
```diff
@@ -0,0 +1,199 @@
...
+http POST http://example.com/pulp/api/v3/repositories/ansible/ansible/05813fa6-cf0b-435b-b54b-5a30fc370848/modify/ \
+    add_content_units:="['/pulp/api/v3/content/ansible/collections/d2bab58c-50f2-4f1d-9cf0-8ceb1680f31b/']"
+```
+
+And you list the dest_repo via the Galaxy V3 API (that the CLI uses) you would see
+`deprecated=False`.
+
+However you could instead use the `copy` API as follows:
+
+```sh
```
ggainey, line 115: bash

staging_docs/user/guides/copy.md
```diff
@@ -0,0 +1,199 @@
...
+}
+```
+
+## Copy All Content
+
+When calling `copy` you can omit the `content` argument and `copy` will copy all content from
+the `source_repo_version` to a new Repository Version created on the `dest_repo`. That would be
+similar to:
+
+```sh
```
ggainey, line 167: bash

staging_docs/user/guides/copy.md
```diff
@@ -0,0 +1,199 @@
...
+]
+```
+
+## Specifying a Destination Base Version
+
+In some situations, you may want the content being copied to be applied not to the latest version
+of the `dest_repo`, and in that case, you can additionally specify the `dest_base_version` and
+that would be used instead of the latest RepositoryVersion of `dest_repo`:
+
+```sh
```
ggainey, line 180: bash

staging_docs/user/guides/copy.md
```diff
@@ -0,0 +1,199 @@
...
+]
+```
+
+## Multi-Repo Copy
+
+You can specify a more complicated `config` option which can express multiple copy operations in
+one call. Each entry is the dictionary of `source_repo_version`, `dest_repo`, and optional
+`content`, in a list form.
+
+```sh
```
ggainey, line 193: bash

pedro-psb (changes requested):

staging_docs/user/guides/roles.md
```diff
@@ -0,0 +1,220 @@
...
+Create a repository and name it the `foo` repository.
+
+```bash
+# Start by creating a new repository named foo
+pulp ansible repository create --name "foo"
+```
+
+Repository create output:
+
+```
```
ggainey, line 23: json

staging_docs/user/guides/roles.md
```diff
@@ -0,0 +1,220 @@
...
+`/etc/ansible/ansible.cfg`):
+
+```
+[galaxy]
+server: http://pulp.example.com/pulp_ansible/galaxy/my_content/
+```
+
+Then you can install without the `-s` url
+
+```
```
ggainey, line 214: bash

staging_docs/user/guides/signature.md
```diff
@@ -0,0 +1,105 @@
...
+validation when installing Collections with `ansible-galaxy`. Check out the workflows below to see
+how to add signature support.
+
+## Setup
+
+In order to verify signature validity on uploads you will need to store your trusted key on the
+repositories `gpgkey` attribute.
+
+!!! note
+You can upload signatures without supplying Pulp any key, but `pulp_ansible` will not
```
ggainey, line 13: indent 4

ggainey (changes requested):

pedro-psb:
Oh, and closes https://github.com/pulp/pulp_ansible/issues/1754

pedro-psb (approved):
LGTM

staging_docs/admin/reference/settings.md
```diff
@@ -0,0 +1,71 @@
+# Settings
+
+`pulp_ansible` provides a few settings to control various features. These settings are settable
+with dynaconf in various as a regular Pulp setting. See the [pulpcore Setting](https://docs.pulpproject.org/en/3.0/nightly/installation/configuration.html#configuration).
```
mdellweg, line 4: I do not like this line break policy. It is not helpful and presents a sizeable entry barrier for contributors. I cannot even check it in a meaningful way in a PR.

gerrod3: I don't understand what you are saying. Are you talking about the number of characters allowed per line? I thought that was just a part of the lint rules.

mdellweg: I'm talking about the fact that linebreaks keep happening in the middle of a sentence (not at syntactical breakpoints). And then when something like a URL comes into play, we simply abandon the rule to break at 100 c.

ggainey: While I think this is worth a discussion - it's not the goal of *this* PR, which is "get the docs into the new docs-stack". We'll have plenty of cleanup to do after that :)

ggainey (approved):

PR #1769, opened 2024-02-26
Fix namespace sync with missing avatar_sha256

mdellweg:
[noissue]

### DIFF
--- a/functest_requirements.txt
+++ b/functest_requirements.txt
@@ -7,3 +7,4 @@ numpy
 requests
 proxy.py~=2.4.3
 trustme~=1.1.0
+pulp-smash @ git+https://github.com/pulp/pulp-smash.git
--- a/pulp_ansible/app/models.py
+++ b/pulp_ansible/app/models.py
@@ -332,16 +332,9 @@ class AnsibleNamespaceMetadata(Content):
 
     @property
     def avatar_artifact(self):
-        if self.avatar_sha256:
-            avatar = self._artifacts.filter(sha256=self.avatar_sha256).first()
-            if avatar is None:
-                log.debug(
-                    f"Artifact({self.avatar_sha256}) is missing for namespace avatar "
-                    f"{self.name}:{self.metadata_sha256}"
-                )
-            return avatar
-
-        return None
+        return self._artifacts.filter(
+            content_memberships__relative_path=f"{self.name}-avatar"
+        ).first()
 
     @hook(BEFORE_SAVE)
     def calculate_metadata_sha256(self):
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -1154,13 +1154,12 @@ class AnsibleContentSaver(ContentSaver):
                 d_content.content.namespace = namespace
                 if d_content.d_artifacts:
                     da = d_content.d_artifacts[0]
-                    # Check to see if avatar failed to download, update metadata if so
+                    # Check to see if avatar failed to download, update metadata if so,
+                    # so that the avatar should be attemtped to be downloaded again.
                     if da.deferred_download:
                         d_content.d_artifacts = None
                         d_content.content.avatar_sha256 = None
-                    # Check to see if upstream didn't have avatar_sha256 set
-                    elif d_content.content.avatar_sha256 is None:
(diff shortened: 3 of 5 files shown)


pulp_ansible/app/models.py
```diff
@@ -332,15 +332,9 @@ class AnsibleNamespaceMetadata(Content):
...
-        if self.avatar_sha256:
-            avatar = self._artifacts.filter(sha256=self.avatar_sha256).first()
-            if avatar is None:
-                log.debug(
-                    f"Artifact({self.avatar_sha256}) is missing for namespace avatar "
-                    f"{self.name}:{self.metadata_sha256}"
-                )
-            return avatar
-
+        ca = self.contentartifact_set.filter(relative_path=f"{self.name}-avatar").first()
```
gerrod3, line 335: Can you check if we need a `select_related('artifact')`  here?

pulp_ansible/app/models.py
```diff
@@ -332,16 +332,9 @@ class AnsibleNamespaceMetadata(Content):
...
-                log.debug(
-                    f"Artifact({self.avatar_sha256}) is missing for namespace avatar "
-                    f"{self.name}:{self.metadata_sha256}"
-                )
-            return avatar
-
-        return None
+        return self._artifacts.filter(
+            content_memberships__relative_path="hbvgfzyzud-avatar"
+        ).first()
```
mdellweg, line 337: I believe this actually produces the proper inner join right away.

```sql
SELECT "core_artifact"."pulp_id", "core_artifact"."pulp_created", "core_artifact"."pulp_last_updated", "core_artifact"."file", "core_artifact"."size", "core_artifact"."md5", "core_artifact"."sha1", "core_artifact"."sha224", "core_artifact"."sha256", "core_artifact"."sha384", "core_artifact"."sha512", "core_artifact"."timestamp_of_interest", "core_artifact"."pulp_domain_id"
FROM "core_artifact"
INNER JOIN "core_contentartifact" ON ("core_artifact"."pulp_id" = "core_contentartifact"."artifact_id")
WHERE ("core_contentartifact"."content_id" = %s AND "core_contentartifact"."relative_path" = %s)
```

pulp_ansible/app/models.py
mdellweg: I updated the query to be about artifacts without indirection.

pulp_ansible/app/tasks/collections.py
```diff
@@ -1154,13 +1154,15 @@ def _pre_save(self, batch):
...
-                        d_content.content.avatar_sha256 = da.artifact.sha256
+                    if False:
+                        # This code invalidates the digest.
+                        # Check to see if avatar failed to download, update metadata if so
+                        if da.deferred_download:
+                            d_content.d_artifacts = None
+                            d_content.content.avatar_sha256 = None
+                        # Check to see if upstream didn't have avatar_sha256 set
+                        elif d_content.content.avatar_sha256 is None:
+                            d_content.content.avatar_sha256 = da.artifact.sha256
```
gerrod3, line 1165:
```suggestion
                      # Check to see if avatar failed to download, update metadata if so
                      if da.deferred_download:
                          d_content.d_artifacts = None
                          d_content.content.avatar_sha256 = None
                          d_content.content.metadata_sha256 = None
```
How about this? If `da.deferred_download` is true that means the avatar failed to download. We null the `avatar_sha256` and `metadata_sha256` (<-this gets recalculated on save), that way on next sync we will try download the image again.

mdellweg: Interesting idea. Let's try.

gerrod3 (approved):

functest_requirements.txt
```diff
@@ -7,3 +7,4 @@ numpy
 requests
 proxy.py~=2.4.3
 trustme~=1.1.0
+pulp-smash @ git+https://github.com/pulp/pulp-smash.git
```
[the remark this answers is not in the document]
mdellweg: Good question. It was some import. I should have checked whether it came from another plugins pytestplugin.
(This is all a mess.)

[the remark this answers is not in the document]
mdellweg: pulp_ansible/tests/functional/utils.py is smash infested...

PR #1815, opened 2024-04-15
Add sha256 uniqueness to CollectionVersion

mdellweg:


pulp_ansible/tests/functional/api/role/test_crud_distribution.py
```diff
@@ -76,4 +76,7 @@ def test_crud_distribution(
     monitor_task(ansible_distro_api_client.delete(distribution.pulp_href).task)
     with pytest.raises(ApiException) as exc_info:
         ansible_distro_api_client.delete(distribution.pulp_href)
-    assert "Not found." in exc_info.value.body
+    assert (
+        "Not found." in exc_info.value.body
+        or "No AnsibleDistribution matches the given query." in exc_info.value.body
+    )
```
mdellweg, line 82: Why on earth do i see a different error message when running this locally?

pulp_ansible/app/tasks/collections.py
```diff
@@ -100,41 +100,27 @@ async def declarative_content_from_git_repo(remote, url, git_ref=None, metadata_
...
+    artifact = Artifact.init_and_validate(artifact_path)
     if metadata_only:
         metadata["artifact"] = None
         metadata["artifact_url"] = None
     else:
-        artifact = Artifact.init_and_validate(artifact_path)
-        try:
-            await sync_to_async(artifact.save)()
-        except IntegrityError:
-            artifact = Artifact.objects.get(sha256=artifact.sha256)
```
mdellweg, line 111: This line would never have worked in an async context.

pulp_ansible/app/tasks/collections.py
```diff
@@ -303,50 +294,55 @@ def import_collection(
...
+    # the importer returns many None values. We need to let the defaults in the model prevail
+    for key in ["description", "documentation", "homepage", "issues", "repository"]:
+        if collection_info[key] is None:
+            collection_info.pop(key)
 
-        # the importer returns many None values. We need to let the defaults in the model prevail
-        for key in ["description", "documentation", "homepage", "issues", "repository"]:
-            if collection_info[key] is None:
-                collection_info.pop(key)
+    collection, created = Collection.objects.get_or_create(
```
gerrod3, line 313: Isn't this a database operation? This breaks your comment about `metadata_only=True`.

pulp_ansible/app/tasks/collections.py
```diff
@@ -625,6 +621,7 @@ async def _add_collection_version(self, api_version, collection_version_url, met
             setattr(collection_version, attr_name, attr_value)
 
         artifact = metadata["artifact"]
+        collection_version.sha256 = artifact["sha256"]
```
gerrod3, line 624: Wrong sha256.

pulp_ansible/app/tasks/collections.py
```diff
@@ -1179,12 +1175,17 @@ def _post_save(self, batch):
...
 
             for d_artifact in d_content.d_artifacts:
                 artifact = d_artifact.artifact
-                with artifact.file.open() as artifact_file, tarfile.open(
-                    fileobj=artifact_file, mode="r"
-                ) as tar:
+                # TODO change logic when implementing normal on-demand syncing
+                # Special Case for Git sync w/ metadata_only=True
+                if artifact_file_name := d_artifact_files.get(d_artifact):
+                    artifact_file = open(artifact_file_name, mode="rb")
```
gerrod3, line 1185: I don't understand this logic. Can you help explain it?

pulp_ansible/app/tasks/upload.py
```diff
@@ -46,6 +46,7 @@ def process_collection_artifact(artifact, namespace, name, version):
 
     # Set CollectionVersion metadata
     collection_info = importer_result["metadata"]
+    collection_info["sha256"] = artifact.sha256
```
gerrod3, line 49: Wrong sha256.

gerrod3 (changes requested):
I know this is based on my original work on this, but there was discussion back then on how the sha256 for the Collection should be the digest of the `MANIFEST.json` file inside the tarball. I actually never got around to finishing all of the changes to my PR, so most of my comments here is me pointing out all the areas I didn't change it.

https://github.com/pulp/pulp_ansible/pull/1119#discussion_r931946478

mdellweg:
> I know this is based on my original work on this, but there was discussion back then on how the sha256 for the Collection should be the digest of the `MANIFEST.json` file inside the tarball. I actually never got around to finishing all of the changes to my PR, so most of my comments here is me pointing out all the areas I didn't change it.
> 
> [#1119 (comment)](https://github.com/pulp/pulp_ansible/pull/1119#discussion_r931946478)

The reason raised for this idea was: A user can rebuild the collection and reupload it. After that they may be confused to find two content units with the same (namespace, name, version) but different artifacts. The whole idea of this pr is to allow having different artifacts claiming the same (namespace, name, version). I as a user would on the other hand be most suspicious if i built a collection, uploaded it downloaded it again and then got a different artifact.
I'd like to revisit that discussion, as it probably decides on whether we can ever implement on_demand sync in a sane way.

pulp_ansible/app/management/commands/datarepair-ansible-collection-sha256.py
```diff
@@ -0,0 +1,56 @@
...
+        failed_units = 0
+        repaired_units = 0
+
+        unit_qs = CollectionVersion.objects.filter(sha256__isnull=True)
+        count = unit_qs.count()
+        print(f"CollectionVersions to repair: {count}")
+        if count == 0:
+            return
+
+        for unit in unit_qs.iterator():
```
gerrod3, line 37: Should we add the`prefetch_related` here to reduce the number of queries being made?

pulp_ansible/app/tasks/collections.py
```diff
@@ -100,29 +114,30 @@ async def declarative_content_from_git_repo(remote, url, git_ref=None, metadata_
...
-        )
-        await sync_to_async(ContentArtifact.objects.get_or_create)(
-            artifact=artifact,
-            content=collection_version,
-            relative_path=collection_version.relative_path,
-        )
+        collection_version, tags = await sync_to_async(create_collection_from_importer)(metadata)
+        await _save_collection_version(collection_version, artifact, tags)
     except ValidationError as e:
         if "unique" in str(e):
```
gerrod3, line 142: You need to update the CollectionVersion.get below on line 148 to include the sha256, also can make it use aget.

pulp_ansible/tests/unit/utils.py
```diff
@@ -52,28 +52,26 @@ def build_cvs_from_specs(specs, build_artifacts=True):
...
         col, _ = Collection.objects.get_or_create(name=spec[0])
-        col.save()
-        cv = CollectionVersion(collection=col, namespace=spec[0], name=spec[1], version=spec[2])
-        cv.save()
+        cv = CollectionVersion.objects.create(
+            collection=col, sha256=artifact.sha256, namespace=spec[0], name=spec[1], version=spec[2]
+        )
         collection_versions.append(cv)
 
-        if build_artifacts:
```
gerrod3, line 61: `build_artifacts` is no longer being used in this method, can we remove it from the signature?

gerrod3 (changes requested):
Small suggestions, everything else looks good.

pulp_ansible/app/tasks/collections.py
mdellweg: Actually, I think this is the place where we still have the old behaviour, and we should not change that.
Remember this is not fixing the issue, but preparing the database to fix the issue with proper lookup later.

gerrod3 (approved):
Hurray! One step closer.

PR #1818, opened 2024-04-19
EXPERIMENT DO NOT MERGE YET!

mdellweg:
This PR contains the ZDU commits for the sha256 field on the collection version model.
It is vitally important to put every single change in the proper commit.
Also do not merge the second part until the first part is released.


mdellweg:
PR with only the first half of the story (If i remember to update properly.):
https://github.com/pulp/pulp_ansible/pull/1815

ggainey (approved):

PR #1837, opened 2024-05-07
Fix import collection marks and signatures

### LINKED ISSUE #1836: Import fails to connect collection version and metadata content
**Version**

**Describe the bug**
Import fails to associate signatures (and probably other content) to collection versions.

**To Reproduce**

**Expected behavior**

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

mdellweg:
The import code was unable to link up a collection version with its metadata because using multiple fields as a foreign key is not supported by django-import-export. The only viable fallback given existing export files is to use the upstream_id.

fixes #1836

### DIFF
--- a/CHANGES/1836.bugfix
+++ b/CHANGES/1836.bugfix
@@ -1,0 +1,1 @@
+Fixed import failing to associate signatures and marks with their collection version.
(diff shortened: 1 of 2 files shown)


mdellweg:
We should probably improve testing this.

mdellweg:
I tested that the updated test catches the error.

mdellweg:
The test has left this PR again. It would not have been backportable.
The followup PR with the test is: https://github.com/pulp/pulp_ansible/pull/1839

ggainey (approved):
lg2m!

PR #1839, opened 2024-05-08
Fix import test

mdellweg:


pulp_ansible/app/serializers.py
```diff
@@ -670,14 +684,20 @@ def is_valid(self, raise_exception=False):
         See Validating Models:
             https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/index.html
         """
-        write_fields = set(CollectionVersionUploadSerializer.Meta.fields) - {"pulp_created"}
+        write_fields = set(CollectionVersionUploadSerializer.Meta.fields) - {
+            "pulp_created",
+            "pulp_last_updated",
+        }
```
mdellweg, line 690: I have a suspicion that we are using this serializer beyond it's scope somewhere.

gerrod3 (approved):

PR #1880, opened 2024-06-13
Improve and comment on a very bad piece of code

mdellweg:
Deprecations must be handeld different in the sync code. Probably a separate stage would do the trick.

[noissue]


pulp_ansible/pytest_plugin.py
```diff
@@ -105,6 +105,8 @@ def _ansible_collection_remote_factory(include_pulp_auth=False, **kwargs):
         if include_pulp_auth:
             kwargs["username"] = bindings_cfg.username
             kwargs["password"] = bindings_cfg.password
+        # Ratelimiting, because our tests hit galaxy pretty hard.
+        kwargs["rate_limit"] = 5
```
gerrod3, line 109: Nice!

gerrod3 (approved):

PR #1922, opened 2024-07-17
Cast the content object to a collection version when rebuilding meta.

### LINKED ISSUE #1921: _rebuild_collection_version_meta is setting and saving data on the wrong model
**Version**
main branch

**Describe the bug**
The _rebuild_collection_version_meta function and parent function were refactored in https://github.com/pulp/pulp_ansible/commit/113f0c0a3689f9c76700538f219a570336aca417 but it changed the function signature of _rebuild_collection_version_meta to take in a content object instead of a collectionversion object. When the code sets the properties from the galaxy-importer results, it's setting them on the content object and saving that which doesn't actually do anything in the database.

I figured this out by turning on query logging in the database and spending a while confused as to why I never saw any UPDATE calls being made to the collectionversion table.

**To Reproduce**
Sync in a collection from any v2 api source (such as old-galaxy) and then trigger a metadata rebuild

**Expected behavior**
The contents, docs_blob and requires_ansible fields on the CollectionVersion models should get updated in the database.

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

jctanner:
The incoming object is a content object, not a collectionversion so setting the fields and saving them is a NO-OP. We have to cast() to get the right object to manipulate the relevant fields.

fixes #1921

### DIFF
--- a/pulp_ansible/tests/unit/test_tasks.py
+++ b/pulp_ansible/tests/unit/test_tasks.py
@@ -101,6 +101,25 @@ class TestCollectionReImport(TestCase):
         expected_pks = sorted([str(x.pk) for x in expected_cvs])
         assert call_pks == expected_pks
 
+    def test_reimport_repository_rebuilds_contents(self):
+        """Make sure the contents field in the CVs are rebuilt."""
+
+        cobject = self.repo.latest_version().content.first()
+        assert cobject is not None, "the repo fixture no longer has content for some unknown reason"
+
+        # set some invalid contents for each CV in the repo ...
+        cv = cobject.cast()
+        cv.contents = ["a", "b", "c"]
+        cv.save()
+
+        # call the rebuild
+        _rebuild_collection_version_meta(cobject)
+
+        # after rebuild, the contents should have been changed back
+        cv2 = cobject.cast()
+        cv2.refresh_from_db()
+        assert cv2.contents != ["a", "b", "c"]
+
 
 class TestCollectionVersionHighest(TestCase):
     """Test Collection Re-Import."""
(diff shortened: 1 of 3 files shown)


gerrod3 (dismissed):
Looks good to me!

pulp_ansible/tests/unit/test_tasks.py
```diff
@@ -101,6 +101,24 @@ def test_reimport_repository_rebuilds_name(self, mock_progress_report, mock_rebu
         expected_pks = sorted([str(x.pk) for x in expected_cvs])
         assert call_pks == expected_pks
 
+    def test_reimport_repository_rebuilds_contents(self):
+        """Make sure the contents field in the CVs are rebuilt."""
+
+        for cobject in self.repo.latest_version().content:
```
mdellweg, line 107: It should be sufficient to test this on one item, right?

mdellweg (commented):
This sounds reasonable.
But i think the test is unnecessarily looping over a lot of items. (The list could even be empty invalidating the test completely.)

mdellweg:
Looks better now. Please squash the commits.

pulp_ansible/tests/unit/test_tasks.py
```diff
@@ -101,6 +101,25 @@ def test_reimport_repository_rebuilds_name(self, mock_progress_report, mock_rebu
         expected_pks = sorted([str(x.pk) for x in expected_cvs])
         assert call_pks == expected_pks
 
+    def test_reimport_repository_rebuilds_contents(self):
+        """Make sure the contents field in the CVs are rebuilt."""
+
+        cobject = self.repo.latest_version().content.first()
+        assert cobject is not None, "the repo fixture no longer has content for some unknown reason"
```
mdellweg, line 108: :+1:

mdellweg (approved):

PR #1989, opened 2024-10-07
Fix some 500 errors in the galaxy api

gerrod3:
Moving the DRF_ACCESS_POLICY in the settings file fixes an error in the Pulp image where the setting doesn't get loaded for pulp_ansible. The second fix is for when you are browsing an empty distribution.

### DIFF
--- a/pulp_ansible/app/settings.py
+++ b/pulp_ansible/app/settings.py
@@ -20,6 +20,11 @@ LOGGING = {
     "dynaconf_merge": True,
 }
 
+DRF_ACCESS_POLICY = {
+    "dynaconf_merge_unique": True,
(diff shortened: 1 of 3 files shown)


pulp_ansible/app/settings.py
```diff
@@ -20,6 +20,11 @@
     "dynaconf_merge": True,
 }
 
+DRF_ACCESS_POLICY = {
+    "dynaconf_merge_unique": True,
+    "reusable_conditions": ["pulp_ansible.app.global_access_conditions"],
+}
+
```
mdellweg, line 27: Uhhh, I never thought the order of settings would matter.

mdellweg (approved):

PR #1995, opened 2024-10-24
Fix Any serializer type regression breaking Ruby bindings

pedro-psb:
See https://github.com/pulp/pulp_rpm/issues/3639

### DIFF
--- a/pulp_ansible/app/fields.py
+++ b/pulp_ansible/app/fields.py
@@ -1,0 +1,12 @@
+from drf_spectacular.utils import extend_schema_field
+from drf_spectacular.types import OpenApiTypes
+from rest_framework import serializers
+
+
+@extend_schema_field(OpenApiTypes.OBJECT)
+class JSONDictField(serializers.JSONField):
+    """A drf JSONField override to force openapi schema to use 'object' type.
+
+    Not strictly correct, but we relied on that for a long time.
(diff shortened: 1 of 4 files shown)


mdellweg (commented):
I'd rather call it JSONDictField.
Also I really think whenever we use this new class deliberately and not by accident, it is actually strictly correct (OK, we may be able to add more specific types in some places, but as long as the field wants a dict, this solution is not wrong about it.).

pulp_ansible/app/fields.py
```diff
@@ -0,0 +1,12 @@
...
+from drf_spectacular.types import OpenApiTypes
+from rest_framework import serializers
+
+
+@extend_schema_field(OpenApiTypes.OBJECT)
+class JSONDictField(serializers.JSONField):
+    """A drf JSONField override to force openapi schema to use 'object' type.
+
+    Not strictly correct, but we relied on that for a long time.
+    See: https://github.com/tfranzel/drf-spectacular/issues/1095
```
mdellweg, line 11: This docstring was what I was referring to that is misleading IMHO.

mdellweg (approved):
I am going to merge this now, so we can get it released.

But I hope we can resort to using the version exposed from pulpcore soon...

pulp_ansible/app/fields.py
pedro-psb: Oh, ok

PR #1997, opened 2024-10-25
Remove test dependency orion-util

mdellweg:
This package has a flaw and does not seem to be well maintained. Also we don't use much of its features anyway.


gerrod3 (approved):
Hurray!

PR #2012, opened 2024-10-30
Update Changelog

pulpbot:


CHANGES.md
```diff
@@ -58,7 +71,18 @@
...
+#### Bugfixes {: #0.21.9-bugfix }
+
+- Set `is_highest` to `False` before importing a collection version to prevent conflicts.
+  [#1986](https://github.com/pulp/pulp_ansible/issues/1986)
+- Fixed the JSONField specification so it doesn't break ruby bindings.
+  See context [here](https://github.com/pulp/pulp_rpm/issues/3639).
+
+---
+
+# ## 0.21.8 (2024-08-14) {: #0.21.8 }
```
mdellweg, line 85: @pedro-psb
Looks familiar?

The use of `##` (h2) and `####` (h4) without `###` (h3) is because we use that level for the plugin api heading, right? Do you think that may be related to what we see here?

pedro-psb: Damn.
Yeah, this header skip is to enable those sub-categories. I don't see how this could interefere (but I'm clueless about this bug, so that's not a helpful insight)

mdellweg: We probably don't strictly have the same format on all old branches and somwhere a regex for "## " matches part of "### " and boom there we are.

mdellweg (approved):

PR #2039, opened 2024-11-21
[PULP-76] Remove is highest

mdellweg:


mdellweg:
@jctanner The description about the search vector sounds scary in this context. Shouldn't search be broken, when the search vector does no longer receive updates?
Can you gauge whether this technology https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/search/ would serve our purpose too?

jctanner:
@mdellweg what "search" are you referring to? 

Galaxy's "search" function is https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/galaxy/v3/viewsets.py#L65, which uses https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/galaxy/v3/viewsets.py#L65 which has it's own flag for "is_highest" ... https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/tasks/collectionversion_index.py#L198

pulp_ansible/app/models.py
```diff
@@ -202,16 +194,9 @@ class CollectionVersion(Content):
     #   a migration file. The trigger only runs when the table is
     #   updated. CollectionVersions are INSERT'ed into the table, so
     #   the search_vector does not get populated at initial creation
-    #   time. In the import or sync workflows, is_highest gets toggled
-    #   back and forth, which causes an UPDATE operation and then the
-    #   search_vector is built.
+    #   time.
     search_vector = psql_search.SearchVectorField(default="")
```
mdellweg, line 198: I'm wondering about this search vector.

gerrod3 (approved):
I also wonder about the search vector, I don't think it is used anymore having been replaced by the index. Maybe we can also remove it in a future PR.

Issue #2081, opened 2025-01-10
[core 3.70 compat] Django storages config deprecation


pedro-psb:
Action for django's storages deprecation:

1. Pre or on core-3.70 compat declaration:
    - [x] Replace legacy with new
    - [x] Settings on signal-handler
2. On core 3.70 compat declaration: (**closes this**)
    - [ ] Use new setting in one CI scenario

See for context:
<https://discourse.pulpproject.org/t/action-required-plugins-should-use-djangos-new-storage-config-for-3-70/1866>

netsandbox:
@mdellweg just a hint: pulp core 3.70.0 was released last week and the Docker image build fails because of the ansible plugin version constraint:

https://github.com/pulp/pulp_ansible/blob/84a32718cbe18216b4067a50eea222d713c818ed/pyproject.toml#L32

mdellweg:
Yep, I released it and that's expected for a so called "breaking change release". I fear pulp_ansible is just the first in line (alphabethically) to fail the builds. We are in the process of producing compatibility releases for all the plugins.

PR #2082, opened 2025-01-10
Django storages config deprecation (1/2)

pedro-psb:
This PR:

- [x] Replaces legacy with new django storage settings
- [x] Settings on signal-handler (not required)

See #2081

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -428,19 +428,18 @@ def _rebuild_collection_version_meta(content_object):
 def _get_backend_storage_url(artifact_file):
     """Get artifact url from pulp backend storage."""
     if (
-        settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem"
+        settings.STORAGES["default"]["BACKEND"] == "pulpcore.app.models.storage.FileSystem"
         or not settings.REDIRECT_TO_OBJECT_STORAGE
     ):
         url = None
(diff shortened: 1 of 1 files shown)


mdellweg:
Again I'm confused how this setting is upgraded. I had expected that this would fail before pulpcore 3.70. But It still seems to do...

pulp_ansible/app/tasks/collections.py
```diff
@@ -428,19 +428,18 @@ def _rebuild_collection_version_meta(content_object):
...
         parameters = {"ResponseContentDisposition": "attachment%3Bfilename=archive.tar.gz"}
         url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
-    elif settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage":
+    elif settings.STORAGES["default"]["BACKEND"] == "storages.backends.azure_storage.AzureStorage":
         url = artifact_file.storage.url(artifact_file.name)
     else:
         raise NotImplementedError(
-            f"The value settings.DEFAULT_FILE_STORAGE={settings.DEFAULT_FILE_STORAGE} "
-            "was not expected"
+            f'The value {settings.STORAGES["default"]["BACKEND"]=} ' "was not expected"
```
mdellweg, line 442:
```suggestion
            f'The value {settings.STORAGES["default"]["BACKEND"]=} was not expected'
```

pedro-psb:
This is a bit confusing. What happens is that django do translate the old name internally, we could use the new name in the codebase all this time. The problem we needed to solve was that the users couldnt (in their settings file).

mdellweg (approved):

PR #2084, opened 2025-01-15
Add stricter upperbound for galaxy-importer to account for dropping of py3.8

gerrod3:
Our older images <core3.49 use python3.8. A recent release of galaxy-importer now uses code not present in py3.8, and it seems they haven't been testing 3.8 for a while. So restrict to last known supported version: https://github.com/ansible/galaxy-importer/commit/109c1ee9744be6c3c337a9efa065482dbd032172


mdellweg:
https://github.com/pulp/pulp_ansible/pull/2085

Issue #2131, opened 2025-03-04
Deadlock when performing multiple syncs with similar content


gerrod3:
https://issues.redhat.com/browse/SAT-27939

PR #2132, opened 2025-03-04
Fix potential deadlock for concurrent syncs

gerrod3:
Set all the extra data the collection version needs in the pre save step, this prevents us from editing an already saved collection (probable cause of the deadlock). If the collection version has bad metadata then it should be fixed with the repair endpoint, not in a sync. 

fixes: #2131
https://issues.redhat.com/browse/SAT-27939

### DIFF
--- a/CHANGES/2131.bugfix
+++ b/CHANGES/2131.bugfix
@@ -1,0 +1,1 @@
+Fixed deadlock when performing multiple syncs with similar collections.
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

dralley (approved):
Vladimir hasn't been able to reproduce a deadlock on the most recent version of this patch, but he did hit an OOM before being able to complete the test.

I don't actually think the OOM is related - at the very least I don't see anything in here that looks like a meaningful change in that respect.  The "tags" (names) are being held a bit longer but that's about it and I doubt it's relevant.

In terms of deadlock resolution this patch looks fine to me, so unless futher evidence points to there being a clear problem with memory somehow I would say we can probably move forwards with this.

PR #2168, opened 2025-04-08
Added namespace progress report

fao89:


rochacbruno (dismissed):

gerrod3 (changes requested):
Have you seen https://github.com/pulp/pulp_ansible/pull/2139/? I'm not sure this takes into account this fix.

pulp_ansible/app/tasks/collections.py
```diff
@@ -1052,6 +1050,15 @@ async def run(self):
             # Ensure PR 'total' is correct before stage finishes
             pb.total = pb.done
 
+        tasks = []
+        msg = _("Parsing Namespace Metadata")
+        async with ProgressReport(message=msg, code="sync.parsing.namespace", total=0) as pr:
```
gerrod3, line 1055:
```suggestion
        async with ProgressReport(message=msg, code="sync.parsing.namespace", total=len(self.namespace_shas)) as pr:
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -1052,6 +1050,15 @@ async def run(self):
...
+            self.parsing_namespace_progress_bar = pr
+            for namespace, namespace_sha in self.namespace_shas.items():
+                tasks.append(loop.create_task(self._add_namespace(namespace, namespace_sha)))
+            await asyncio.gather(*tasks)
+            pr.total = pr.done
```
gerrod3, line 1060:
```suggestion
```
Since we know the total before hand we don't have to set it to done. In fact the amount we are going to add vs the amount that did get added can differ, so maybe let it be reflected in the `done` vs `total` amounts.

pulp_ansible/app/tasks/collections.py
```diff
@@ -720,6 +717,7 @@ async def _add_namespace(self, name, namespace_sha):
             namespace = AnsibleNamespaceMetadata(**namespace)
             dc = DeclarativeContent(namespace, d_artifacts=da)
             await self.put(dc)
+            await self.parsing_namespace_progress_bar.aincrement()
             return True
 
         return False
```
gerrod3, line 723: We don't need the 
```suggestion
            await self.parsing_namespace_progress_bar.aincrement()

        log.info("Failed to find namespace...")
```
The `return True/False` is no longer needed, so maybe we add some log statements for when we fail to add the namespace metadata.

pulp_ansible/tests/functional/api/collection/test_sync.py
```diff
@@ -374,3 +375,5 @@ def test_sync_progress_report(
     for pr in progress_reports:
         if pr.message == "Parsing CollectionVersion Metadata":
             assert pr.total == pr.done
+        if pr.message == "Parsing Namespace Metadata":
+            assert pr.total == pr.done
```
gerrod3, line 379:
```suggestion
            assert pr.total == pr.done == 1
```
I feel that we should know the total amount for this test sync since it is only one collection.

gerrod3 (changes requested):
Re: my earlier review. I do think this takes care of namespace optimization fix that I made in #2139, so I am willing to close that in favor of this. Can you add a bugfix changelog to this PR as well?

pulp_ansible/app/tasks/collections.py
fao89: I kept this because I only increment when the download succeeds,
so this would just the total of downloads

gerrod3 (approved):

PR #2189, opened 2025-04-30
Fix broken collection import logs

bmclaughlin:
Connected pulp_ansible.app.tasks.upload.process_collection_artifact logger to collection_import handler.

fixes: PULP-525

### DIFF
--- a/CHANGES/+import-logs.bugfix
+++ b/CHANGES/+import-logs.bugfix
@@ -1,0 +1,1 @@
+Fixed broken collection import logs.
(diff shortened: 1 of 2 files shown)


CHANGES/525.bugfix
gerrod3, line 1: When we don't file a github issue we will typically name our changelogs with words like so:
`+import-logs.bugfix` or `+artifact-logger.bugfix`. This way towncrier won't get confused and try to link an old issue with this one.

PR #2208, opened 2025-05-27
Remove tags table

mdellweg:
Tags of collection versions are now only handled by an array field. The migration in this commit is save to run as long as every component of pulp is running on pulp_ansible >= 0.26.


pulp_ansible/app/viewsets.py
```diff
@@ -1189,27 +1187,20 @@ class AnsibleDistributionViewSet(DistributionViewSet, RolesMixin):
...
     ViewSet for Tag models.
     """
 
     endpoint_name = "pulp_ansible/tags"
-    queryset = Tag.objects.all()
+    queryset = (
+        CollectionVersion.objects.all()
+        .annotate(tag=Func(F("tags"), function="unnest"))
+        .values("tag")
+        .values(count=Count("pk"), name=F("tag"))
```
gerrod3, line 1200: I know I already approved this in the last PR, but sometimes SQL is straight magic...

pulp_ansible/app/viewsets.py
```diff
@@ -1189,27 +1187,20 @@ class AnsibleDistributionViewSet(DistributionViewSet, RolesMixin):
     }
 
 
-class TagViewSet(NamedModelViewSet, mixins.ListModelMixin):
+class TagViewSet(viewsets.GenericViewSet, mixins.ListModelMixin):
```
gerrod3, line 1190: Any particular reason for changing this from a NamedModelViewSet?

pulp_ansible/app/urls.py
```diff
@@ -249,5 +249,12 @@
     path(GALAXY_API_ROOT + "v1/", include(v1_urls)),
     path(GALAXY_API_ROOT + "v3/", include(v3_urls)),
     path(GALAXY_API_ROOT, GalaxyVersionView.as_view()),
-    re_path(r"^pulp/api/v3/ansible/copy/$", CopyViewSet.as_view({"post": "create"})),
+    path(
+        settings.V3_API_ROOT_NO_FRONT_SLASH + "pulp_ansible/tags/",
+        TagViewSet.as_view({"get": "list"}),
+    ),
+    path(
+        settings.V3_API_ROOT_NO_FRONT_SLASH + "ansible/copy/",
```
gerrod3, line 257: Good change on the hardcoded API_ROOT! Shame we used 'ansible/' for copy and 'pulp_ansible/' for tags.

pulp_ansible/app/tasks/upload.py
```diff
@@ -75,12 +75,8 @@ def process_collection_artifact(artifact, namespace, name, version):
...
 
 
-def finish_collection_upload(collection_version, tags, origin_repository):
-    """After CollectionVersion has been created update its tags and latest_version."""
-    for name in tags:
-        tag, created = Tag.objects.get_or_create(name=name)
-        collection_version.tags.add(tag)
-
+def finish_collection_upload(collection_version, origin_repository):
+    """After CollectionVersion has been created update its repository."""
```
gerrod3, line 79: Think of any ideas for removing this workaround now that tags are solved?

pulp_ansible/app/migrations/0062_pivot_collectionversion_tags.py
```diff
@@ -0,0 +1,140 @@
...
+    || (
+      SELECT
+        setweight(to_tsvector(
+          coalesce(string_agg(cvc ->> 'name', ' '), '')
+        ), 'C')
+      FROM jsonb_array_elements(cv.contents) AS cvc
+    )
+    || setweight(to_tsvector(coalesce(description, '')), 'D')
+    FROM ansible_collectionversion cv
+    WHERE cv.content_ptr_id = NEW.content_ptr_id
```
gerrod3, line 76: If this is from the original TS_VECTOR_SELECT should it also be present in the new one?

pulp_ansible/app/migrations/0062_pivot_collectionversion_tags.py
```diff
@@ -0,0 +1,140 @@
...
+"""
+
+# Creates a database function and a trigger to update collection search
+# vector field.
+# In the new version, tags are availabel at create and content should never be updated,
+# so having both insert and update in the trigger from now on is probably fast _and_ safe.
```
gerrod3, line 82: Does the trigger only activate when the fields used in the search vector are updated or anytime a collection-version is updated? I'm thinking about how on uploads we potentially do two saves, 1 insert and 1 update and the second update is for only for the CV's repository field which isn't used in the search vector. Not sure if this is a big deal or not.

gerrod3 (changes requested):
Mostly questions, with the big ones around the gnarly migration.

pulp_ansible/app/viewsets.py
mdellweg: Yes, there's no model for it anymore.
And Pulp's viewset chasing magic goes boom if it finds two viesets for `CollectionVersion`.

pulp_ansible/app/tasks/upload.py
mdellweg: looking...

pulp_ansible/app/migrations/0062_pivot_collectionversion_tags.py
mdellweg: So we _are_ violating the immutable state of the content...

I had possible data repair commands in mind that may save to content for whatever reason. And they really shouldn't be surprised. Again, this new query acts only on the inserted / updated row, like a generated field. I believe it to still be faster.

pulp_ansible/app/viewsets.py
mdellweg: Well this is sql through the lenses of Django ORM.
The one thing that makes it really weird is the way "unnest" works.
It explodes the arrayfield into individual rows, a bit like left join.

pulp_ansible/app/tasks/upload.py
mdellweg: I think the name clash is still not resolved and we need the workaround.
But some comments around it are more than outdated, and it is only that one thing left. I will fold it into the serializer create call. That should surround it with the better context to understand it.

PR #2251, opened 2025-07-11
Make undeprecating collections play by the rules

mdellweg:
Modifying the RepositoryContent table was never a good idea, and with the new plans about changing the datamodel around repository versions, it isn't a viable option anymore.


gerrod3 (approved):
Nice!

PR #2309, opened 2025-09-17
Add results to syncronize tasks

mdellweg:
If a new repository version is created, it is returned as the task result now.

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -77,6 +77,13 @@ from pulp_ansible.app.tasks.utils import (
     parse_metadata,
 )
 
+
+try:
+    from pulpcore.plugin.serializers import RepositoryVersionSerializer
+except ImportError:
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -213,26 +220,25 @@ def sync(remote_pk, repository_pk, mirror, optimize):
...
     first_stage = CollectionSyncFirstStage(remote, repository, is_repo_remote, optimize)
-    if first_stage.should_sync is False:
+    if first_stage.should_sync:
+        d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
+        repository_version = d_version.create()
+
+        if repository_version is not None:
+            repository.last_synced_metadata_time = first_stage.last_synced_metadata_time
+            repository.save(update_fields=["last_synced_metadata_time"])
+            if RepositoryVersionSerializer is not None:
```
ggainey, line 236: Nice, I like this pattern.

PR #2319, opened 2025-09-23
Enable import/export tests when domains are enabled.

ggainey:


ggainey:
> How come these tests do not fail on lowerbounds where pulpcore 3.63 is installed? Is that expected?
> 
> https://github.com/pulp/pulp_ansible/actions/runs/17959822120/job/51080436446?pr=2319#step:14:25134
> 
> ```
> usr/local/lib/python3.11/site-packages/pulp_ansible/tests/functional/api/test_export_import.py::test_export_then_import[True] PASSED [ 85%]
> usr/local/lib/python3.11/site-packages/pulp_ansible/tests/functional/api/test_export_import.py::test_export_then_import[False] PASSED [ 86%]
> ```

That is a Really Good Question, given that 3.63 does have the "not if DOMAIN_ENABLED" guard : https://github.com/pulp/pulpcore/blob/3.63/pulpcore/app/viewsets/exporter.py#L124

I have no idea.

ggainey:
> I have no idea.

I lie - the lowerbounds-test runs with domains disabled. These tests have always worked there - all this change does, it make them run when domains *are* enabled, in the azure/s3 envs.

mdellweg (approved):

Issue #2350, opened 2025-12-16
Fix cross-repo search tag filter for ArrayField


bmclaughlin:
**Version**
Please provide the versions of the pulpcore and pulp_ansible packages in use, and how they are installed.  If you are using Pulp via Katello, please provide the Katello version.
Working on upgrading galaxy_ng to pulpcore 3.73 and ran into this bug.

**Describe the bug**
In `/pulp_ansible/pulp_ansible/app/galaxy/v3/filters.py:190`
`qs = qs.filter(collection_version__tags__name=tag)`

However, Tags is no longer a Model, it's an ArrayField on CollectionVersion.  So that line needs to change to:
`qs = qs.filter(collection_version__tags__contains=[tag])`

**To Reproduce**
It was picked up in galaxy_ng's `integration (standalone, x_repo_search)` CI run:
```
Exception Type: FieldError at /api/galaxy/v3/plugin/ansible/search/collection-versions/
Exception Value: Unsupported lookup 'name' for ArrayField or join on the field not permitted, perhaps you meant ne?
Raised during: pulp_ansible.app.galaxy.v3.viewsets.CollectionVersionSearchViewSet
```

**Expected behavior**
The line of code to return the correct queryset. 

**Additional context**
I'm about to open a PR, I tested locally to ensure it fixed the error.

PR #2351, opened 2025-12-16
Fix cross-repo search tag filter for ArrayField

bmclaughlin:
The tags field on CollectionVersion is now an ArrayField instead of a ForeignKey to Tag model. Updated filter_by_tags to use the contains lookup which is appropriate for ArrayFields.

Issue: AAP-53522

Fixes: #2350

### DIFF
--- a/pulp_ansible/app/galaxy/v3/filters.py
+++ b/pulp_ansible/app/galaxy/v3/filters.py
@@ -187,6 +187,4 @@ class CollectionVersionSearchFilter(FilterSet):
         Returns:
             Queryset of CollectionVersion that matches all tags
         """
-        for tag in value.split(","):
-            qs = qs.filter(collection_version__tags__name=tag)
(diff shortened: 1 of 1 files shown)


pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -188,5 +188,6 @@ def filter_by_tags(self, qs, name, value):
             Queryset of CollectionVersion that matches all tags
         """
         for tag in value.split(","):
-            qs = qs.filter(collection_version__tags__name=tag)
+            # tags is an ArrayField, use contains lookup
+            qs = qs.filter(collection_version__tags__contains=[tag])
```
gerrod3: How about:
`return qs.filter(collection_version__tags__contains=value.split(","))`

gerrod3 (approved):

PR #2454, opened 2026-03-06
[PULP-402] Greatly improve memory usage during collection syncs

gerrod3:
The memory reduction comes in two parts:

1. Refactor first stage to limit number of coroutines running at once to 100
2. Avoid when possible json serialization of docs_blob, files, manifest & contents fields

The memory growth should remain flat (or extremely minimal) after first batch (500) processing of collection versions. Most syncs should now complete with less than 1 GB of memory usage.

https://issues.redhat.com/browse/PULP-402
Assssisted by: clause-opus-4.6

### Results
Perform a sync on a clean system with this requirements file:
```json
{
'collections': 
  [
    {'name': 'amazon.aws'},
    {'name': 'community.general'}, 
    {'name': 'ansible.netcommon'}, 
    {'name': 'ansible.utils'}, 
    {'name': 'ansible.posix'}, 
    {'name': 'community.crypto'}
  ]
}
```
| Metric | Original | No Docs/Files JSON serialize | All Optimizations |
|---|---|----|----|
|  Peak RSS  | ~4160 MB | 897 MB | 598 MB |
| Duration  | ~10 min  | ~8 min | ~6.5 min |
| Reduction | ~~ | 78% | 86%|

### Explanation

The majority of the savings comes from not loading the JSON fields into memory as dicts. The fields can be relatively large dicts, with docs_blob sometimes being multiple MB. Python's pymalloc allocator fragments memory for the many small nested objects within the field's dict and never returns it to the OS, causing RSS to grow monotonically. We bypass the object allocation with three changes:
1. Only read the field as a json string into memory before the insertion into the database
2. Use raw SQL to directly insert the json strings into PG to avoid the ORM serialization
3. Use a custom manager that defers these memory hog fields during sync

The second part of the savings comes from refactoring the first stage to limit the number of concurrently running coroutines. We spawn 1 coroutine per collection version to sync and each one does a fetch to grab its metadata. This metadata is again JSON that gets trapped in memory until the coroutine finishes. Because we ran them all at once this would cause a rapid increase of memory per batch that wouldn't decrease till the first stage finished. We fix this with these three changes:
 1. All the methods return lists of coroutines instead of creating tasks that auto-start
 2. The `run` method uses a single `gather` with a batched (100) list of the coroutines.
 3. Move around the metadata processing inside `add_collection_version` to be able to free it as soon as possible

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -1,4 +1,5 @@
 import asyncio
+from collections.abc import Coroutine
 import functools
 import hashlib
 import json
@@ -7,6 +8,7 @@ import tarfile
 from collections import defaultdict
 from gettext import gettext as _
 from operator import attrgetter
+from pathlib import Path
 from urllib.parse import urljoin
 from uuid import uuid4
 import tempfile
@@ -14,7 +16,7 @@ import tempfile
 import yaml
 from aiohttp.client_exceptions import ClientError, ClientResponseError
 from asgiref.sync import sync_to_async
-from django.db import transaction
+from django.db import connection, transaction
 from django.db.models import Q
 from django.db.utils import IntegrityError
 from django.utils.dateparse import parse_datetime
@@ -77,6 +79,7 @@ from pulp_ansible.app.tasks.utils import (
     parse_collections_requirements_file,
     parse_metadata,
 )
+from pulp_ansible.app.utils import set_collection_deferred_fields
 
 
 try:
@@ -230,6 +233,7 @@ def sync(remote_pk, repository_pk, mirror, optimize):
 
     first_stage = CollectionSyncFirstStage(remote, repository, is_repo_remote, optimize)
     if first_stage.should_sync:
+        set_collection_deferred_fields(["docs_blob", "manifest", "files", "contents"])
         d_version = AnsibleDeclarativeVersion(first_stage, repository, mirror=mirror)
         repository_version = d_version.create()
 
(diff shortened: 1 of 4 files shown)


pulp_ansible/app/tasks/collections.py
```diff
@@ -1211,6 +1217,51 @@ def _pre_save(self, batch):
...
+                    id_match = pre_id == collection_version.pulp_id
+                    if not (pre_adding and id_match):
+                        skip_count += 1
+                        continue
+                    files = d_content.extra_data.pop("files_raw")
+                    if docs_blob_path := d_content.extra_data.get("docs_blob_path"):
+                        with open(docs_blob_path, "r") as docs_blob_file:
+                            blob = docs_blob_file.read()
+                        sql = (
+                            "UPDATE ansible_collectionversion"
```
mdellweg: This sounds particularly troubling. Content is supposed to be immutable, because it is shared.
Update violates it. That means there is a race condition here.

mdellweg (commented):
This is a rather huge change. But there _is_ a recurring theme: Too much unbounded data fields (do we really need them) in the database put a lot of pressure on the memory. Can we gain something by moving them into a separate one-to-one table?

mdellweg (approved):

dralley:
@gerrod3 Is the `json-stream` package, which we already depend on for import/export, helpful at all?

https://github.com/daggaz/json-stream?tab=readme-ov-file#-what-are-the-problems-with-the-standard-json-package

Issue #712, opened 2021-11-17
As an user I want to sync collections on demand


pulpbot:
Author: @fao89 (fao89)

Redmine Issue: 7470, https://pulp.plan.io/issues/7470

---

Enable on demand sync for collections

pulpbot:
From: @fao89 (fao89)
Date: 2020-09-08T14:05:26Z

---

our _post_save method is heavily dependent on immediate policy:
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/tasks/collections.py#L432-L456

pulpbot:
From: @fao89 (fao89)
Date: 2020-09-24T13:06:27Z

---

Do we have an endpoint that brings the MANIFEST.json data?
~~~ json
{
 "collection_info": {
  "namespace": "pulp",
  "name": "pulp_installer",
  "version": "3.6.3-1",
  "authors": [
   "Pulp Project Developers <pulp-dev@redhat.com>"
  ],
  "readme": "README.md",
  "tags": [
   "pulp",
   "pulpcore",
   "content",
   "software_repositories",
   "lifecycle_management",
   "content_delivery"
  ],
  "description": "A collection of Ansible roles to install & configure Pulp 3 from PyPI",
  "license": [
   "GPL-2.0-or-later"
  ],
  "license_file": null,
  "dependencies": {},
  "repository": "https://github.com/pulp/pulp_installer",
  "documentation": "https://pulp-installer.readthedocs.io/",
  "homepage": "https://pulpproject.org/",
  "issues": "https://pulp.plan.io/projects/pulp/issues?query_id=156"
 },
 "file_manifest_file": {
  "name": "FILES.json",
  "ftype": "file",
  "chksum_type": "sha256",
  "chksum_sha256": "54757b4a488e5be8666ded617aa213740e03e3387dd30a7a9484611fd5e2fac4",
  "format": 1
 },
 "format": 1
}
~~~

pulpbot:
From: @fao89 (fao89)
Date: 2020-09-24T13:14:48Z

---

Apparently, the version endpoint brings all information we need: 

URL: https://cloud.redhat.com/api/automation-hub/v3/collections/cyberark/conjur_collection/versions/1.0.3/
~~~ json
{
    "version": "1.0.3",
    "certification": "certified",
    "href": "/api/automation-hub/v3/collections/cyberark/conjur_collection/versions/1.0.3/",
[...]

PR #2471, opened 2026-03-15
Sync only latest version of collections by default

dhageman:
The behavior of the current collection sync without a requirements file is to pull all collections and every version available of each collection.  This can lead to failures in the collection sync process due to memory inefficiencies in how collection syncs are performed. 

This PR switches the default behavior to only sync the latest version of each collection that is available.  
 
The current behavior can be achieved by providing a requirements file with a version set to a wildcard or version range.

Assisted By: Claude Code

### 📜 Checklist

- [x] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [x] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [x] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [ ] (For new features) - User documentation and test coverage has been added

### DIFF
--- a/CHANGES/+collection_sync_latest.removal
+++ b/CHANGES/+collection_sync_latest.removal
@@ -1,0 +1,1 @@
+Optimize collection sync to only fetch the latest version when no requirements filter is provided.
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -992,7 +992,23 @@ class CollectionSyncFirstStage(Stage):
 
     async def _find_all_collections(self):
         if self._unpaginated_collection_version_metadata:
-            await self._find_all_collections_from_unpaginated_data()
+            loop = asyncio.get_event_loop()
+            tasks = []
+            for namespace, collections in self._unpaginated_collection_metadata.items():
+                for name, collection in collections.items():
+                    highest = (collection.get("highest_version") or {}).get("version")
+                    version_spec = f"=={highest}" if highest else "*"
+                    entry = RequirementsFileEntry(
+                        name=f"{namespace}.{name}",
+                        version=version_spec,
+                        source=None,
+                    )
+                    tasks.append(
+                        loop.create_task(self._fetch_collection_metadata(entry))
+                    )
+            self.parsing_metadata_progress_bar.total = len(tasks)
+            await self.parsing_metadata_progress_bar.asave(update_fields=["total"])
+            await asyncio.gather(*tasks)
             return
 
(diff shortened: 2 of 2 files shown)


mdellweg:
I am uncomfortable with changing default behavior.
This will almost certainly break long standing existing workflows.

gerrod3:
@dhageman We recently just merged #2454 to reduce the memory of collection syncs. I would wait till we release it and try it out before many further changes to the sync behavior. It was also backported to 0.24, 0.25, 0.28, and 0.29

dhageman:
@mdellweg, @gerrod3 - I appreciate the feed back!

I have confidence that memory issues will be better with the recent patches.  Those are welcome changes!

I was hesitant to submit this PR because memory consumption was the most pressing issue, but I changed my mind after discussing it with some colleagues.  I think it is a good way to start a conversation if the current behavior is the best default behavior.  

I don't believe old collections are reaped from the current repositories for collections.  This means the amount of data that is synchronized continues to grow.  New organizations starting fresh syncs will discover ever-growing synchronization times paired with increased storage requirements.  

Is it reasonable to assume that an organization will want a local copy of every version of every collection available since the beginning of time?

Is it more reasonable to sync all the collections but only the latest version?  The additive nature of syncs will continue to pick up the latest version.  

Remember, the original behavior can be achieved with an explicit requirements file.  

I have reached out to a few additional people to get their feedback on whether a change of this nature is worth the effort.  If not - all good - at least we had the conversation!

gerrod3:
We probably won't accept changing the default behavior, but I think we would accept a new field on the remote `latest_version` or `sync_latest_versions=#` to allow modifying the behavior of the sync. This is what other plugins do.

As for reducing disk space usage on initial sync, this is what on-demand syncs are for. I think we are pretty close to being able to solve this in pulp-ansible now #712.

MasonMines2006:
Hello @dhageman! Team member from the Ansible Hub team here. This ticket would solve a similar issue for us, as we are trying to add a "latest" only sync for us as well, which directly depends on this.

Would it be possible to scope this to remove the breaking change to the default behavior, and instead just add the ability to sync the latest version of a collection?

MasonMines2006:
Hey @dhageman ! Drafted a PR with this approach: #2563. It adds version: "latest" as a specifier instead of changing the default behavior. Would love your feedback!

PR #2492, opened 2026-04-02
Reduce memory usage during collection deletion

Funi1234:
### Problem
Deleting collections with many versions causes worker processes to consume excessive memory and get killed with SIGKILL. This occurs because Django loads all fields (including large JSON fields) for each CollectionVersion into memory. When multiplied across many versions, this causes the worker process to be terminated.

### Solution
This PR applies targeted QuerySet field selection using `.only()` to load only the fields needed for deletion operations, avoiding the large JSON fields that aren't required.

**Changes:**
- `pulp_ansible/app/galaxy/v3/views.py`: 
  - Use `.only("pk")` when iterating collection versions and their repositories in `CollectionViewSet.destroy()`
  - Batch `AnsibleRepository` lookup with `filter(pk__in=...)` to prevent N+1 queries
  - Use `.only("namespace", "name", "version")` when loading collection dependents
- `pulp_ansible/app/tasks/deletion.py`: 
  - Use `.only("pk")` when loading collection versions and iterating their repositories in `delete_collection()` task

By loading only the required fields, we avoid pulling large JSON blobs (`docs_blob`, `metadata`, `dependencies`, etc.) into memory when they're not needed for the deletion operation.

### Related Work
- #2454 - Applied similar memory optimizations to collection sync operations

### DIFF
--- a/pulp_ansible/app/galaxy/v3/views.py
+++ b/pulp_ansible/app/galaxy/v3/views.py
@@ -80,7 +80,7 @@ from pulp_ansible.app.viewsets import (
 
 from pulp_ansible.app.tasks.deletion import delete_collection_version, delete_collection
 
-from pulp_ansible.app.utils import filter_content_for_repo_version, set_collection_deferred_fields
+from pulp_ansible.app.utils import filter_content_for_repo_version
 
 DOMAIN_ENABLED = settings.DOMAIN_ENABLED
 
@@ -446,15 +446,9 @@ class CollectionViewSet(
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
-        # Defer loading large JSON fields to prevent memory exhaustion
-        set_collection_deferred_fields(
-            ["contents", "docs_blob", "manifest", "files", "dependencies"]
-        )
-
         repositories = set()
-        for version in collection.versions.all():
-            for repo in version.repositories.all():
-                repositories.add(repo)
+        for version in collection.versions.only("pk"):
+            repositories.update(version.repositories.only("pk"))
 
         async_result = dispatch(
             delete_collection,
@@ -469,9 +463,9 @@ def get_collection_dependents(parent):
     """Given a parent collection, return a list of collection versions that depend on it."""
     key = f"{parent.namespace}.{parent.name}"
     return list(
-        CollectionVersion.objects.exclude(collection=parent).filter(
-            dependencies__has_key=key, pulp_domain=get_domain()
-        )
+        CollectionVersion.objects.exclude(collection=parent)
+        .filter(dependencies__has_key=key, pulp_domain=get_domain())
+        .only("namespace", "name", "version")
     )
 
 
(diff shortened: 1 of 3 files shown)


Funi1234:
This will need to be backported for 0.25 (AAP 2.4/2.5/2.6) and 0.28 (AAP upstream/2.7) please.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -446,6 +446,11 @@ def destroy(self, request: Request, *args, **kwargs) -> Response:
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
+        # Defer loading large JSON fields to prevent memory exhaustion
+        set_collection_deferred_fields(
+            ["contents", "docs_blob", "manifest", "files", "dependencies"]
+        )
+
```
gerrod3: Looking at the code for this call we don't need to use this utility method since the only fields we need are the `pk`. How about this instead:

```python
repositories = set()
for version in collection.versions.only("pk"):
    repositories.update(version.repositories.values_list("pk", flat=True))
```
And then change `get_collection_dependents` to also only retrieve the needed fields:
```python
key = f"{parent.namespace}.{parent.name}"
return list(
    CollectionVersion.objects.exclude(collection=parent).filter(
        dependencies__has_key=key, pulp_domain=get_domain()
    ).only("namespace", "name", "version")
)
```

pulp_ansible/app/tasks/deletion.py
```diff
@@ -71,6 +72,9 @@ def delete_collection(collection_pk):
     2. Run orphan_cleanup to delete the CollectionVersions
     3. Delete Collection
     """
+    # Defer loading large JSON fields to prevent memory exhaustion
+    set_collection_deferred_fields(["contents", "docs_blob", "manifest", "files", "dependencies"])
+
```
gerrod3: Let's do the same here as well, use `only("pk")` on the `collection.versions.all()` query to only retrieve the needed pk.

gerrod3 (changes requested):
`set_collection_deferred_fields` is a blunt instrument used to work around having to change multiple spots in our collection sync code (some of them being in pulpcore!) and as such we shouldn't overuse when trying to optimize pulp-ansible. For this case we can easily fix the issues by being smarter with the querysets we write.

pulp_ansible/app/galaxy/v3/views.py
```diff
@@ -447,9 +447,8 @@ def destroy(self, request: Request, *args, **kwargs) -> Response:
             )
 
         repositories = set()
-        for version in collection.versions.all():
-            for repo in version.repositories.all():
-                repositories.add(repo)
+        for version in collection.versions.only("pk"):
+            repositories.update(version.repositories.only("pk"))
```
gerrod3, line 452: One final change for maximum performance: 
```suggestion
        for version in collection.versions.only("pk"):
            repositories.update(version.repositories.only("pk"))
        repositories = AnsibleRepository.objects.filter(pk__in=repositories)
```
I took another look at our pulpcore code and basically when you pass a list of exclusive/shared resources to `dispatch` it is going to perform a query on each resource if they are not the detail/child model. `version.repositories` returns `core.Repository` objects (the parent model), but we want `AnsibleRepository` objects (the child model), so doing 1 extra query here will save us N queries down the road.

gerrod3 (changes requested):

gerrod3 (approved):
Thanks for the fixes!

PR #2555, opened 2026-05-20
Append `pk` to CV ordering filters for deterministic ordering

jerabekjiri:
Issue: https://redhat.atlassian.net/browse/AAP-75413

### Description
When using `order_by` with non-unique fields (e.g., name, namespace), the cross-repo search API can return results in inconsistent order across paginated requests. This causes items to change order on a same result.

This is reproducible in `https://console.redhat.com/api/automation-hub/v3/plugin/ansible/search/collection-versions/?repository_name=staging&order_by=namespace` (only when adding `order_by=namespace`). I suspect this is because many collection versions that share same namespace. Because I'm not able to reproduce locally with small dataset

### How it works
The fix appends pk as a tiebreaker to every explicit ORDER BY. For example, ?order_by=name previously generated:
`ORDER BY collection_version.name`
Now generates:
`ORDER BY collection_version.name, pk`

 This ensures that rows sharing the same name are always returned in a consistent, deterministic order based on their primary key.

### Changes
  - Add pk to `SemanticVersionOrderingFilter` for all ordering requests (name, namespace, pulp_created, version)
  - pk is only added when an explicit ordering is requested — unordered queries are not affected  
  - Also claude suggested to add pk after `-rank` in `filter_by_q` for deterministic full-text search results when multiple documents share the same relevance score
  - Add unit tests for `CollectionVersionSearchFilter` ordering behavior

 ### Workaround (prior)
 One of the workaround is to add &order_by=pulp_created as a secondary sort field. This workaround remains compatible with the fix.

### DIFF
--- a/pulp_ansible/app/galaxy/v3/filters.py
+++ b/pulp_ansible/app/galaxy/v3/filters.py
@@ -10,13 +10,14 @@ from django_filters import (
 )
 from rest_framework.exceptions import ValidationError
 
+from pulpcore.filters import StableOrderingFilter
 from pulpcore.plugin.models import RepositoryVersion
 from pulpcore.plugin.viewsets import LabelFilter
 
 from pulp_ansible.app import models
 
 
-class SemanticVersionOrderingFilter(filters.OrderingFilter):
+class SemanticVersionOrderingFilter(StableOrderingFilter):
     def filter(self, qs, value):
         if value is not None and any(v in ["version", "-version"] for v in value):
             order = "-" if "-version" in value else ""
--- a/pulp_ansible/app/galaxy/v3/filters.py
+++ b/pulp_ansible/app/galaxy/v3/filters.py
@@ -10,9 +10,8 @@ from django_filters import (
 )
 from rest_framework.exceptions import ValidationError
 
-from pulpcore.filters import StableOrderingFilter
 from pulpcore.plugin.models import RepositoryVersion
-from pulpcore.plugin.viewsets import LabelFilter
+from pulpcore.plugin.viewsets import LabelFilter, StableOrderingFilter
 
 from pulp_ansible.app import models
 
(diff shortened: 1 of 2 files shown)


pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -31,9 +31,13 @@ def filter(self, qs, value):
                 f"{order}collection_version__version_minor",
                 f"{order}collection_version__version_patch",
                 f"{order}prerelease",
+                "pk",
```
mdellweg: First, this should be solved by using the pulpcore.filters.StableOrderingFilter.

And second, i think we can just order the version field itself.

mdellweg (changes requested):

pulp_ansible/app/galaxy/v3/filters.py
jerabekjiri: ah nice, so I can just use that instead of my solution

jerabekjiri:
```
ERROR: Detected bad imports from pulpcore:
pulp_ansible/app/galaxy/v3/filters.py:13:from pulpcore.filters import StableOrderingFilter

Plugins should import from pulpcore.plugin.
```
I don't see it in the pulpcore.plugin, it needs to be exposed first

jerabekjiri: added it in https://github.com/pulp/pulpcore/pull/7737

pulp_ansible/app/galaxy/v3/filters.py
```diff
@@ -176,7 +158,7 @@ def filter_by_q(self, queryset, name, value):
             function="ts_rank",
             output_field=db_fields.FloatField(),
         )
-        return qs.annotate(rank=ts_rank_fn).order_by("-rank")
+        return qs.annotate(rank=ts_rank_fn).order_by("-rank", "pk")
```
mdellweg: Is this still needed then?

jerabekjiri: removed

mdellweg (approved):

PR #2563, opened 2026-06-24
Add sync_latest_versions field to CollectionRemote

MasonMines2006:
##### SUMMARY

Adds a `sync_latest_versions` integer field to `CollectionRemote` that limits how many of the highest versions to sync per collection. `0` (default) syncs all versions, preserving existing behavior. When set to a positive number, only the N highest non-pre-release versions matching the requirement spec are synced.

This follows the `retain_package_versions` pattern from pulp_rpm but filters pre-download instead of post-download, saving both sync time and bandwidth. The requirements.yml format is unchanged.

##### EXAMPLE

```python
# Sync only the latest version of each collection
remote.sync_latest_versions = 1

# Sync the 3 newest versions of each collection
remote.sync_latest_versions = 3

# Sync all versions (default, existing behavior)
remote.sync_latest_versions = 0
```

Works with version specs in requirements.yml - the limit applies within the matched range:
```yaml
collections:
  - name: community.crypto
    version: ">=2.0.0,<3.0.0"  # with sync_latest_versions=1, syncs only the newest in this range
```

##### WHAT CHANGED

- **Model:** `sync_latest_versions = PositiveIntegerField(default=0)` on `CollectionRemote`
- **Serializer:** Field exposed in the API
- **Sync pipeline:** `_limit_to_highest_versions()` filters matched versions after requirement matching, in both paginated and unpaginated paths. Pre-releases are excluded from the count.
- **Migration:** `0066_collectionremote_sync_latest_versions.py`

##### TESTING

Tested against `https://galaxy.ansible.com/`:
- `sync_latest_versions=1` synced exactly 1 version of community.crypto (latest stable)
- `sync_latest_versions=0` synced all versions
- Pinned versions still respected with `sync_latest_versions=1`
- Version ranges with `sync_latest_versions=1` synced the newest within the range
- All 13 unit tests pass, 4 new functional tests pass

Related: [AAP-50432](https://redhat.atlassian.net/browse/AAP-50432), #2471

Assisted-by: Claude Code, model: claude-opus-4-6 <noreply@anthropic.com>

### 📜 Checklist

- [x] Commits are cleanly separated with meaningful messages
- [x] A changelog entry has been added for any significant changes
- [x] Follows the Pulp policy on AI Usage
- [ ] (For new features) - User documentation and test coverage has been added

[AAP-50432]: https://redhat.atlassian.net/browse/AAP-50432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dragid10:
🤖 **Code Review**

### High Care Amount

**1. The paginated path duplicates an HTTP request to the collection metadata endpoint.**

`_resolve_latest_version` fetches `{collection_endpoint}{namespace}/{name}` to get `highest_version`. But `_fetch_paginated_collection_metadata` (line 816-817) already downloads that same URL. This means every paginated "latest" sync makes a redundant network call. Could you explain the thinking behind not reusing the metadata that `_fetch_paginated_collection_metadata` already fetches? One option: resolve "latest" inside `_fetch_paginated_collection_metadata` itself, where the collection metadata is already available.

**2. No error handling when the collection has zero versions.**

If `all_versions` is empty in either branch, `max(stable)` raises `ValueError: max() arg is an empty sequence`. This could happen if a collection exists but has no published versions. A `CollectionNotFound` or similar error with a clear message would be more helpful than a traceback.

**3. `"latest"` is case-sensitive and undiscoverable.**

The check is `requirements_entry.version == "latest"` (exact match). If a user writes `"Latest"` or `"LATEST"`, it falls through to `AnsibleSpec("Latest")`, which will raise a confusing `ValueError` from `semantic_version`. Have you considered a case-insensitive comparison, or is there precedent in the Ansible/Galaxy ecosystem for case-sensitive version specifiers that would make that surprising?

### Medium Care Amount

**4. Duplicated version-resolution logic between unpaginated and paginated branches.**

The "filter pre-releases, fall back to all, take max" pattern appears twice in `_resolve_latest_version`. If this logic ever changes (e.g., to also exclude yanked versions), both copies need updating. Extracting a small helper like `_pick_highest_stable(versions: list[dict]) -> str` would eliminate that.

**5. The unpaginated branch silently skips missing collections.**

`self._unpaginated_collection_version_metadata` is a `defaultdict(lambda: defaultdict(list))`, so accessing a missing collection returns an empty list rather than raising an error. Combined with #2, `max([])` would crash. The existing `_read_from_downloaded_metadata` handles this by raising `CollectionNotFound`; worth matching that behavior here.

### Medium-Low Care Amount

**6. Functional tests hit `galaxy.ansible.com` directly.**

All four tests sync from the live Galaxy server, making them network-dependent and rate-limited. I notice existing functional tests do the same, so this may be the convention. But `test_sync_latest_does_not_affect_wildcard` uses `community.molecule` and asserts `content.count > 1`, which would break if that collection is ever consolidated to a single version.

MasonMines2006:
Addressed the following feedback from in-person review:

1. **Duplicate HTTP request** - Moved "latest" resolution into `_fetch_paginated_collection_metadata` where the collection metadata is already fetched. No extra API call now.
2. **Empty collection handling** - Added `CollectionNotFound` for missing/empty collections in the unpaginated path. Extracted `_pick_highest_stable()` which returns `None` for empty lists.
3. **Case sensitivity** - Check is now case-insensitive (`.lower() == "latest"`).
4. **Duplicated logic** - Extracted `_pick_highest_stable()` helper used by both paths.
5. **Missing collection in unpaginated path** - Raises `CollectionNotFound` matching existing behavior.
6. **Tests hitting live Galaxy** - Left as-is since existing tests follow the same convention.

MasonMines2006:
**Update:** The limit now only applies to collections with `version: "*"` (the default when no version is specified). Explicit pins and version ranges are always fully respected.

**Known tradeoff:** The parser defaults unversioned collections to `"*"`, so there's no way to distinguish "user explicitly wrote `*`" from "user left version blank." If `sync_latest_versions` is set, both are limited. Users who want all versions of a specific collection can use `version: ">=0.0.0"` as a workaround, or use a separate remote with `sync_latest_versions=0`.

MasonMines2006:
Updated: reverted the "*"-only restriction. The limit now applies to all version specs including ranges, matching the behavior @mdellweg  described. Users who need specific versions can use exact pins (`==3.1.4`).

CHANGES/2563.feature
```diff
@@ -0,0 +1 @@
+Added support for `version: "latest"` in collection requirements files to sync only the latest stable version of a collection.
```
mdellweg: Please adjust the changelog here...

pulp_ansible/app/models.py
```diff
@@ -421,6 +421,7 @@ class CollectionRemote(Remote, AutoAddObjPermsMixin):
     token = EncryptedTextField(null=True)
     sync_dependencies = models.BooleanField(default=True)
     signed_only = models.BooleanField(default=False)
+    sync_latest_versions = models.PositiveIntegerField(default=0)
```
mdellweg: Can we call this `sync_highest_versions` in accordance with other language in the ansible ecosystem.

mdellweg: Also we need to make this ZDU compatible. Please add `null=True` and remove the default.

pulp_ansible/app/tasks/collections.py
```diff
@@ -867,29 +875,41 @@ async def _read_from_downloaded_metadata(self, name, namespace, requirement) ->
...
+                    self._add_collection_version(
+                        self._api_version, collection_version_url, col_version_metadata
                     )
+                )
         self.parsing_metadata_progress_bar.total += len(coros)
         await self.parsing_metadata_progress_bar.asave(update_fields=["total"])
         return coros
 
+    def _limit_to_highest_versions(self, version_dicts):
+        """Keep only the N highest non-pre-release versions."""
```
mdellweg:
```suggestion
        """Select specified number of highest versions. Prefer normal versions over prereleases."""
```

pulp_ansible/app/tasks/collections.py
```diff
@@ -867,29 +875,41 @@ async def _read_from_downloaded_metadata(self, name, namespace, requirement) ->
...
+        stable = [v for v in version_dicts if not Version(v["version"]).prerelease]
+        if not stable:
+            stable = version_dicts
+        stable.sort(key=lambda v: Version(v["version"]), reverse=True)
+        return stable[: self.sync_latest_versions]
```
mdellweg: I think we should still sync the prerelease if there was never a real version (in the same way as `is_highest` can select a prerelease).

You should be able to use
```
def sort_key(v):
    version = Version(v["version"])
    return (version.prerelease, version)
```
on the whole list.

mdellweg (changes requested):

pulp_ansible/app/tasks/collections.py
```diff
@@ -829,24 +830,31 @@ async def _fetch_paginated_collection_metadata(
...
-                            collection_version_detail_url,
-                        )
-                    )
+                    matched_versions.append(collection_version)
             next_value = self._get_response_next_value(api_version, collection_versions_list)
             if not next_value:
                 break
             page_num = page_num + 1
 
+        if self.sync_highest_versions:
```
mdellweg, line 839: This silently works for both "None" and "0"...

pulp_ansible/app/serializers.py
```diff
@@ -248,6 +248,16 @@ class CollectionRemoteSerializer(RemoteSerializer):
         required=False,
     )
 
+    sync_highest_versions = serializers.IntegerField(
+        help_text=_(
+            "The number of highest versions of each collection to sync. "
+            "Unset or null syncs all versions. 1 syncs only the highest version."
```
mdellweg, line 254: "unset" and "null" are the same.

mdellweg (approved):

PR #2566, opened 2026-07-10
feat(remotes/git): use username/password for authentication against git server

### LINKED ISSUE #2565: Git remotes ignore user authentication
**Is your feature request related to a problem? Please describe.**
I want to sync from a private git repo using a git remote. However, pulp_ansible will ignore what I set for username/password and run into an error during clone.

**Describe the solution you'd like**
If I set username/password on a git remote, pulp_ansible should use them!

**Bug vs. Feature Request**
Given the remote has the username/password fields on it (they are inherited from pulpcore), this issue feels like a bug to users: "I used the fields but they don't work!" While from a pulp_ansible developer point of view this issue is a feature request, since authentication was never a supported workflow in the past. Feel free to convert this ticket to type "bug" for the user point of view.

**Additional context**
Here is a detailed description of a workflow that illustrates the problem:
```
# Create a private GitHub repo to test with:

1. Create the git repo: go to "+" drop down menu top right on GitHub > New repository
   Choose my_private_test_repo for the Repository name; set "Choose visibility" to "Private"; Select Add README;
2. Create access tokens: go to Settings > Developer Settings > Personal access tokens > Fine-grained tokens > Generate new token
   Select "Only select repositories" and choose your test repo; add the Contents (Read-only) Repositories permission; click "Generate token"
   Copy the token immediately, it is only shown once!
   Also create and copy a Contents: "Read and write" token for your repo!

# Environmental variables:

export USER_NAME='quba42'
export READ_TOKEN='my_token'
export WRITE_TOKEN='my_token'

# Add a minimal Ansible collection to our private Git repo:

git clone https://${USER_NAME}:${WRITE_TOKEN}@github.com/${USER_NAME}/my_private_test_repo.git
cd my_private_test_repo
ansible-galaxy collection init test.minimal
mv test/minimal/* . # We need the collection in the Git repo root for pulp_ansible?!
rmdir test/minimal/
rmdir test
# Uncomment requires_ansible in meta/runtime.yml!
git add --all
git commit -m 'Add test/minimal empty ansible collection'
git push

# Now try syncing this with Pulp:

Note: I am using Pulp oci-env and Pulp CLI (with [this change](https://github.com/pulp/pulp-cli/pull/1440)) both locally installed in separate virtual python environments to test with.

workon oci-env
oci-env compose up

workon pulp_cli
alias pulp='pulp -p oci-env'
pulp status

NAME='ansible-from-private-github-repo'
REMOTE_OPTIONS=(
  --url=https://github.com/${USER_NAME}/my_private_test_repo.git
  --username=${USER_NAME}
  --password=${READ_TOKEN}
)

pulp ansible remote --type git create --name=${NAME} ${REMOTE_OPTIONS[@]}
pulp ansible repository create --name=${NAME} --remote=git:${NAME}
pulp ansible repository sync --name=${NAME}
```

You will be getting the following error:
```
$ pulp show --href /pulp/api/v3/tasks/019f4b39-4976-7331-9e12-e654c4d3b572/ | jq -r .error.traceback
  File "/usr/local/lib/python3.11/site-packages/pulpcore/tasking/tasks.py", line 81, in _execute_task
    result = task_function()
             ^^^^^^^^^^^^^^^
  File "/src/pulp_ansible/pulp_ansible/app/tasks/git.py", line 56, in synchronize
    repository_version = d_version.create()
                         ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pulpcore/plugin/stages/declarative_version.py", line 172, in create
    loop.run_until_complete(pipeline)
  File "/usr/lib64/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pulpcore/plugin/stages/api.py", line 238, in create_pipeline
    await asyncio.gather(*futures)
  File "/usr/local/lib/python3.11/site-packages/pulpcore/plugin/stages/api.py", line 40, in __call__
    await self.run()
  File "/usr/local/lib/python3.11/site-packages/asgiref/sync.py", line 557, in thread_handler
    raise exc_info[1]
  File "/src/pulp_ansible/pulp_ansible/app/tasks/git.py", line 109, in run
    d_content = await declarative_content_from_git_repo(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/src/pulp_ansible/pulp_ansible/app/tasks/collections.py", line 150, in declarative_content_from_git_repo
    gitrepo = Repo.clone_from(
              ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/git/repo/base.py", line 1568, in clone_from
    return cls._clone(
           ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/git/repo/base.py", line 1439, in _clone
    finalize_process(proc, stderr=stderr)
  File "/usr/local/lib/python3.11/site-packages/git/util.py", line 512, in finalize_process
    proc.wait(**kwargs)
  File "/usr/local/lib/python3.11/site-packages/git/cmd.py", line 423, in wait
    raise GitCommandError(remove_password_if_present(self.args), status, errstr)

$ pulp show --href /pulp/api/v3/tasks/019f4b39-4976-7331-9e12-e654c4d3b572/ | jq -r .error.description
Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v --depth=1 --recurse-submodules -- https://github.com/quba42/my_private_test_repo.git 9b962ec3-26f4-4b4b-af2f-0e09fa213214
  stderr: 'Cloning into '9b962ec3-26f4-4b4b-af2f-0e09fa213214'...
fatal: could not read Username for 'https://github.com': No such device or address'
```

quba42:
closes #2565

### A note on attribution

These initial code changes were written by my colleague @Thorben-D, but I have been tasked with shepherding this PR to completion and merge. I have tested that the initial implementation works against the workflow described in the accompanying ticket, and I have added a changelog.

### 📜 Checklist

- [x] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [x] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [x] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [x] (For new features) - User documentation and test coverage has been added

### DIFF
--- a/pulp_ansible/app/tasks/collections.py
+++ b/pulp_ansible/app/tasks/collections.py
@@ -9,7 +9,7 @@ from collections import defaultdict
 from collections.abc import Coroutine
 from gettext import gettext as _
 from pathlib import Path
-from urllib.parse import urljoin
+from urllib.parse import quote, urljoin, urlparse, urlunparse
 from uuid import uuid4
 
 import yaml
@@ -137,19 +137,52 @@ def _save_collection_version(collection_version, artifact):
         )
 
 
-async def declarative_content_from_git_repo(remote, url, git_ref=None, metadata_only=False):
+def _add_auth_to_git_url(url, remote):
+    if remote.password:
+        parsed = urlparse(url)
+        encoded_password = quote(remote.password, safe="")
+
+        if remote.username:
+            encoded_username = quote(remote.username, safe="")
+            new_netloc = f"{encoded_username}:{encoded_password}@{parsed.netloc}"
+        else:
+            new_netloc = f"token:{encoded_password}@{parsed.netloc}"
+
+        git_url = urlunparse(
+            (parsed.scheme, new_netloc, parsed.path, parsed.params, parsed.query, parsed.fragment)
+        )
+    else:
+        git_url = url
+
+    return git_url
+
+
+async def declarative_content_from_git_repo(
+    remote, url, git_ref=None, metadata_only=False, authenticate_url=False
+):
     """Returns a DeclarativeContent for the Collection in a Git repository."""
+    if authenticate_url:
+        clone_url = _add_auth_to_git_url(url, remote)
+    else:
+        clone_url = url
+
     if git_ref:
         try:
             gitrepo = Repo.clone_from(
-                url, str(uuid4()), depth=1, branch=git_ref, multi_options=["--recurse-submodules"]
+                clone_url,
+                str(uuid4()),
+                depth=1,
+                branch=git_ref,
+                multi_options=["--recurse-submodules"],
             )
         except GitCommandError:
-            gitrepo = Repo.clone_from(url, str(uuid4()), multi_options=["--recurse-submodules"])
+            gitrepo = Repo.clone_from(
+                clone_url, str(uuid4()), multi_options=["--recurse-submodules"]
+            )
             gitrepo.git.checkout(git_ref)
     else:
         gitrepo = Repo.clone_from(
-            url, str(uuid4()), depth=1, multi_options=["--recurse-submodules"]
+            clone_url, str(uuid4()), depth=1, multi_options=["--recurse-submodules"]
         )
     commit_sha = gitrepo.head.commit.hexsha
     metadata, artifact_path = sync_collection(gitrepo.working_dir, ".")
@@ -802,7 +835,7 @@ class CollectionSyncFirstStage(Stage):
 
     async def _add_collection_version_from_git(self, url, gitref, metadata_only) -> list[Coroutine]:
         d_content = await declarative_content_from_git_repo(
-            self.remote, url, gitref, metadata_only=False
+            self.remote, url, gitref, metadata_only=False, authenticate_url=False
         )
         await self.put(d_content)
         return []
--- a/CHANGES/2565.feature
+++ b/CHANGES/2565.feature
@@ -1,0 +1,1 @@
+Support syncing from private repos using git remotes.
--- a/pulp_ansible/app/tasks/git.py
+++ b/pulp_ansible/app/tasks/git.py
@@ -107,7 +107,11 @@ class GitFirstStage(Stage):
             message="Cloning Git repository for Collection", code="sync.git.clone"
         ) as pb:
             d_content = await declarative_content_from_git_repo(
-                self.remote, self.remote.url, self.remote.git_ref, self.metadata_only
+                self.remote,
+                self.remote.url,
+                self.remote.git_ref,
+                self.metadata_only,
+                authenticate_url=True,
             )
             await self.put(d_content)
             await pb.aincrement()
--- a/pulp_ansible/tests/unit/test_tasks.py
+++ b/pulp_ansible/tests/unit/test_tasks.py
@@ -4,6 +4,7 @@ from django.test import TestCase
 
 from pulp_ansible.app.models import AnsibleDistribution, AnsibleRepository, CollectionVersion
 from pulp_ansible.app.tasks.collections import (
+    _add_auth_to_git_url,
     _rebuild_collection_version_meta,
     rebuild_repository_collection_versions_metadata,
 )
@@ -114,3 +115,43 @@ class TestCollectionReImport(TestCase):
         cv2 = cobject.cast()
         cv2.refresh_from_db()
         assert cv2.contents != ["a", "b", "c"]
+
+
+class TestAddAuthToGitUrl(TestCase):
+    """Test add_auth_to_git_url method."""
+
+    def test_add_auth_to_git_url_without_auth(self):
+        """Test that URL is unchanged when no password is provided."""
+        remote = mock.Mock()
+        remote.username = None
+        remote.password = None
+        url = "https://my-git-server.com/baz/my_public_repo.git"
+        final_url = _add_auth_to_git_url(url, remote)
+        assert final_url == url
+
+    def test_add_auth_to_git_url_with_username_and_password(self):
+        """Test adding username and password to a git URL."""
+        remote = mock.Mock()
+        remote.username = "foo"
+        remote.password = "bar"
+        url = "https://my-git-server.com/baz/my_private_repo.git"
+        final_url = _add_auth_to_git_url(url, remote)
+        assert final_url == "https://foo:bar@my-git-server.com/baz/my_private_repo.git"
+
+    def test_add_auth_to_git_url_with_password_only(self):
+        """Test adding password only (token) to a git URL."""
+        remote = mock.Mock()
+        remote.username = None
+        remote.password = "bar"
+        url = "https://my-git-server.com/baz/my_private_repo.git"
+        final_url = _add_auth_to_git_url(url, remote)
+        assert final_url == "https://token:bar@my-git-server.com/baz/my_private_repo.git"
+
+    def test_add_auth_to_git_url_with_special_characters(self):
+        """Test that special characters in username/password are URL-encoded."""
+        remote = mock.Mock()
+        remote.username = "foo:@"
+        remote.password = "bar:@"
+        url = "https://my-git-server.com/baz/my_private_repo.git"
+        final_url = _add_auth_to_git_url(url, remote)
+        assert final_url == "https://foo%3A%40:bar%3A%40@my-git-server.com/baz/my_private_repo.git"


quba42:
Note my accompanying Pulp CLI change for ease of testing: https://github.com/pulp/pulp-cli/pull/1440

quba42:
Applied ruff linting.

pulp_ansible/app/tasks/collections.py
```diff
@@ -138,17 +138,39 @@ def _save_collection_version(collection_version, artifact):
 
 async def declarative_content_from_git_repo(remote, url, git_ref=None, metadata_only=False):
     """Returns a DeclarativeContent for the Collection in a Git repository."""
+
+    if remote.password:
+        parsed = urlparse(url)
+        # Not sure if quote is really necessary here
```
mdellweg: Do you have a plan to find out?

(I guess a " " in a password would be a discriminator.)

quba42: I will run some experiments.

pulp_ansible/app/tasks/collections.py
```diff
@@ -138,17 +138,39 @@ def _save_collection_version(collection_version, artifact):
...
+        encoded_password = quote(remote.password, safe="")
+
+        if remote.username:
+            encoded_username = quote(remote.username, safe="")
+            new_netloc = f"{encoded_username}:{encoded_password}@{parsed.netloc}"
+        else:
+            new_netloc = f"token:{encoded_password}@{parsed.netloc}"
+
+        git_url = urlunparse(
```
quba42, line 150: About test coverage:

I am unsure how one would go about writing a functional test for this, since we don't have any simple way of running a git fixture that requires authentication. I also did not find anything like this for the other remote types or for pulp_rpm.

A unit test that checks the git_url is assembled as expected should be feasible. Perhaps assembling the git_url should be its own helper method (which would make it easier to write unit tests for it)?

I also don't see much by way of existing crud tests for the git remote, but I am not sure how valuable adding those would be?

Any thoughts on what I should/should not pursue to achieve some test coverage?

mdellweg: > Perhaps assembling the git_url should be its own helper method (which would make it easier to write unit tests for it)?

Not exactly test driven development, but at least test driven refactoring.

I agree with your assessment.

pulp_ansible/app/tasks/collections.py
quba42: I have been testing using GitHub access tokens, and as far as I can tell those do not include quotable characters, so I do not have any immediate ideas how to figure out what the effect of quote is. :neutral_face:

quba42: So, based on my GitHub usage, I have not observed any relevant special characters within tokens, and they are straight up forbidden in usernames. That being said, the cost of encoding is small and GitHub is not the only Git server out there. My understanding is that if there are relevant special characters in the username/password, then we do need to encode, so in conclusion: I think we should leave this as is. (And I should drop the comment).

mdellweg: Well the question is probably how url_parse handles it. Or if adding quotemarks changes the username or password as later used. In short, the quoting should match the way it is unquoted later. And i have no clue about that.

pulp_ansible/app/tasks/collections.py
quba42: Placed the changes into the `add_auth_to_git_url` function and added several unit tests for it.

quba42:
I will have a look at the failed test. From where it is happening, it could well be related to these changes.

quba42:
I can reproduce the failing test locally don't have any ideas about fixing it yet though.

I also renamed my method to start with `_` for internal method.

quba42:
Ok, for me locally all kinds of tests (including that one and including for main branch) are failing. Not sure how to continue. :unamused:

pulp_ansible/app/tasks/collections.py
sbernhard, line 201: Would this store the username / password in the Database in plain text or also show the username/password in logs?

pulp_ansible/app/tasks/collections.py
```diff
@@ -136,8 +136,30 @@ def _save_collection_version(collection_version, artifact):
...
+        )
+    else:
+        git_url = url
+
+    return git_url
+
+
 async def declarative_content_from_git_repo(remote, url, git_ref=None, metadata_only=False):
     """Returns a DeclarativeContent for the Collection in a Git repository."""
+    url = _add_auth_to_git_url(url, remote)
```
sbernhard: how would this behave if url is not a "http" url but file or ssh? should the username/password only be sued if http is used?
if the url does already contain username/password, it would be added again with the add_auth_to_git_url method.

pulp_ansible/app/tasks/collections.py
quba42: Good catch! I don't think we want to use the URL with authentication here. The way I think it should work is that we save the original URL here, and then the remote artifact should recombine this with the authentication by checking the remote once the artifact is actually download. On the remote the authentication is stored in an encrypted DB field that is not being logged.

I will run some tests with on_demand syncs to confirm.

pulp_ansible/app/tasks/collections.py
quba42: I think it should work the same way for other protocols. If the user already added authentication to the URL itself, and also username/password fields on the remote, then a sane behavior might be to overwrite the auth from the original URL with the auth from the remote, or else throw an error since the user should not use things this way.

First I will run some tests to confirm what is actually happening in that case.

sbernhard: I would propose not to overwrite it and use the already existing username/password but maybe write a log message. 
Please add a test for this use-case.

pulp_ansible/app/tasks/collections.py
quba42: auth is no longer part of the url, so this is resolved.

pulp_ansible/app/tasks/collections.py
quba42: Now that we are no longer overwriting anything in the user supplied URL, I would argue it is user responsibility not to supply two conflicting auth mechanisms. Suppose the user supplies an ssh URL and sets up Pulp to have access to the relevant ssh key, and then also sets a username and password on the remote. This does not make a lot of sense and I am not sure we should try to guess the users intentions.

pulp_ansible/app/tasks/collections.py
```diff
@@ -137,20 +138,39 @@ def _save_collection_version(collection_version, artifact):
...
-async def declarative_content_from_git_repo(remote, url, git_ref=None, metadata_only=False):
+def _set_auth_environment(url, remote, authenticate):
+    if authenticate and remote.password:
+        username = remote.username if remote.username else "token"
+        auth = base64.b64encode(f"{username}:{remote.password}".encode()).decode()
+        environment = {
+            "GIT_CONFIG_COUNT": "1",
+            "GIT_CONFIG_KEY_0": f"http.{url}/.extraHeader",
+            "GIT_CONFIG_VALUE_0": f"Authorization: Basic {auth}",
+        }
```
sbernhard:
```suggestion
        auth_url = url.rstrip("/")
        environment = {
            "GIT_CONFIG_COUNT": "1",
            "GIT_CONFIG_KEY_0": f"http.{auth_url}/.extraHeader",
            "GIT_CONFIG_VALUE_0": f"Authorization: Basic {auth}",
        }
```

sbernhard: or, directly within ```{url.rstrip('/')}```

quba42: Applied.

quba42:
Update: We ended up changing the approach to send the basic auth to the git clone command via environmental variables. This prevents the credentials from being stored in any git config files on disk, and also prevents them from leaking via the error message when a clone fails.

All other changes I added, like reusing `common_clone_params` for the three invocations of `Repo.clone_from`, are intended to improve the readability. 

Now that tests are green again with the new approach, this is ready for review.

sbernhard (approved):
Thanks!

pulp_ansible/app/tasks/collections.py
```diff
@@ -137,20 +139,39 @@ def _save_collection_version(collection_version, artifact):
...
 
 
-async def declarative_content_from_git_repo(remote, url, git_ref=None, metadata_only=False):
+def _set_auth_environment(url, remote):
+    username = remote.username if remote.username else "token"
+    auth = base64.b64encode(f"{username}:{remote.password}".encode()).decode()
+    environment = {
+        "GIT_CONFIG_COUNT": "1",
+        "GIT_CONFIG_KEY_0": f"http.{url.rstrip('/')}/.extraHeader",
+        "GIT_CONFIG_VALUE_0": f"Authorization: Basic {auth}",
```
mdellweg, line 148: Uhhh you are injecting an authentication header into git. Well...

What is that about the GIT_CONFIG_COUNT?

pulp_ansible/app/tasks/collections.py
```diff
@@ -137,20 +139,39 @@ def _save_collection_version(collection_version, artifact):
...
         except GitCommandError:
-            gitrepo = Repo.clone_from(url, str(uuid4()), multi_options=["--recurse-submodules"])
+            rmtree(clone_path, ignore_errors=True)
+            gitrepo = Repo.clone_from(**common_clone_params)
             gitrepo.git.checkout(git_ref)
     else:
-        gitrepo = Repo.clone_from(
-            url, str(uuid4()), depth=1, multi_options=["--recurse-submodules"]
-        )
+        gitrepo = Repo.clone_from(**common_clone_params, depth=1)
```
mdellweg, line 174: Ugh. Convoluted retry logic. Not your fault.

But does the second path never fail on the "depth=1" option?

pulp_ansible/app/tasks/collections.py
quba42: Apparently this just tells Git that I am providing exactly one config setting via these variables (`_KEY_0`, `_VALUE_0`). As to why this Git env var interface wants me to announce this using GIT_CONFIG_COUNT, I am not sure. But my understanding is that is how this interface is defined.

pulp_ansible/app/tasks/collections.py
quba42: The way I read it, is that if we have a git_ref, there is a particular error that we want to catch and try again in another way. If we don't have a git_ref (else branch) and the command we use there fails, then we just fail at that point. I did not exactly question why the code was structured this way. But I guess if the second attempt is meant to handle an issue with `depth=1`, then it should really apply to the else path as well...

quba42: Did I mention I think I made this at least slightly more readable? :laughing:

quba42: Should I change it to:
```
    try:
        if git_ref:
            gitrepo = Repo.clone_from(**common_clone_params, depth=1, branch=git_ref)
        else:
            gitrepo = Repo.clone_from(**common_clone_params, depth=1)
    except GitCommandError:
        rmtree(clone_path, ignore_errors=True)
        gitrepo = Repo.clone_from(**common_clone_params)
        if git_ref:
            gitrepo.git.checkout(git_ref)
```
?

mdellweg: Or even:
```
if git_ref:
    common_clone_params["ref"] = git_ref
```
If there's a chance to pass that as a kwarg.

mdellweg: After a bit of discussion, we found that the retry path is actually for specifying a commit instead of  branch name. So this suggestion i moot.

mdellweg (approved):

PR #2589, opened 2026-07-23
Taught to use find_api_root() and prepare for pulp-api v4.

ggainey:
https://redhat.atlassian.net/browse/PULP-600


ggainey:
RE lowerbounds - need to either bump LB to core/108, or backport find_api_root() to 105. 108 isn't a supported-branch, so I think the latter is preferable.

mdellweg:
Please bump the lb to the appropriate 3.105.z after backporting.

mdellweg (approved):

PR #2627, opened 2026-09-01
Add test coverage for PQC and OpenPGP v6 collection signatures

dralley:
Parametrizes signature upload and server-side signing tests across traditional (v4), OpenPGP v6 (RFC9580), and post-quantum ML-DSA cipher suites. Verification runs through pulpcore's gpg_verify (pysequoia), so no gpg binary or gnupg library is used at runtime; swaps the functional test dependency from python-gnupg to pysequoia accordingly.

Assisted-By: Claude Opus 4.8


.ci/scripts/calc_constraints.py
```diff
@@ -26,10 +26,28 @@
...
+    branch_versions = []
+    for branch in branch_names:
+        version = Version(branch)
+        if len(version.release) == 2:
+            display_version = Version(f"{branch}.0")
+            check_version = Version(f"{branch}.999999")
+        else:
+            display_version = version
+            check_version = version
+        branch_versions.append((display_version, check_version))
```
dralley: This is needed because the lower bound was bumped to 3.117.2

Should this be mirrored over to the template?

mdellweg: I don't even understand how it works.
But apparently every change here should come via the plugin.

So i would suggest we rename `supported_versions` to `support_branches` here and then the old line 32 could be
`v for v in supported_branches if requirement.specifier.contains(Version(f"{v}.999999"))`.
Would that work?

.ci/scripts/calc_constraints.py
```diff
@@ -26,10 +26,14 @@
...
-    supported_versions.append(template["latest_release_branch"])
+    supported_branches = [
+        *template["supported_release_branches"],
+        template["latest_release_branch"],
+    ]
     applicable_versions = sorted(
-        requirement.specifier.filter((Version(v) for v in supported_versions))
+        Version(branch)
+        for branch in supported_branches
+        if requirement.specifier.contains(Version(f"{branch}.999999"))
```
mdellweg, line 36: :+1: This looks much better. Also renaming "v" to "branch" makes it really clear now.

mdellweg (approved):

